PR #4, opened 2017-07-29
upgrade to neo 2.0.1 breaks build, missing lib directive.

coranos:
neo 2.0.1 does not have a netstandard1.6 lib on nuget, only a net461 lib, this causes the build to fail claiming an incompatibility between libraries.

The CLI prompt was still "ant"

The CLI allows importing and exporting of blocks, but the help text doesn't say the ability exists.

I apologize if I did something the wrong way, this is my first github pull request.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -272,9 +272,11 @@ namespace Neo.Shell
                 "\tshow gas\n" +
                 "\tclaim gas\n" +
                 "\tcreate address [n=1]\n" +
-                "\timport key <wif|path>\n" +
-                "\texport key [address] [path]\n" +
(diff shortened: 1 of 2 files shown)


erikzhang:
There isn't a command of "import blocks" in the code.
And there isn't a 1.6 version of netcoreapp.
I'll have to close this pr this time.

coranos:
ah, I typoed and didn't catch it i meant to add netcoreapp1.0 

The "neo" project doesn't target netcoreapp1.0 

https://github.com/neo-project/neo/blob/master/neo/neo.csproj

The "neo-cli" project does target netcoreapp1.0 

https://github.com/neo-project/neo-cli/blob/master/neo-cli/neo-cli.csproj

This causes the neo-cli to fail to build when I do dotnet publish -c release on neo-cli.

PR #5, opened 2017-08-01
added unconnectedPeers and badPeers to OnShowNodeCommand

coranos:
updated to netcoreapp1.1 as the City of Zion unit tests are based on netcoreapp1.1 , so this allows the nuget packages to be used by their neo-cli.

added unconnectedPeers and badPeers to OnShowNodeCommand.

This is useful for debugging the NEO network. For example there was a problem seeing china from chicago, usa on July 31st. I saw the unconnectedPeers node count drop from ~250 to ~70.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -14,6 +14,7 @@ using System.IO.Compression;
 using System.Linq;
 using System.Security;
 using System.Threading.Tasks;
+using System.Net;
 
 namespace Neo.Shell
 {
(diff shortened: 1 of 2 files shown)


coranos:
NuGet.Build.Tasks.Pack is required on VS2017 on OSX to run "create NuGet Package".
otherwise I get the error:
error MSB4057: The target "Pack" does not exist in the project

i run nuget package before running "dotnet publish -c release" because it gives clearer error messages when packages are linked improperly.

erikzhang:
Did you try "dotnet restore"?

PR #22, opened 2017-08-23
Add partial Linux compilation instructions.

Sjors:
These instructions should make it easier to compile on Linux.

However, I couldn't figure out all the steps...


erikzhang:
Did you view the documentations on [docs.neo.org](http://docs.neo.org)?

Sjors:
I read the English instructions [here](http://docs.neo.org/en-us/node/setup.html), but they don't explain how to compile.

erikzhang:
```
dotnet publish -c Release
```

Sjors:
Thanks. I'll try that and update the README. I was also able to run the downloaded binary on Ubuntu.

Sjors:
I updated the instruction to show two alternatives: using the pre-compiled binaries from Github or installing from source. In both cases `show state` returns information, so I guess it works.

`create wallet` throws `error` though, but that might be a separate issue.

Sjors:
If you don't add `.db3` to wallet file name, it will throw `error`. I added an example command that works (`create wallet wallet.db3`).

PR #28, opened 2017-08-30
GetAddressBalance rpc call

Pwnmanship:
Added an rpc call in the neo-cli rpc to get balances of all assets grouped in 'confirmed', 'unconfirmed' and 'spendable'.

This makes it easier to see incomming transactions on an address and helps brokers and exchanges who receive coins on an address.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -286,3 +286,10 @@ __pycache__/
 *.btm.cs
 *.odx.cs
 *.xsd.cs
+*.ldb
+neo-cli/Chain/CURRENT
+neo-cli/Chain/LOCK
+neo-cli/Chain/MANIFEST-038636
+neo-cli/Chain/LOG
+neo-cli/libleveldb.dll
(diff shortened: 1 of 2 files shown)


erikzhang:
Why not just use "getaccountstate"?

Pwnmanship:
@erikzhang because getaccountstate doesn't provide the amount received unconfirmed and confirmed. We like to check how many coins are comming to the address and how many are confirmed.

Pwnmanship:
Unconfirmed = every coin that is not yet confirmed yet
Confirmed = every coin that is already confirmed
Spendable = every coin that is confirmed and hasn't been spend yet (aka the balance you can use)

We mostly use the first one so we can check if we got an incomming transaction.

erikzhang:
Using this way to check for incomming transactions is not accurate and may cause mistakes.
You need to use `getblock` command to look at every transactions in blocks, and check the outputs of transactions for deposit.

PR #37, opened 2017-09-21
Console colors

shargon:


neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -78,10 +83,24 @@ private void RunConsole()
...
+            Console.WriteLine();
+
             while (running)
             {
-                if (ShowPrompt) Console.Write($"{Prompt}>");
+                if (ShowPrompt)
+                {
+                    Console.ForegroundColor = ConsoleColor.Green;
+                    Console.Write($"{Prompt}> ");
+                    Console.ForegroundColor = ConsoleColor.Yellow;
```
erikzhang: If ```ShowPrompt == false```, the color won't be changed to yellow

shargon: should be fixed now :)

PR #44, opened 2017-10-07
reset the console color

coranos:
updated to netcoreapp 2.0

CLI is now color, and the color was not reset on exit.


coranos:
What is preventing you from updating to 2.0?

When I run it in 2.0 everything runs fine.

I get a compiler warning because SCrypt hasn't been updated to 2.0, but that's it.

Will you accept the PR is I remove coreapp 2.0?

erikzhang:
Yes, I will.

coranos:
removed all but resetting the console color.

PR #45, opened 2017-10-10
Create address command

shargon:
`Create address` fix for non numeric parameters, and it will be added to the help command

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -27,7 +27,7 @@ namespace Neo.Shell
 
         protected LocalNode LocalNode { get; private set; }
         protected override string Prompt => "neo";
-        public override string ServiceName => "neo-cli";
+        public override string ServiceName => "NEO-CLI";
 
         private void ImportBlocks(Stream stream)
         {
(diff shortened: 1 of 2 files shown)


neo-cli/Shell/MainService.cs
```diff
@@ -109,9 +109,11 @@ private bool OnCreateAddressCommand(string[] args)
                 Console.WriteLine("error");
                 return true;
             }
-            ushort count = 1;
-            if (args.Length >= 3)
-                count = ushort.Parse(args[2]);
+
+            ushort count;
+            if (args.Length < 3 || !ushort.TryParse(args[2], out count) || count == 0)
```
erikzhang: If you enter a parameter that is not a number, you should get an error instead of the default value.

neo-cli/Shell/MainService.cs
```diff
@@ -267,6 +270,7 @@ private bool OnHelpCommand(string[] args)
                 "\texit\n" +
                 "Wallet Commands:\n" +
                 "\tcreate wallet <path>\n" +
+                "\tcreate address <count>\n" +
```
erikzhang: It is already in help message:
https://github.com/neo-project/neo-cli/blob/master/neo-cli/Shell/MainService.cs#L279

PR #54, opened 2017-11-16
Show wallet height when executing "show state" command

chenzhitong:
Show wallet height when executing "show state" command. Keep in line with neo-gui.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -660,7 +660,7 @@ namespace Neo.Shell
 
         private bool OnShowStateCommand(string[] args)
         {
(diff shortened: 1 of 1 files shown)


neo-cli/Shell/MainService.cs
```diff
@@ -660,7 +660,7 @@ private bool OnShowPoolCommand(string[] args)
 
         private bool OnShowStateCommand(string[] args)
         {
-            Console.WriteLine($"Height: {Blockchain.Default.Height}/{Blockchain.Default.HeaderHeight}, Nodes: {LocalNode.RemoteNodeCount}");
+            Console.WriteLine($"Height: {Program.Wallet.WalletHeight - 1}/{Blockchain.Default.Height}/{Blockchain.Default.HeaderHeight}, Nodes: {LocalNode.RemoteNodeCount}");
```
erikzhang: It will overflow the `uint` when `WalletHeight == 0`

Issue #72, opened 2017-12-26
Discussion: Add Travis-CI to run a few tests on new code pushes


metachris:
I propose adding a continuous integration system to neo-cli to test the basic functionality on every code push. This includes:

* Building and running neo-cli
* Creating wallets and addresses
* Running json-rpc requests
* Querying nep-5 token balances

The reason is that exchanges are integrating NEO and NEP-5 tokens and rely on the latest neo-cli releases. Issues like #71 can have a major impact, where the latest release doesn't allow nep-5 related json-rpc requests.

I would start with a simple `.travis.yml` file like this: https://github.com/metachris/neo-cli/blob/travis-ci/.travis.yml. 

Depending on whether you want to have the Dockerfiles and test scripts also in this repository, I'd start with test scripts that verify the above functionality.

Issue #71, opened 2017-12-21
Error 400 Access Denied on neo-cli 2.5.2


apisit:
I just updated neo-cli on the server to the latest release `2.5.2` 
Server is running Ubuntu 16.04

When I tried to get totalSupply of RPX token I received the following response. In fact, every request to RPX Smart Contract returns error 400.  Another RPC call like `getblockcount` is fine.  

#### Request
```
{
  "jsonrpc": "2.0",
  "method": "invokefunction",
  "params": [
    "ecc6b20d3ccac1ee9ef109af5a7cdb85706b1df9",
    "totalSupply"
  ],
  "id": 3
}
```
#### Response
```
{
	"jsonrpc": "2.0",
	"id": 3,
	"error": {
		"code": -400,
		"message": "Access denied"
	}
}
```

#### Curl
```

curl --request POST \
  --url http://node1.o3.network:10332/ \
  --header 'accept: application/json' \
  --header 'content-type: application/json' \
  --data '{
  "jsonrpc": "2.0",
  "method": "invokefunction",
  "params": [
    "ecc6b20d3ccac1ee9ef109af5a7cdb85706b1df9",
    "totalSupply"
  ],
  "id": 3
}'
```

What can possibly happen here?

Issue #70, opened 2017-12-20
Create address returns "error", does not create new address (v.2.5.2)


titanomachy:
Latest wallet (2.5.2) on Debian.
`neo> create address` returns an "error" string, no address is being created.
`neo> create address 1` returns an "error" string, no address is being created.
`neo> create address 3` returns an "error" string, no address is being created. Since nothing is being logged for this one, I am not sure what is causing it. Is there a debug mode?

metachris:
* Are you sure you have an open wallet before trying `create address`?
* What Debian version are you using.

Tried reproducing it with Debian Stretch, dotnet 2.0 and neo-cli 2.5.2, and it works for me:

    # dotnet neo-cli.dll
    NEO-CLI Version: 2.5.2.0

    neo> create wallet test
    password: ***
    password: ***
    address: AeptyfhH8kGCTupLgUSdee8P9X4VXrP1K9
    pubkey: 03c3baa1d9b3dd6a0f088e58167d9f452c6bfadb631e76f371dfc2b50b360f03c3
    neo> create address
    [1/1]
    export addresses to address.txt
    neo>

titanomachy:
1. Yes, the wallet is open. I just double checked.
2. Debian 8.9

I will compile the wallet to a debug-binary and report back. I don't work with .NET, but is there no other way to be able to implement a debug mode? Extended error messages could be helpful too.

erikzhang:
You can build a debug client with this:
```
dotnet publish -c Debug
```

titanomachy:
The error it returned was "error: Specified method is not supported.". The wallet is open and the commands I used were `create address` and `create address 1`. Both returned this error.

Could we not have a more explicit error feedback in this wallet? I think it would help me, others, and the developers to decrease the time spend on debugging and answering issues posted here on Github.

Eg. The program is actually explicit enough when I don't have my wallet open and I try to use the "create address" command. It states "Console.WriteLine("You have to open the wallet first.");", but if I check the source code for the OnCreateAddressCommand() method in the MainService class it returns the string "error" when the argument length is longer than 3. Should it not just tell me that I use a too high number and that it should have no more than 3 positions (or better: be lower than 999)?

In my case, for the `create address` command, the initial "error" did not tell my what went wrong. In the debug mode the error "Specified method is not supported" tells me a little more, but not why exactly it did not support this method, because the "help" function tells me I should use the command that way.

[...]

PR #77, opened 2017-12-27
Initial continuous integration with Travis CI

metachris:
Proposal of an initial continuous integration setup using [Travis CI](https://travis-ci.org) (see also #72).

If merged and [enabled at Travis CI](https://docs.travis-ci.com/user/getting-started/#To-get-started-with-Travis-CI), code pushes to GitHub will automatically build the code and run a number of tests to verify the neo-cli functionality. 

* The steps executed by Travis CI are defined in `.travis.yml`
* The proposed setup uses a Docker image based on the [microsoft/dotnet:2.0-sdk](https://hub.docker.com/r/microsoft/dotnet/) base image to build the code
* Currently 2 test stages are implemented:
  * **neo-cli tests using [expect](https://linux.die.net/man/1/expect)**: these tests interacts with neo-cli like any normal user, send commands and wait for correct output. This could catch errors like #70. Note: if output of neo-cli commands that are tested with `expect` change, the `test-neo-cli.expect` script will also need to be updated accordingly.
  * **JSON-RPC tests**: neo-cli is started in the background with `dotnet neo-cli.dll --rpc`, and the JSON-RPC api is tested with `curl` commands. Currently one json-rpc test is implemented: querying `totalSupply` of the RPX token. This can catch errors such as #71.

Example output of Travis CI: https://travis-ci.org/metachris/neo-cli/builds/322348356

This PR is an initial CI setup which can easily be extended with more tests in the future.

Looking forward to hearing your thoughts, and whether you are interested in all of those tests or just a part of it, or something completely different. I'd also like to know how you want to name the directory which contains the continuous integration scripts (in this proposal it's simply `/ci`).

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -1,0 +1,14 @@
+This are files for the continuous integration with Travis CI.
+
+On each code push, the following tasks are executed:
+
+* Build the latest code
+* Verify the basic neo-cli functionality using [expect](https://linux.die.net/man/1/expect)
+* Verify JSON-RPC functionality with curl
+
+The CI integration consists of the following parts:
+
+* `Dockerfile`: the system to build neo-cli and to run the tests
+* `build-and-test.sh`: this builds the Docker image, starts it and runs the tests inside. This is useful for testing the CI run on a local dev machine.
+* `run-tests-in-docker.sh`: is run inside the Docker container and executes the tests
+* `test-neo-cli.expect`: [expect](https://linux.die.net/man/1/expect) script which verifies neo-cli functionality
--- a/ci/README.md
+++ b/ci/README.md
@@ -1,4 +1,4 @@
-This are files for the continuous integration with Travis CI.
+These are files for the continuous integration with Travis CI.
 
 On each code push, the following tasks are executed:
 
--- a/ci/run-tests-in-docker.sh
+++ b/ci/run-tests-in-docker.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+#
+# This script is run inside the Docker container and tests neo-cli
+#
+set -e
+
+cd /opt/neo-cli
+
+# Run tests with expect
+expect /opt/ci/test-neo-cli.expect
+
+# Start neo-cli in background for additional JSON-RPC tests
+screen -dmS node1 bash -c "dotnet neo-cli.dll --rpc"
+
+# Wait a little bit
+sleep 3
+
+# Test a RPX smart contract query
+JSONRPC_RES=$( curl --silent \
+  --request POST \
+  --url localhost:10332/ \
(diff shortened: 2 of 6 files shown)


ci/run-tests-in-docker.sh
```diff
@@ -0,0 +1,43 @@
...
+#
+set -e
+
+cd /opt/neo-cli
+
+# Run tests with expect
+expect /opt/ci/test-neo-cli.expect
+
+# Start neo-cli in background for additional JSON-RPC tests
+screen -dmS node1 bash -c "dotnet neo-cli.dll --rpc"
```
AshRolls: We should make sure this screen is closed after tests have run

ci/README.md
```diff
@@ -0,0 +1,14 @@
+This are files for the continuous integration with Travis CI.
```
AshRolls: 'This' should be 'These'

ci/run-tests-in-docker.sh
metachris: The Docker container is killed after the tests, so it wouldn't matter.

ci/README.md
metachris: fixed

AshRolls:
I like these changes, they will add more robustness to this repo. Given the amount of functionality exposed through `neo-cli` for the main `neo` repo, we should try and restrict the testing to general `neo-cli` rather than trying to stretch tests to cover all `neo` functionality. 

`/ci` is fine for folder name

erikzhang:
I have set it up. So how do I test it?

metachris:
To run it locally, you'll need to have [Docker CE](https://docs.docker.com/engine/installation/) installed. Then just run `./ci/build-and-test.sh`.

If you want to set it up with this GitHub repository, enable this repository at Travis CI ([see this guide](https://docs.travis-ci.com/user/getting-started/#To-get-started-with-Travis-CI)), and merge this PR. On all code pushes (including the merge commit), the build at Travis is triggered. [This](https://travis-ci.org/metachris/neo-cli/builds/322348356) is how an example output at Travis looks like.

erikzhang:
So, anyone has tested it locally?

metachris:
I have, and it works for me (OSX). Also works when executed by Travis-CI which uses Ubuntu Linux to run the builds.

metachris:
Works! You can see the output of the builds here: https://travis-ci.org/neo-project/neo-cli

If you want to add a travis-ci build status badge for the master branch like this: [Build Status](https://travis-ci.org/neo-project/neo-cli), you can use this code for the readme (markdown format): 

    [Build Status](https://travis-ci.org/neo-project/neo-cli)

PR #123, opened 2018-03-13
Parallelize create address procedure (Issue #122)

### LINKED ISSUE #122: Neo cli create address hangs up node
I am running neo-cli on 8 core machine.
when I am running "create address 2000" then it only put load on one core, that makes node down.
And even the address generation is too slow

PFA
screen shot 2018-03-13 at 1 57 21 pm

shargon:
Close #122

image


neo-cli/Shell/MainService.cs
```diff
@@ -112,17 +113,28 @@ private bool OnCreateAddressCommand(string[] args)
...
-                addresses.Add(account.Address);
-                Console.SetCursorPosition(0, Console.CursorTop);
-                Console.Write($"[{i}/{count}]");
-            }
+
+                lock (addresses)
+                {
+                    addresses.Add(account.Address);
+                    Console.SetCursorPosition(0, Console.CursorTop);
+                    Console.Write($"[{Interlocked.Increment(ref x)}/{count}]");
```
erikzhang: We don't need `Interlocked.Increment` because we have already created a critical section by locking `addresses`.

a1onzo:
hi,
when i try to create some address, i got this error:
`neo> create address 10000
[171/10000]error
neo> create address 2000
[92/2000]error
neo> create address 2000
[116/2000]error
neo> create address 100
[92/100]error`
i wonder if there is max number limit on command `create address [n=1]` ,
or too many thread created then got a system error?
can u help me ?
neo-cli version : 2.8.0
os version: Ubuntu 16.04.4

PR #144, opened 2018-04-29
RPC improve

belane:
* Add bind address setting for RPC
* Enforce HTTP Basic Auth for wallet operations

need core funcionality on https://github.com/neo-project/neo/pull/226


erikzhang:
It should be implemented in plugins, I think.

belane:
Maybe this can implement it while we build a plugins system, what do you think?
Then should we move all wallet operations through RPC to a plugin?

PR #156, opened 2018-05-16
Create a multi-party signature contract using neo-cli

### LINKED ISSUE #150: Create a multi-party signature contract using neo-cli
Hello, 
I would like to perform this step using neo-cli:
http://docs.neo.org/en-us/network/private-chain.html#5-extracting-neo-and-gas

To do that, I need to create this multi-party signature contract, and this does not seem to be implemented in neo-cli. 
Is there any way I can perform the step mentioned using a linux machine?

Thank you

lock9:
I tried to make it as close as possible to the neo-python version.
I implemented tests, but I'm not committing them here (I had to do some changes to be able to test some things)

Close #150


purplesoft-io:
@erikzhang why you put create multisig contract under import command section?
is it a import command or creation command?
and if input length is bigger than 254 char?
I think json parameter implementation is better.
https://github.com/neo-project/neo-cli/pull/155

neo-cli/Shell/MainService.cs
```diff
@@ -418,11 +420,49 @@ private bool OnImportCommand(string[] args)
...
+                Console.WriteLine("Error. Use at least 2 public keys to create a multisig address.");
+                return true;
+            }
+
+            int m = int.Parse(args[2]);
+            int n = args.Length - 3;
+
+            if (m < 1 || m > n || n > 1024)
+            {
+                Console.WriteLine("Error. Invalid parameters.");
```
purplesoft-io: return true missing?

neo-cli/Shell/MainService.cs
```diff
@@ -418,11 +420,49 @@ private bool OnImportCommand(string[] args)
             {
                 case "key":
                     return OnImportKeyCommand(args);
+                case "multisigaddress":
+                    return OnImportMultisigAddress(args);
```
purplesoft-io: Console.ReadLine() allow 254 chars max by default.... if you have 10 public keys?

lock9:
Hello @erikzhang , your code is working fine. I added the return statement like @purplesoft-io  mentioned.
The characters limitation is something that we should definitely fix. The problem exist, but I won't be able to catch it in my unit tests.
Solutions:
- Use a file containing one public key per line;
- Increase the Console Buffer to ~68.000 chars (66 * 1024 + 'import multisigaddress X ');

lock9:
I increased the buffer size of the console to 68k chars. Note: I'm only redirecting the ReadLine to this wrapper, we might want to do this with all console calls.

erikzhang:
> why you put create multisig contract under import command section?

Because the key is already in the wallet, and we just "import" the contract.

erikzhang:
@lock9 Can't we just use `Console.OpenStandardInput`?

lock9:
@erikzhang  haha indeed. I think I had this in my mind because it's how I managed to implement unit tests

lock9:
@erikzhang  I'm not sure how you expect this to be implemented.
If this has to be performed only once, where do you think we should put it?
`byte[] inputBuffer = new byte[1024 * 66 + 128];
			Stream inputStream = Console.OpenStandardInput(inputBuffer.Length);
			Console.SetIn(new StreamReader(inputStream, Console.InputEncoding, false, inputBuffer.Length));`

lock9:
I moved those lines to Program.cs

neo-cli/Program.cs
```diff
@@ -21,7 +21,11 @@ private static void CurrentDomain_UnhandledException(object sender, UnhandledExc
         static void Main(string[] args)
         {
             AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
-            new MainService().Run(args);
+			byte[] inputBuffer = new byte[1024 * 66 + 128];
```
erikzhang: What is the use of `inputBuffer`?

lock9:
Sorry, when I read the code I thought it was being used by the SetIn method.

neo-cli/Program.cs
```diff
@@ -21,7 +21,11 @@ private static void CurrentDomain_UnhandledException(object sender, UnhandledExc
         static void Main(string[] args)
         {
             AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
-            new MainService().Run(args);
+			var bufferSize = 1024 * 66 + 128;
```
erikzhang: 1024 * 67 + 128
67 is because of 1 space.

PR #162, opened 2018-05-21
Neo cli transaction signature

purplesoft-io:
added 3 new command.
1) sign 
2) broadcast signature
3) lock

let me know if my code needs changes.

Enjoy!
Tiz


erikzhang:
@purplesoft-io 
1. We don't need `lock` in neo-cli.
2. `broadcast tx` is enough.
3. `sign` is good.

purplesoft-io:
hi @erikzhang, I've removed lock and broadcast signature function...
broadcast tx is enough to broadcast signed transaction, I check neo-gui code and I see that use LocalNode.Relay() not LocalNode.GetRemoteNodes().enqueueMessage()?
Is it correct?
Thanks Tiz

purplesoft-io:
Hi @erikzhang,
let me know if this pull is ok.
Bye Tiz

sirpy:
I'm trying to transfer funds from the private blockchain multi sig account to one of the accounts
i've signed the transaction using the newly added "sign" command with 3 out of 4 accounts
i've tried using the "broadcast tx" but it fails. If i give it as input the JSON of the signed transaction the cli crashes. Perhaps the input for "broadcast tx" is something else than the JSON?
I've then tried to restore the code you removed for the "signature" method. It has a bug where it expect 2 arguments but actually only 1 argument is used and needed (the JSON).
When I fed it the JSON object of the signed transaction it seemed to work ok and returned a hash. But the account balances stayed the same. Still the neos are in the initial contract and i'm still unable to claim any gas.
Any ideas?

erikzhang:
Sorry, I'm confused. "broadcast TX" can only send transactions in the memory pool.
I think we need to add a new command "relay".

sirpy:
purplesoft added relay in this commit https://github.com/neo-project/neo-cli/pull/162/commits/7c93fb4c61d4cbddeb0d4bc68b786e26c0864105
... it was called "signature", it tried it, it went without any errors, but it didnt seem that the transaction itself went through, the account balances stayed the same...

purplesoft-io:
Hi @sirpy, in previous version I've implemented broadcast signature and I've tested it with success.
you need to sign and broadcast signature from each wallet... 3 signature and 3 broadcast signature are required.

If you need help, I'm on discord/telegram too.
Bye Tiz

tiziano@purplesoft.io

purplesoft-io:
see istructions on:
http://docs.neo.org/en-us/network/private-chain.html

> Here we want to send the NEO from the contract address to the normal address. To do so, open any of the four wallet, click transaction, transfer and enter the recipient address, in order to transfer 100 million NEO to this recipient address.
The system will then notify you that there have not been enough signatures to complete the transaction. Copy the code to your clipboard, open the second wallet, click transaction, signature and paste the code that you have just copied. Click signature and copy the code. Open the third wallet, click transaction, signature and paste the code that you have just copied. Click the signature button. At this time you will notice a pop-up window that displays a broadcast button in the lower left corner, which means all the signatures required to send the transaction have been collected. Click broadcast. Once the transfer transaction begins broadcasting it will take about 15 seconds for successful remittance to the account.

sirpy:
That's what i did, but I didnt broadcast it from each account, only once! so that must be it..
Thanks!
which telegram channel?

blahd:
From my understanding this feature only supports multi-signature transactions. 

The original feature request in https://github.com/neo-project/neo-gui/issues/4 was to support standard single signature transactions.

PR #180, opened 2018-06-20
Add support support for deploying contracts. Also add support for quoted arguments that may contain spaces.

jsolman:
Linux users can't easily run neo-gui for deploying contracts, and it seems like neo-cli should support deploying and invoking contracts. This gives it the ability to deploy contracts and allows passing params surrounded in quotes so spaces can be included in cli commands.

### DIFF
--- a/neo-cli/Services/ConsoleServiceBase.cs
+++ b/neo-cli/Services/ConsoleServiceBase.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Reflection;
 using System.Security;
 using System.Text;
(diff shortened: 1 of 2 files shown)


jsolman:
This needs more work. I'll re-open it when it's ready.

shargon:
Please, create use unit test for verify all possible inputs

vncoelho:
Great move, @jsolman .

A nice place for checking this would be the neocompiler eco.

When you finished it we can check it out there. Let us know.
It would be easy to test the deployment of several different examples and the interaction with invokes.

jsolman:
I have some changes that add invoking as well. I'm not going to have time to work on this for a little while though, so the PR is just going to sit open for a while.

vncoelho:
Hi, @jsolman, 
Just saw the PR again.

Do you still believe that it is a good feature for neo-cli?
Because one can use several clients for deploying, in particular, in the front-end of the web browsers.

The only problem I see is to keep adding more features to neo-cli and makes it even harder maintenance.
* Because every wallet we would need to use would require a syncing process, right, since not light-wallet functionalities is in the DB? 

If the majority agrees.
I believe that with a little more improvements in the javascripts functionalities few people will keep using a c# based client. But we never know the future...

jsolman:
Closing in favor of the updated branch #273

PR #182, opened 2018-06-22
feat: added getunclaimedgas and claimgas RPC methods.

bmclain:
Exchanges require a method for claiming gas via RPC. 

This PR adds two methods: 

- `getunclaimedgas` which returns the available and unavailable gas.
- `claimgas` which creates a claim TX and relays the tx.

Let me know what you think. It would be great to have this feature in the next release (2.7.7).

### DIFF
--- a/neo-cli/Network/RPC/RpcServerWithWallet.cs
+++ b/neo-cli/Network/RPC/RpcServerWithWallet.cs
@@ -1,10 +1,12 @@
 ﻿using Neo.Core;
 using Neo.Implementations.Wallets.NEP6;
 using Neo.IO;
+using Neo.Shell;
 using Neo.IO.Json;
 using Neo.SmartContract;
 using Neo.Wallets;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


bmclain:
Another potential feature would be to submit an address as a param to `claimgas` to be used as the output for the claim tx.

erikzhang:
This feature can be implemented in plugins. See https://github.com/neo-project/neo/blob/master/neo/Plugins/RpcPlugin.cs

bmclain:
Is it safe to assume that the existing RPC methods are set in stone and no new RPC methods will be added in the future? I only ask because claiming gas and getting the unclaimed amount of gas seem to be core functionality imho. Is there a criteria to decide what should be a plugin and what should be a core RPC call?

erikzhang:
The gas distribution mechanism will be modified in Neo 3.0 and we are not going to accept any relevant pull request before that.
So at present this kind of the features should be implemented in plugins.

PR #187, opened 2018-06-29
close on ^D without errors

fyrchik:
Hello!

I usually use Ctrl+D to exit from all REPLs, however neo-cli exits with error:
```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Neo.Services.ConsoleServiceBase.RunConsole()
   at Neo.Services.ConsoleServiceBase.Run(String[] args)
   at Neo.Program.Main(String[] args)
[1]    14067 abort (core dumped)  dotnet neo-cli/bin/Release/netcoreapp2.0/neo-cli.dll /rpc
```
neo-python, for example, handles this correctly.

This small PR fixes it.

### DIFF
--- a/neo-cli/Services/ConsoleServiceBase.cs
+++ b/neo-cli/Services/ConsoleServiceBase.cs
@@ -14,7 +14,7 @@ namespace Neo.Services
 
         protected bool ShowPrompt { get; set; } = true;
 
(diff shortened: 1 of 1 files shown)


penlite:
Looks good.

PR #188, opened 2018-06-29
NEO-CLI v2.9.0

erikzhang:
- Use Neo v3.0 lib from neo-project/neo#288
- Improve CLI command: `show state`


neo-cli/Shell/MainService.cs
```diff
@@ -908,70 +956,13 @@ protected internal override void OnStart(string[] args)
                     case "-r":
                         useRPC = true;
                         break;
-                    case "--nopeers":
-                        nopeers = true;
-                        break;
-                    case "-l":
-                    case "--log":
```
vncoelho: What happened to application log and log, @erikzhang?

neo-cli/Settings.cs
```diff
@@ -30,12 +30,10 @@ public Settings(IConfigurationSection section)
     internal class PathsSettings
     {
         public string Chain { get; }
-        public string ApplicationLogs { get; }
```
vncoelho: The plan is to handle it later on as a plugin?

neo-cli/Network/RPC/RpcServerWithWallet.cs
```diff
@@ -1,240 +0,0 @@
...
-                case "dumpprivkey":
-                    if (Program.Wallet == null)
-                        throw new RpcException(-400, "Access denied");
-                    else
-                    {
-                        UInt160 scriptHash = Wallet.ToScriptHash(_params[0].AsString());
-                        WalletAccount account = Program.Wallet.GetAccount(scriptHash);
-                        return account.GetKey().Export();
-                    }
-                case "invoke":
```
vncoelho: This solves several interesting points. Perhaps, better to handle like this. Good move.
@igormcoelho

On the other hand, it would be nice for developers to have a testing function for invoke.
What do you suggest? Should this test invoke(s) and notifications checks be developed as a plugin?

erikzhang:
ApplicationLogs is now moved into neo-plugins.
https://github.com/neo-project/neo-plugins/tree/master/ApplicationLogs

vncoelho:
I just saw it there some minutes ago. Thanks, @erikzhang.
How is the best way to initialize logs and Simple Policy?

erikzhang:
> Which is the best way to initialize logs and Simple Policy?

To use a plugin, just put it in the Plugins directory, neo-cli or neo-gui automatically load all the plugins at startup.

RavenXce:
Hi, I would like know if we will be doing an intermediate 2.x.x release with the latest VM that has NEP-8, or is it meant to go into 3.0 only? Thanks!

erikzhang:
> Hi, I would like know if we will also be doing an intermediate 2.x.x release with the latest VM that has NEP-8, or is it meant to go into this 3.0 release only? Thanks!

The first release with NEP-8 will be a preview version of NEO 3.0.
https://www.nuget.org/packages/Neo/3.0.0-preview2-03

Edgegasm:
Can we have a changelog/release notes for 2.9.0 please?

erikzhang:
https://github.com/neo-project/neo-cli/blob/master/CHANGELOG.md

Done.

PR #196, opened 2018-07-12
settings.cs fix

wturyn-cf:
This fixes inconsistency between `config.json` and `settings.cs` handling of `unlockWallet` section in `UnlockWalletSettings` method. In particular: `config.json` contains `Path` and `Password` sections but they are referenced as `WalletPath` and `WalletPassword`.
Also removes bad condition  - `section.Value` is always `null` because it doesn't contain any value only nested sections.

### DIFF
--- a/neo-cli/Settings.cs
+++ b/neo-cli/Settings.cs
@@ -74,7 +74,8 @@ namespace Neo
 
         public UnlockWalletSettings(IConfigurationSection section)
         {
-            if (section.Exists()) {
+            if (section.Exists()) 
+            {
                 this.Path = section.GetSection("Path").Value;
                 this.Password = section.GetSection("Password").Value;
                 this.StartConsensus = bool.Parse(section.GetSection("StartConsensus").Value);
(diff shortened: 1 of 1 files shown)


neo-cli/Settings.cs
```diff
@@ -74,13 +74,10 @@ public class UnlockWalletSettings
 
         public UnlockWalletSettings(IConfigurationSection section)
         {
-            if (section.Value != null)
```
erikzhang: If the config.json doesn't contain the `UnlockWallet` section, then `section.Value != null` returns `false`. This is to be compatible with the old configuration file.

wturyn-cf: I'm going to add commit with `if (section.Exists())` but right now github tells me that there's heavy load and my commits don't go through.

erikzhang: You are right. I will wait for you to add the commit before merging.

wturyn-cf: My commits went through.

PR #203, opened 2018-07-19
Add getwalletheight RPC call

DaveWK:
This adds a RPC call to retrieve the wallet height. This helps with understanding if the wallet has caught up to the block height.

It returns 400 Access denied if a wallet is not loaded in the same way as other RPC calls.

### DIFF
--- a/neo-cli/Network/RPC/RpcServerWithWallet.cs
+++ b/neo-cli/Network/RPC/RpcServerWithWallet.cs
@@ -52,10 +52,7 @@ namespace Neo.Network.RPC
                     if (Program.Wallet == null)
                         throw new RpcException(-400, "Access denied.");
                     else
-                    {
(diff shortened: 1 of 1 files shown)


DaveWK:
@erikzhang should I put this against the neo 3.0 repo instead? this is critical for knowing what height the balance returned by getbalance is.

erikzhang:
> should I put this against the neo 3.0 repo instead?

Yes, please.

DaveWK:
It looks like the 3.0 repo you've removed the wallet-related rpc from neo-cli, and the wallet loading functionality doesn't exist in the /neo/ repo; which should this code live in?

DaveWK:
@erikzhang can this be merged? this really makes it impossible to know if the balance reported is the balance at a particular height..

PR #225, opened 2018-08-29
Minnor plugin for stop loading on a defined height

vncoelho:
Hey, Erik,

We are creating some snapshot tools for reading the storage of any desired block.
This plugin will help us to design it in a better manner.

However, we am still getting this error:
```cs
Shell/MainService.cs(47,57): error CS0117: 'Plugin' does not contain a definition for 'LoadingPlugins' [/opt/neo-cli/neo-cli/neo-cli.csproj]
```
It also happen when we try to include any other one.
```cs
Shell/MainService.cs(47,57): error CS0117: 'Plugin' does not contain a definition for 'Policies' [/opt/neo-cli/neo-cli/neo-cli.csproj]

```
In fact, this is the first Plugin on the neo-cli itself, thus, maybe something is missing.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -5,6 +5,7 @@ using Neo.Network.P2P;
 using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
 using Neo.Persistence.LevelDB;
+using Neo.Plugins;
 using Neo.Services;
 using Neo.SmartContract;
 using Neo.Wallets;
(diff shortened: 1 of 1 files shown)


erikzhang:
If you are creating a plugin, I think it should not be in neo-cli.

Please refer to the code here: https://github.com/neo-project/neo-plugins

vncoelho:
Hi, @erikzhang,

But we need it in the function `` private static IEnumerable<Block> GetBlocks(Stream stream, bool read_start = false)``

We already looked there, the design will be something simple like this:
```cs
namespace Neo.Plugins
{
    public class MaxHeightPlugin : Plugin, ILoadingPlugins
    {
        public void UpdateMaxHeight(ref uint end)
        {
	    if(Settings.Default.MaxHeight != -1 && Settings.Default.MaxHeight < end && Settings.Default.MaxHeight > 1 );
		end = Settings.Default.MaxHeight;
        }

    }
}
```

vncoelho:
As you suggested, let's move it to  `private void OnImport(IEnumerable<Block> blocks)`  function

PR #228, opened 2018-09-08
Deleting import blocks from neo-cli v3.0

vncoelho:
@erikzhang, I deleted the functions.
I tested here and, apparently, it worked as expected. Thanks for the help and teachings.

image


vncoelho:
Sorry, @erikzhang the title was wrong, it is actually: "Deleting import blocks from neo-cli"
I think that we may need to delete the function and the call around here, right? That how I proceeded for using the plugin.

erikzhang:
> Now we also need to delete the export...

This is what I want to say.

PR #236, opened 2018-10-03
Modify claim operation to 50 claims per tx

txhsl:
This PR modified claim gas operation to limited 50 claims per transaction. This may create more than 1 ClaimTransaction during 1 cmd but make it possible to claim gas from 10k transactions. Extra output is provided to show the progress.

### DIFF
--- a/neo-cli/Shell/Coins.cs
+++ b/neo-cli/Shell/Coins.cs
@@ -14,6 +14,7 @@ namespace Neo.Shell
     {
         private Wallet current_wallet;
         private NeoSystem system;
+        public static int MAX_CLAIMS_AMOUNT = 50;
 
         public Coins(Wallet wallet, NeoSystem system)
         {
@@ -51,7 +52,7 @@ namespace Neo.Shell
         }
 
 
-        public ClaimTransaction Claim()
+        public ClaimTransaction[] Claim()
         {
 
             if (this.AvailableBonus() == Fixed8.Zero)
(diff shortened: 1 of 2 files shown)


erikzhang:
Can you change it to send only one transaction at a time by default?

You can add a command line option:

```
claim gas [all]
```

Send one transaction only without `all`, otherwise send multiple.

txhsl:
@erikzhang Sure, now it's an optional choice.

neo-cli/Shell/MainService.cs
```diff
@@ -506,12 +506,33 @@ private bool OnClaimCommand(string[] args)
...
-                    ClaimTransaction tx = coins.Claim();
-                    if (tx != null)
+                    if (args.Length > 2)
                     {
-                        Console.WriteLine($"Tranaction Suceeded: {tx.Hash}");
+                        switch (args[2].ToLower())
+                        {
+                            case "all":
+                                ClaimTransaction[] txs = coins.ClaimAll();
+                                if (txs.Length > 0)
```
erikzhang: The `ClaimAll()` method may return `null` and cause an `NullReferencException` on this line.

erikzhang:
https://github.com/neo-project/neo-cli/blob/13773f78ae7f7ccc78ae0aca1c7ff2f86a499f40/neo-cli/Shell/Coins.cs#L70

Shouldn't this line be:

```c#
Claims = claims.Take(MAX_CLAIMS_AMOUNT).ToArray(),
```
?

txhsl:
:yum: , that's comprehensive.

neo-cli/Shell/Coins.cs
```diff
@@ -87,6 +87,55 @@ public ClaimTransaction Claim()
...
+                Console.WriteLine($"no gas to claim");
+                return null;
+            }
+
+            CoinReference[] claims = current_wallet.GetUnclaimedCoins().Select(p => p.Reference).ToArray();
+            if (claims.Length == 0) return null;
+
+            using (Snapshot snapshot = Blockchain.Singleton.GetSnapshot())
+            {
+                int claim_count = claims.Length / MAX_CLAIMS_AMOUNT + 1;
```
erikzhang: If `claims.Length == MAX_CLAIMS_AMOUNT`, `claim_count` should be 1.

neo-cli/Shell/Coins.cs
```diff
@@ -87,6 +87,55 @@ public ClaimTransaction Claim()
...
+                        {
+                            new TransactionOutput
+                            {
+                                AssetId = Blockchain.UtilityToken.Hash,
+                                Value = snapshot.CalculateBonus(claims.Skip(i * MAX_CLAIMS_AMOUNT).Take(MAX_CLAIMS_AMOUNT)),
+                                ScriptHash = current_wallet.GetChangeAddress()
+                            }
+                        }
+                    };
+                    txs.SetValue((ClaimTransaction)SignTransaction(tx), i);
```
erikzhang: `SignTransaction()` may return `null`, so this line should handle it, right?

neo-cli/Shell/Coins.cs
```diff
@@ -128,10 +129,13 @@ public ClaimTransaction[] ClaimAll()
                             }
                         }
                     };
-                    txs.SetValue((ClaimTransaction)SignTransaction(tx), i);
+                    if ((tx = (ClaimTransaction)SignTransaction(tx)) != null)
```
erikzhang: Should it break the loop if `tx` is `null`?

neo-cli/Shell/MainService.cs
```diff
@@ -512,7 +512,7 @@ private bool OnClaimCommand(string[] args)
                         {
                             case "all":
                                 ClaimTransaction[] txs = coins.ClaimAll();
-                                if (txs != null && txs.Length > 0)
+                                if (txs.Length > 0)
```
txhsl: I think `List<>` won't be `null`, I removed this judgement.

erikzhang: I mean `tx`, not `txs`.

neo-cli/Shell/Coins.cs
txhsl: Stop the procession and leave the left unclaimed?

txhsl:
@erikzhang maybe some change like that? https://github.com/neo-project/neo-cli/pull/236/commits/419ef1159cf388b030fe98f5a1cf981a6b30a499

PR #240, opened 2018-10-15
Missing IP parameter

igormcoelho:
Taken from previous working versions: IPAddress.Any parameter


belane:
@igormcoelho there is already an open PR https://github.com/neo-project/neo-cli/pull/233 to address  https://github.com/neo-project/neo/pull/382 change.

PR #247, opened 2018-10-20
Command not found message instead of error

vncoelho:
Example:
```cs
neo> dump storage
error
neo> abcd
error: command not found abcd
neo>
```


neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -26,7 +26,7 @@ protected virtual bool OnCommand(string[] args)
                     Console.WriteLine(Assembly.GetEntryAssembly().GetName().Version);
                     return true;
                 default:
-                    Console.WriteLine("error");
+                    Console.WriteLine("error: command not found " + args[0].ToLower());
```
erikzhang: Printing the original command name is better.

```suggestion
                    Console.WriteLine("error: command not found " + args[0]);
```

PR #261, opened 2018-11-03
v2.9.2

erikzhang:


vncoelho:
Hey, @erikzhang,

You want the renegeration to be done also by hand, right? It is a good move.
But how would the `PrepareRequest Payload` arrives again?

We are extending this method to ` public class SetViewNumber { public byte ViewNumber; public ConsensusPayload PrepareRequestPayload; public byte[][] SignedPayloads; }`
Which would be the message to be sent between nodes.

neo-cli/Shell/MainService.cs
```diff
@@ -414,8 +414,7 @@ private bool OnHelpCommand(string[] args)
                 "\tshow pool [verbose]\n" +
                 "\trelay <jsonObjectToSign>\n" +
                 "Advanced Commands:\n" +
-                "\tstart consensus\n" +
-                "\tchange view <viewnumber>\n");
```
belane: Why do you remove help? It can be useful.

erikzhang: This can be misleading to average users.

PR #263, opened 2018-11-14
Add plugins help

belane:
- Allow loaded plugins.
- Call plugins help.

 (Needs https://github.com/neo-project/neo/pull/466)

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -66,6 +66,8 @@ namespace Neo.Shell
                     return OnExportCommand(args);
                 case "help":
                     return OnHelpCommand(args);
+                case "plugins":
+                    return OnPluginsCommand(args);
                 case "import":
                     return OnImportCommand(args);
                 case "list":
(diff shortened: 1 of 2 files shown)


neo-cli/Shell/MainService.cs
```diff
@@ -391,6 +394,23 @@ private bool OnHelpCommand(string[] args)
                 "\trelay <jsonObjectToSign>\n" +
                 "Advanced Commands:\n" +
                 "\tstart consensus\n");
+
+            Plugin.SendMessage(new string[] { "plugin-help" });
```
erikzhang: Just send "help"?

erikzhang: Maybe it can be like this:

```
help [plugin-name]
```

The plugin only processes its own help messages. If the `plugin-name` parameter is omitted, then no plugin will process this message. The help information for the node is printed by default.

belane: Cool, done.

PR #273, opened 2018-12-17
Add support for deploying contracts + initial support for invoking contracts. Also add support for quoted arguments that may contain spaces.

jsolman:
Linux users can't easily run neo-gui for deploying contracts, and it seems like neo-cli should support deploying and invoking contracts. This gives it the ability to deploy contracts and allows passing params surrounded in quotes so spaces can be included in cli commands.

This is copied from #180 but updated to work with the latest code base. I tested deploying a 38Kbyte contract. Worked nicely.

### DIFF
--- a/neo-cli/Services/ConsoleServiceBase.cs
+++ b/neo-cli/Services/ConsoleServiceBase.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
 using System.Reflection;
 using System.Security;
 using System.Text;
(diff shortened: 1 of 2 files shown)


neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -120,6 +121,54 @@ private void RunConsole()
...
+                    {
+                        outputArgs.Add(builder.ToString());
+                        builder.Clear();
+                    }
+                }
+
+                for (int index = 0; index < line.Length; index++)
+                {
+                    char c = line[index];
+                    if (c == '\\')
```
erikzhang: Do we need escape characters? This is very inconvenient for the input of the file path under Windows.

```
del "C:\Program Files\Git\git-cmd.exe"
```

jsolman: We could use a different method of escape. If someone needs to include a `”` symbol in the quote we need a way to do that.

jsolman: Maybe we could escape using ‘%’ symbol instead.

erikzhang: There seems to be no chance to include quotes in the argument.

jsolman: what if someone needs to invoke a method and needs to include quote character in a parameter?

erikzhang: This is possible. Then he may also need '\n' or other characters.

erikzhang: I think '\' is okey. We only need to support '\\', '\"', '\n' and '\r'.

jsolman: I think we should support `’\x##` also to cover all others.

jsolman: Escapes also probably shouldn’t be processed if not in quotes. That change needs to be made also.

jsolman: I’m going to sleep now though feel free to adjust it if you want :-)

erikzhang: OK. I will do it now.

neo-cli/Shell/MainService.cs
```diff
@@ -90,6 +93,8 @@ protected override bool OnCommand(string[] args)
                     return OnStartCommand(args);
                 case "upgrade":
                     return OnUpgradeCommand(args);
+                case "contract":
```
erikzhang: Can we change to a format like `deploy contract`, which is consistent with the style of other commands.

neo-cli/Shell/MainService.cs
```diff
@@ -887,6 +1085,11 @@ protected internal override void OnStart(string[] args)
                 }
             store = new LevelDBStore(Path.GetFullPath(Settings.Default.Paths.Chain));
             system = new NeoSystem(store);
+            HandleStart(useRPC);
+        }
+
+        private void HandleStart(bool useRPC)
```
erikzhang: What is the difference with the original?

jsolman: Just made it easier to comment it out for me when I wanted to delay starting manually. Can revert this piece if you like. It could have done something explicit here to wait for plugins to be ready for it to start, but your solution of handling delayed start in the core is fine.

neo-cli/Shell/MainService.cs
jsolman: I made a common prefix for contract related commands and so had both commands start with contract for a common method to handle branching after that point. If you like the wording better as `deploy contact` and `invoke contract`, you could switch it, but I was also thinking of adding `contract getstorage`, and then maybe there are other contract commands also in the future.

superboyiii:
@jsolman Hi, Jeff. I found your logic has conflict with the simplePolicy plugin which could cause the result that there might be a contract which is deployed by neo-cli but still be rejected by simplePolicy.

this is the mainservice
image

this is the simplePolicy
image

For example, if there's a smart contract and tx.size equals 1025, it should be relayed without fee by your logic but it will be rejected by simplePolicy.

jsolman:
@superboyiii the code that calculates it starts with the .001 and then adds based on the script length. I agree that if there are a lot of utxo for the gas it could have the length be too long and the transaction violate the policy. We can improve the calculation in the near future.

PR #274, opened 2018-12-19
Allow run as a windows service

erikzhang:


neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -102,17 +106,54 @@ public static SecureString ReadSecureString(string prompt)
 
         public void Run(string[] args)
         {
-            OnStart(args);
-            RunConsole();
-            OnStop();
+            if (Environment.UserInteractive)
+            {
+                if (args.Length > 0 && args[0] == "/install")
```
shargon: Maybe we should check first if the system is Windows

PR #285, opened 2019-01-10
Help menu enhancement

superboyiii:
Add simple introduction for the plugins' names which neo-cli supports till now.


vncoelho:
This info could be displayed on `OnHelpPlugin` if someone type help plugin or something like that, it would be more clear.

shargon:
This command should show the current loaded plugins, not a hardcoded version

vncoelho:
That is true, @shargon, can be.
But it can also show optional plugins to be loaded, as an informative `help plugins`.

PR #319, opened 2019-03-14
Add `MaxGasInvoke` to `Settings`

### LINKED ISSUE #318: Cancel the limit of free gas when using "invoke", "invokefunction", "invokescript" rpc
When using "invokescript " rpc for big smart contract invoking, it seems because of the free gas limit, the response always returns "Fault Break". Since these kinds of rpc is just for invoking using on local, I think this limit is better to be canceled.

erikzhang:
Closes #318

### DIFF
--- a/neo-cli/Settings.cs
+++ b/neo-cli/Settings.cs
@@ -64,6 +64,7 @@ namespace Neo
         public ushort Port { get; }
         public string SslCert { get; }
         public string SslCertPassword { get; }
+        public Fixed8 MaxGasInvoke { get; }
 
         public RPCSettings(IConfigurationSection section)
         {
@@ -71,6 +72,7 @@ namespace Neo
             this.Port = ushort.Parse(section.GetSection("Port").Value);
             this.SslCert = section.GetSection("SslCert").Value;
             this.SslCertPassword = section.GetSection("SslCertPassword").Value;
+            this.MaxGasInvoke = Fixed8.Parse(section.GetValue("MaxGasInvoke", "0"));
         }
     }
 
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -1133,7 +1133,8 @@ namespace Neo.Shell
                     Settings.Default.RPC.Port,
                     wallet: Program.Wallet,
                     sslCert: Settings.Default.RPC.SslCert,
-                    password: Settings.Default.RPC.SslCertPassword);
+                    password: Settings.Default.RPC.SslCertPassword,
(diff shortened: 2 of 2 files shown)


erikzhang:
@vncoelho Did you test it before approving?

vncoelho:
No, @erikzhang, the designed sounded enough for the issue, we need to check the design on the use of this limit after that.

Since it was  not a draft I though it would be done in parts. But fell free to add anything else here.

vncoelho:
Fell free to close or modify it, @erikzhang.

vncoelho:
Tested with https://github.com/neo-project/neo/commit/bc14798fa20d68140c1605ae916a3587ec1288ac, limits are working, but the same reply with tx null.

erikzhang:
You need to open a wallet for creating `tx` and `script`.

vncoelho:
You are right, it works, @erikzhang, but that wallet needs to have the limit in GAS for that.

```js
{
  "jsonrpc": "2.0",
  "id": 5,
  "result": {
    "script": "67534941a4574c5095a6d23e90c6a4836d0f3fc026",
    "state": "HALT, BREAK",
    "gas_consumed": "60.743",
    "stack": [],
    "tx": "d1011567534941a4574c5095a6d23e90c6a4836d0f3fc02600d3fb2f010000000001876e1c85bbffa2926d681c33b68e2fd0371d7ee4c2af9dc51c87f66a8fbff98a010001e72d286979ee6cb1b7e65dfddfb2e384100b8d148e7758de42e4168b71792c6000111024010000002baa76ad534b886cb87c6b3720a34943d9000fa9014140eb83593a3e16f5ada039635886e2b997ce1944baf117d1ffa68916f5b5c4c2ac6f65c8634f1b006e7d7abe177cf1c68d7b3040837842d136e58ec9799b8f69b4232102b3622bf4017bdfe317c58aed5f4c753f206b7db896046fa7d774bbc4bf7f8dc2ac"
  }
}
```

In this sense, a generic RPC would need to keep GAS in its wallet.

vncoelho:
There is also a minor thing,
It is not working with decimals
image
With 50.743 it failled.

Then, we needed 51GAS for checking this invoke aforementioned:
image

igormcoelho:
Vitor, as far as I know, Gas fees cannot be fractional yet, as this is still open: https://github.com/neo-project/neo/issues/319
I have a proposal to fix this, I'll try to work on that as soon as possible.

vncoelho:
That is true, @igormcoelho, I had forgot.

There is also a ceiling in the plugins code.

PR #322, opened 2019-03-14
Update config files

shargon:
Update config files for https://github.com/neo-project/neo-cli/pull/320

### DIFF
--- a/neo-cli/config.json
+++ b/neo-cli/config.json
@@ -6,7 +6,8 @@
     },
     "P2P": {
       "Port": 10333,
-      "WsPort": 10334
+      "WsPort": 10334,
+      "MaxConnectionsPerAddress": 3
     },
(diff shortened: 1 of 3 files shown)


jsolman:
Since it is the default, you don’t have to add it, maybe add the other new ones as @vncoelho said if you are going to add this.

erikzhang:
https://github.com/neo-project/neo-cli/blob/b4afa6d9485f72b1f4387cbd4a173b3419ba7b0f/neo-cli/Settings.cs#L59

It has a default value, and you don't have to add it. It is generally not recommended to modify these parameters.

vncoelho:
Makes sense, @erikzhang. From this perspective, maybe the config would become a little heavy with plenty of information.

shargon:
ok, we can close it

PR #324, opened 2019-03-15
Fix the policy of extra fee

superboyiii:
Fix the policy of extra fee.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -282,9 +282,9 @@ namespace Neo.Shell
         {
             Fixed8 fee = Fixed8.FromDecimal(0.001m);
 
-            if (tx.Script.Length > 1024)
+            if (tx.Size > 1024)
             {
-                fee += Fixed8.FromDecimal(tx.Script.Length * 0.00001m);
+                fee += Fixed8.FromDecimal((tx.Size - 1024) * 0.00001m);
             }
 
             return Program.Wallet.MakeTransaction(new InvocationTransaction
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -187,7 +187,7 @@ namespace Neo.Shell
             if (tx.Gas < Fixed8.Zero) tx.Gas = Fixed8.Zero;
             tx.Gas = tx.Gas.Ceiling();
 
-            tx = DecorateScriptTransaction(tx);
+            tx = DecorateInvocationTransaction(tx);
 
             return SignAndSendTx(tx);
         }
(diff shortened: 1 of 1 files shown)


neo-cli/Shell/MainService.cs
```diff
@@ -282,9 +282,9 @@ public InvocationTransaction DecorateScriptTransaction(InvocationTransaction tx)
         {
             Fixed8 fee = Fixed8.FromDecimal(0.001m);
 
-            if (tx.Script.Length > 1024)
+            if (tx.Size > 1024)
             {
-                fee += Fixed8.FromDecimal(tx.Script.Length * 0.00001m);
+                fee += Fixed8.FromDecimal((tx.Size - 1024) * 0.00001m);
```
jsolman: The way it is now is more correct for how the transaction fee is calculated.

jsolman: Specifically it currently uses the tx size not the script size:

https://github.com/neo-project/neo-plugins/blob/08cda72187ccd1fe8722856959644b393784c328/SimplePolicy/SimplePolicyPlugin.cs#L109

shargon: I think that should be the `Size`, but not with this subtraction

superboyiii: Yeah, I think the policy need more discussion for determination. But at least, CLI's policy and Plugin's policy should be the same.

jsolman: So looking at this again my reasoning must have been that the difference between the script size and the transaction size should usually always fit in the initial .001 fee. I’d the fee has to add some utxo’s it may make the transaction bigger. So as @shargon said, let’s not subtract the 1024 here so there is some headroom.

QingmingZi: So, with tx.size, can you only perform inaccurate calculations? Do not subtract 1024 means that the new utxo size can not exceed 1024? But this algorithm, if the new UTXO is more than 1024 bytes (such as the amount of new UTXO is 0.00000001GAS), whether the constructed transaction will still be unable to send.

jsolman: You are right, but in that extreme case people can collapse their gas utxo’s first.

jsolman: In Neo 3.0 we GAS won’t be a UTXO based asset anyway.

superboyiii: @jsolman Yes, agree with you. Just fixed it.

superboyiii:
I think it should not only applied on [deploy] and [invoke] smart contract but all Invocation Transaction or Contract Transaction. Since SimplePolicy will check all kinds of transactions whether main net is checkout to v2.10.0 or further. But I think there'll not be 0.001 gas on initial while Contract Transaction. I'll work on this.

PR #327, opened 2019-03-20
Shutdown Gracefully on Kill or Interrupt.

jsolman:
With this PR Interrupt due to `Control+C` or process kill will properly be caught and cause the node to be stopped in the same way as typing `exit`.


erikzhang:
In fact, I prefer `Control+C` to kill the process directly. Sometimes when debugging, you will want to use `Control+C` when you want to close the process immediately.

jsolman:
It still generally closes in less than 1 second. It is a best practice to shutdown cleanly.

vncoelho:
I also like this, Erik...hauajaha
ctrl+c is the power. 

But I think @jsolman wants to avoid crashes.

jsolman:
> Sometimes when debugging, you will want to use Control+C when you want to close the process immediately.

@erikzhang 
You can send a more forceful type of signal if you want to kill it immediately.

shargon:
A lot of red message appears before close

image

Maybe we could avoid this to prevent "scares"

jsolman:
@shargon I was looking for the best way to do it. I haven’t found a trivial way yet since we don’t have a top level actor.

jsolman:
I’ll see if I can find a way to get rid of them also in this PR after I finish the consensus PR I’m working on now

shargon:
Any news @jsolman ?

vncoelho:
I think that maybe this patch can come to neo-cli 3.0, however, let's not modify 2.x.

Is there something else to be done here, @jsolman?

Erik, I also prefer ctrl+c to kill direct. However, in the last core developers meeting I think that the majority agreed that we should avoid crashes...aehauhuea

jsolman:
There is no reason this should not be on both branches IMO.

vncoelho:
neo-cli 2.x can historically stay with its behavior, this will not really impact in anything critical.

shargon:
conflicts

lock9:
Hi @jsolman, could you solve the conflicts? Do you want us to solve it for you?

vncoelho:
@jsolman, when resolving conflicts I did not include this:
```cs
<<<<<<< GracefulExitOnKillOrInterrupt
            ReadLine();
            stop = true;
            task.Wait();
=======
```

neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -267,9 +296,28 @@ public void Run(string[] args)
...
+
+            return readLineTask.Result;
+        }
+
         private void RunConsole()
         {
-            bool running = true;
+            _running = true;
+            // Register sigterm event handler
+            AssemblyLoadContext.Default.Unloading += SigTermEventHandler;
```
shargon: We should move this to OnStart

shargon:
@igormcoelho this works well for you in linux?

shargon:
@vncoelho with my patch is working too?

jsolman:
Thanks for rebasing.

vncoelho:
I did not test yet, @shargon. I want to finish the P2P...aheuahueaea

vncoelho:
@shargon, In fact, ctrl+c has been always working on linux.

Just failing on that other PR that you made and closed...aehauheuaea ajajaja
I would need to test this patch.

But is this an agreement that this change is good?
I got some `Akka` errors after killing it with ctrl+c. But the same error was obtained when typing exit.
I think it happens when we kill it very soon, before it really initialized.

vncoelho:
@shargon, @jsolman and @erikzhang,

This works nice and ok with `ctrl+c`.

However, if the client is still killed very fast (just after initialization), this error is still obtained:

```cs
[ERROR][08/19/2019 19:37:29][Thread 0004][akka://NeoSystem/system/IO-TCP/$a] Monitored actor [[akka://NeoSystem/user/$b#1681348270]] terminated
Cause: Akka.Actor.DeathPactException: Monitored actor [[akka://NeoSystem/user/$b#1681348270]] terminated
   at Akka.Actor.ActorBase.Unhandled(Object message)
   at Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message)
   at Akka.Actor.ActorCell.ReceiveMessage(Object message)
   at Akka.Actor.ActorCell.AutoReceiveMessage(Envelope envelope)
   at Akka.Actor.ActorCell.Invoke(Envelope envelope)
```

shargon:
In my windows works well, but this exceptions is something to fix in other PR, i think

vncoelho:
@shargon, I prefer to wait for @erikzhang or @superboyiii approval.

vncoelho:
@Tommo-L, can you give a look at this?
#469

Tommo-L:
> @Tommo-L, can you give a look at this?
> #469

I think it also can fix the `exit` problem. I'll close #469

vncoelho:
image

That is why I did not want to merge it and told to wait for @erikzhang review's.
I think it should be reverted. It is quite bad as it is. Application is never finished when errors are presented, during tests and experimental things.

jsolman:
The red messages on Akka are just notifying the actors shutdown. A different top level actor needs to be added to avoid them.

I have run this on Linux a long time and never saw the hang you are describing @vncoelho. What are the steps you are using to reproduce the hang? What hardware are you running on? What flavor of Linux?

PR #329, opened 2019-03-21
Add the calculation of extra gas fee for "send" function.

superboyiii:
Add the calculation of extra gas fee for "send" function.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -977,7 +977,7 @@ namespace Neo.Shell
 
                 if (tx.Size > 1024)
                 {
-                    fee += Fixed8.FromDecimal(tx.Size * 0.00001m);
+                    fee += Fixed8.FromDecimal(tx.Size * 0.00001m + 0.001m);
                     tx = Program.Wallet.MakeTransaction(null, new[]
                     {
                         new TransferOutput
@@ -994,7 +994,7 @@ namespace Neo.Shell
                     Console.WriteLine("Insufficient funds");
                     return true;
                 }
-                
+                 
             }
             ContractParametersContext context = new ContractParametersContext(tx);
             Program.Wallet.Sign(context);
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -974,11 +974,28 @@ namespace Neo.Shell
                         ScriptHash = scriptHash
                     }
                 }, fee: fee);
+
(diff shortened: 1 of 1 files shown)


neo-cli/Shell/MainService.cs
```diff
@@ -974,11 +974,27 @@ private bool OnSendCommand(string[] args)
                         ScriptHash = scriptHash
                     }
                 }, fee: fee);
+
+                if (tx.Size > 1024)
+                {
+                    fee += Fixed8.FromDecimal(tx.Size * 0.00001m);
```
jsolman: The fee may make the transaction even larger by some bytes better add .0005 more in addition to this to ensure it should be enough in most cases.

jsolman: I guess it only needs it if the fee currently is less than .0005 though.

superboyiii: Yes, might be larger. But eh... I think 1024*0.00001=0.01024 gas is enough to be the headroom for the check of SimplePolicy. A little confused about why it should be 0.0005 for more additional gas....

jsolman: At first glance I thought this was using `.000001`*size. In this case I think this is using too much headroom actually. It should be able to use `FeePerExtraByte` the policy uses multiplied size + ensuring .0005 additional for gas utxo’s.

superboyiii: Oh! I see! Yes, should be an initial gas for  little overstep.

jsolman: Yeah, actually my last message was a mistake. I'll delete it.

jsolman: Yes, the policy does indeed use `.00001` fee per extra byte, so it may need that little bit extra for adding the gas :-)

superboyiii: Yes, the extra input+output should be the length as (34+60=94)byte. 94*0.00001=0.00094 so that 0.001 gas is actually a reasonable value to be the initial fee.

erikzhang:
Maybe we should take the `fee` parameter as `max_fee`. The automatically calculated `fee` should not exceed `max_fee`, otherwise the caller cannot estimate the cost.

jsolman:
@erikzhang this is a good idea. The caller should probably specify their desired fee and a max fee though, because they may want to pay more than the minimum required fee. Perhaps an additional max_fee can be added and it will fail to send and provide and error if it would need more than the max fee.

PR #340, opened 2019-04-04
Fix neo changes

shargon:
Currently neo-cli can't compile with the current neo master branch, with this patch you are able to do that.


erikzhang:
neo-cli is not going to compile with neo master branch.

shargon:
If neo-cli is not synchronized with the neo changes, i was not able to debug the whole chain, and in the most cases is necessary.

PR #346, opened 2019-04-19
Fix/invoke error

KickSeason:
1. wallet tx null check 
2. chose to relay.
3. change ReadPassword to ReadUserInput with safe mode.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -252,8 +252,15 @@ namespace Neo.Shell
                 Console.WriteLine("Engine faulted.");
                 return true;
             }
-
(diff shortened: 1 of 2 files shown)


neo-cli/Shell/MainService.cs
```diff
@@ -252,8 +252,15 @@ private bool OnInvokeCommand(string[] args)
                 Console.WriteLine("Engine faulted.");
                 return true;
             }
-
+            if (ReadUserInput("relay tx(no|yes)", false) != "yes") {
```
shargon: Maybe this check after the tx==null?

KickSeason:
ReadUserInput use unsafe mode as default

PR #351, opened 2019-04-24
Allow to install plugins from file (if exists)

shargon:


erikzhang:
Why are there so many "changes" that have not changed?

shargon:
I think that was the EOL, I will check my new installation of VS 2019. I think that github solved the issue when i resolved the conflicts inside the browser

PR #357, opened 2019-05-16
3.0.0-CI00044

erikzhang:


vncoelho:
A historical achievement. Congratulations, Erik.

neo-cli/Shell/MainService.cs
```diff
@@ -142,44 +145,27 @@ private bool OnBroadcastCommand(string[] args)
...
-            var tx = LoadScriptTransaction(
+            byte[] script = LoadDeploymentScript(
                 /* filePath */ args[1],
-                /* paramTypes */ args[2],
-                /* returnType */ args[3],
-                /* hasStorage */ args[4].ToBool(),
-                /* hasDynamicInvoke */ args[5].ToBool(),
-                /* isPayable */ args[6].ToBool(),
-                /* contractName */ args[7],
-                /* contractVersion */ args[8],
```
shargon: all of this metadata, will be inside the smart contract?

neo-cli/Shell/MainService.cs
```diff
@@ -815,19 +730,8 @@ private bool OnListAddressCommand(string[] args)
         private bool OnListAssetCommand(string[] args)
         {
             if (NoWallet()) return true;
-            foreach (var item in Program.Wallet.GetCoins().Where(p => !p.State.HasFlag(CoinState.Spent)).GroupBy(p => p.Output.AssetId, (k, g) => new
-            {
```
shargon: we will add this in the future?

neo-cli/Shell/MainService.cs
erikzhang: They will be in manifest.

neo-cli/Shell/MainService.cs
erikzhang: Did not find a good way to list all NEP-5 assets.

shargon: Maybe we should iterate all contracts, and try to find it according to NEP10?

shargon: Or maybe we should add a new command for wallets, add nep5 token, and only look if you append it before?

erikzhang: I will build this function into `Wallet`.

shargon:
When https://github.com/neo-project/neo/pull/753 will be integrated, i can approve it

erikzhang:
@shargon It will be. We are going to test it before approving.

erikzhang:
I will draft a preview release later (after the pricing model). I think we can merge this first. @shargon @vncoelho

neo-cli/neo-cli.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Copyright>2016-2019 The Neo Project</Copyright>
     <AssemblyTitle>Neo.CLI</AssemblyTitle>
-    <Version>3.0.0-preview1</Version>
+    <Version>2.10.1</Version>
```
shargon: Is not 3?

erikzhang: After we are ready for release the preview, I will increase the version number.

shargon:
Maybe we need to solve the test error before merge

erikzhang:
The error is because of the wrong uploaded package on the NuGet. There is no way to solve it now.

PR #358, opened 2019-05-16
Fix protocols

shargon:
LowPriorityThreshold is long


erikzhang:
See https://github.com/neo-project/neo/pull/753

PR #372, opened 2019-05-23
Fix uninstall CoreMetrics (3x)

shargon:
Close https://github.com/neo-project/neo-cli/issues/366


shargon:
Is not possible to fix without travis

PR #373, opened 2019-05-25
3.0.0-CI00057

erikzhang:


neo-cli/Shell/MainService.cs
```diff
@@ -82,8 +75,6 @@ protected override bool OnCommand(string[] args)
                     return OnOpenCommand(args);
                 case "close":
                     return OnCloseCommand(args);
-                case "rebuild":
```
vncoelho: bye....

vncoelho:
What are we waiting for merging?
Could we merge and later update?

PR #375, opened 2019-05-29
Don't hardcode System.Text.Encoding.Unicode (which is UTF-16LE) as output encoding

intelfx:
GNU/Linux systems generally use UTF-8, not UTF-16.

Using UTF-16 may "appear" to work with ASCII characters because NULs are silently skipped by terminal emulators, but this causes problems when scripting neo-cli.

For now, I moved this configuration under Win32 condition, but I don't know what is the right thing to do with Win32. Maybe it makes sense to remove this statement altogether.

### DIFF
--- a/neo-cli/Services/ConsoleServiceBase.cs
+++ b/neo-cli/Services/ConsoleServiceBase.cs
@@ -272,8 +272,10 @@ namespace Neo.Services
             bool running = true;
             string[] emptyarg = new string[] { "" };
             if (Environment.OSVersion.Platform == PlatformID.Win32NT)
+            {
                 Console.Title = ServiceName;
-            Console.OutputEncoding = Encoding.Unicode;
+                Console.OutputEncoding = Encoding.Unicode;
+            }
 
             Console.ForegroundColor = ConsoleColor.DarkGreen;
             Console.WriteLine($"{ServiceName} Version: {Assembly.GetEntryAssembly().GetVersion()}");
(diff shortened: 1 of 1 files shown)


erikzhang:
Maybe we can set it to UTF-8. I will test it under Windows.

intelfx:
@erikzhang The idea is not to hardcode it, either to UTF-16 or UTF-8 or ASCII or anything, but rather to stick with the platform default (i. e. not set it at all) unless there is a good reason to do otherwise.

Could you please drop this line completely and check it under Windows?

erikzhang:
I've removed it. Can you test it on linux?

shargon:
Yes, is better to remove it

intelfx:
@erikzhang Everything under `if (Environment.OSVersion.Platform == PlatformID.Win32NT)` doesn't affect Linux anyway, so it's fine.

shargon:
@intelfx but the change is outside this conditional, so it affect to linux as well

intelfx:
@shargon The initial change (that moves `Console.OutputEncoding = Encoding.Unicode;` inside the conditional) was submitted by me — of course I tested it on Linux, because fixing behavior under Linux was the motivation for this change in the first place.

The second change submitted by @erikzhang removes the statement from inside the conditional, so it can only affect Win32 platforms. That's what I asked him to test as I don't have access to Windows.

shargon:
This change affect linux because previously was outside the conditional

image

intelfx:
Oh. Sorry. I misread the diff.

erikzhang:
Need someone to test it on linux. I've tested it on windows.

intelfx:
Yes, it works on Linux, at least in a Docker container with .NET. Didn't test natively, but I think it'll be okay.

PR #376, opened 2019-05-29
Fix uninstall plugin (2x)

shargon:
image


superboyiii:
There's a conflict. I fixed this in https://github.com/neo-project/neo-cli/pull/452

PR #377, opened 2019-05-29
Fix uninstall plugin (3x)

shargon:
Same as https://github.com/neo-project/neo-cli/pull/376 but for neo3


shargon:
@erikzhang  Could you merge it, without travis i can't

PR #378, opened 2019-06-01
allow free tx as default on invocations

igormcoelho:
I think that it's not necessary that neo-cli enforces a minimum fee of 0.001 on all transactions, only on those that exceed 1024 bytes.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -283,7 +283,7 @@ namespace Neo.Shell
 
         public InvocationTransaction DecorateInvocationTransaction(InvocationTransaction tx)
         {
-            Fixed8 fee = Fixed8.FromDecimal(0);
+            Fixed8 fee = Fixed8.Zero;
 
             if (tx.Size > 1024)
             {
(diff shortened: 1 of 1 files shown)


shargon:
give me a couple of hours for check the gas consumption

shargon:
https://github.com/neo-project/neo-plugins/blob/master/SimplePolicy/SimplePolicyPlugin.cs#L102 Good :)

neo-cli/Shell/MainService.cs
```diff
@@ -283,10 +283,11 @@ private bool OnInvokeCommand(string[] args)
 
         public InvocationTransaction DecorateInvocationTransaction(InvocationTransaction tx)
         {
-            Fixed8 fee = Fixed8.FromDecimal(0.001m);
+            Fixed8 fee = Fixed8.FromDecimal(0);
```
shargon:
```suggestion
            Fixed8 fee = Fixed8.Zero;
```

igormcoelho: Committed change.

vncoelho:
How is this going? Can you test this, @shargon ? I can set the environment for testing here as well.

igormcoelho:
Thanks for testing @shargon, I'll try to improve testing capabilities of neo-cli in a next commit.

igormcoelho:
@erikzhang a more complete testing structure is proposed here: https://github.com/neo-project/neo-cli/pull/382

PR #379, opened 2019-06-03
Some fixes (2x)

shargon:


neo-cli/Shell/MainService.cs
```diff
@@ -704,6 +715,19 @@ private bool OnImportKeyCommand(string[] args)
             catch (FormatException) { }
             if (prikey == null)
             {
+                var file = new FileInfo(args[2]);
+
+                if (!file.Exists)
+                {
+                    Console.WriteLine($"Error: File '{args[2]}' doesn't exists");
```
shargon:
```suggestion
                    Console.WriteLine($"Error: File '{file.FullName}' doesn't exists");
```

vncoelho:
Is it ready, Shargon?

shargon:
Ready

PR #382, opened 2019-06-03
Tests for OnInvoke neo-cli

igormcoelho:
I guess we can separate a few methods to guarantee better testability directly on neo-cli project.


igormcoelho:
@erikzhang @shargon this effectively creates a template which we can easily use for future testings.

tests/protocol.json
```diff
@@ -0,0 +1,30 @@
+{
+  "ProtocolConfiguration": {
```
erikzhang: This file could be removed.

tests/.gitignore
```diff
@@ -0,0 +1,2 @@
+bin/
+obj/
```
erikzhang: This file could be removed.

neo-cli/Shell/MainService.cs
```diff
@@ -184,9 +184,9 @@ private bool OnDeployCommand(string[] args)
             return SignAndSendTx(tx);
         }
 
-        private bool OnInvokeCommand(string[] args)
+        internal protected void ParseParametersInvoke(string[] args, out UInt160 scriptHash, out ContractParameter[] parameters)
```
erikzhang: Why we need this?

igormcoelho: I separated OnInvokeCommand in two parts, to allow testing of this method (without getting to the last part of tx relay). This allowed to check that parameters were being parsed, and that transaction was being built correctly.
Still, I agree it's kind of ugly that we cannot fully test the complete thing (OnInvokeCommand). I think I just got a better idea.

igormcoelho:
@erikzhang @shargon this is the complete template for tests, I just removed any other extra change.
I have some ideas for good testing here, but I guess we should start applying these on Neo 3.0 first (neo-cli 3.x).

.gitignore
```diff
@@ -1,288 +0,0 @@
-## Ignore Visual Studio temporary files, build results, and
```
vncoelho: Can this .gitignore be really removed?

igormcoelho: did I remove something? I'll check.

igormcoelho: @vncoelho

PR #390, opened 2019-06-15
Fix 293

shargon:
Close https://github.com/neo-project/neo-cli/issues/293


erikzhang:
We can check if the file exists.

PR #391, opened 2019-06-16
Ping nodes while show state

shargon:
Related to https://github.com/neo-project/neo/pull/832
Close https://github.com/neo-project/neo-cli/issues/287

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -973,8 +973,7 @@ namespace Neo.Shell
             });
             Console.ReadLine();
             cancel.Cancel();
-            task.Wait();
-            broadcast.Wait();
+            Task.WaitAll(task, broadcast);
             Console.WriteLine();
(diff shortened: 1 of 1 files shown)


neo-cli/Shell/MainService.cs
```diff
@@ -934,6 +934,14 @@ private bool OnShowStateCommand(string[] args)
             bool stop = false;
             Console.CursorVisible = false;
             Console.Clear();
+            Task broadcast = Task.Run(async () =>
+            {
+                while (!stop)
+                {
+                    system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(Blockchain.Singleton.Height)));
+                    await Task.Delay(Blockchain.TimePerBlock);
```
erikzhang: We should use `CancellationToken` here. Otherwise we will have to wait at most 15 seconds to exit from the loop.

PR #393, opened 2019-06-18
Fix import multisigaddress method on solo chain

superboyiii:
Don't need at least 5 elements for the command of import multisigaddress on solo chain.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -672,9 +672,9 @@ namespace Neo.Shell
         {
             if (NoWallet()) return true;
 
-            if (args.Length < 5)
+            if (args.Length < 4)
             {
-                Console.WriteLine("Error. Use at least 2 public keys to create a multisig address.");
+                Console.WriteLine("Error. Invalid parameters.");
                 return true;
             }
 


superboyiii:
That's your works, hahaha!  [#703](https://github.com/neo-project/neo/issues/703)

vncoelho:
Aehuaheuaheaea, I got it...

I never used the neo-cli for multi-sig. I will build the client and test, then approve it. Maybe it take 1-3 days until I can focus on it, but I will check it out.

If someone could test it before.

superboyiii:
On master branch (NEO3.X), it's already applied on. I just sync it to the latest.

superboyiii:
I've tested it before on NEO3.0.

vncoelho:
Sorry for the delay. Testing it now.

vncoelho:
even the example in the docs is showing error:

https://docs.neo.org/en-us/node/cli/cli.html

`import multisigaddress 1 037ebe29fff57d8c177870e9d9eecb046b27fc290ccbac88a0e3da8bac5daa630d 03b34a4be80db4a38f62bb41d63f9b1cb664e5e0416c1ac39db605a8e30ef270cc
`

superboyiii:
> Updated, it is working, I created a wallet and forgot the extension, then has not actually been created.
> 
> A wallet opened was still missing.
> 
> However, the `error` message should be improved because it was not reporting any wallet open issue.
> It was probably reaching `wallet.Save();` because the default wallet was not NEP6.

Yes, that's correct. We should approve the 'error' message to help developers know the wrong point.

PR #399, opened 2019-06-21
Show multi sig address in `list address command

shargon:


neo-cli/Shell/MainService.cs
```diff
@@ -739,7 +739,18 @@ private bool OnListAddressCommand(string[] args)
             if (NoWallet()) return true;
             foreach (Contract contract in Program.Wallet.GetAccounts().Where(p => !p.WatchOnly).Select(p => p.Contract))
             {
-                Console.WriteLine($"{contract.Address}\t{(contract.Script.IsStandardContract() ? "Standard" : "Nonstandard")}");
+                var type = "Nonstandard";
```
vncoelho: Maybe add "Nonstandard-deployed" and "Nonstandard-verification-only" ?

shargon: for watch only?

shargon: Take a look

PR #410, opened 2019-07-18
Dockerfile for running neo-cli

ksemaev:
This is an example of dockerfile I use in kubernetes. This one is tweaked a little for plain docker or docker-compose use.

The difference with kuber is in custom start-script

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,11 +42,18 @@ RUN wget -q https://github.com/neo-project/neo-cli/releases/download/$VERSION/ne
 #RUN chmod a+x /usr/local/bin/start-script.sh
 #ENTRYPOINT ["/bin/bash", "/usr/local/bin/start-script.sh"]
 
-WORKDIR /home/neo
 VOLUME /home/neo/data/
-USER neo
+WORKDIR /home/neo/data
+
+#####  In kubernetes volumes are attached with predefined UID. In plain docker it's a bit harder to achieve, so this one is commented out
+#USER neo
 
 EXPOSE 10332
 EXPOSE 10334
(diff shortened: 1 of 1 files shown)


ksemaev:
@lock9 I have absolutely no idea about this dockerfile that travis is testing: https://github.com/neo-project/neo-cli/blob/master/ci/Dockerfile

I just suggested to use this one: https://github.com/ksemaev/neo-cli/blob/dockerfile/Dockerfile

lock9:
Sorry @ksemaev ! I saw that the error was related to Docker, that is why I thought it was caused by you. I guess we need to fix travis before.

vncoelho:
Hi @ksemaev, thanks for this PR.

I am not sure if this needs to be merged here, because we already have this simply docker here: https://github.com/neo-project/neo-cli/blob/master/ci/Dockerfile
What do you think?

About the kubernets, I remember we discussed that in another thread. It would be good to have all the guidelines for everybody to run that. But maybe this could come in the docs. What do you think?

Maybe the docker file and all kubernets setup come be described around here: https://docs.neo.org/docs/en-us/node/cli/setup.html

vncoelho:
@ksemaev, if that proceed, please close this PR and open a detailed description here: https://github.com/neo-project/docs

That would surely be a guidelines for many people.
If that goes well we could also update the README of this project pointing to the new description.

vncoelho:
@ksemaev, as in the previous comment, I believe that we already have a similar docker.
But it would be great to have a complete guidelines for kubernets on the docs.

Let's leave it closed, but fell free to reopon.

PR #422, opened 2019-08-02
Set the net to main by default

Liaojinghui:
Set the net to main by default in case the user does not set that param.

### DIFF
--- a/neo-cli/Settings.cs
+++ b/neo-cli/Settings.cs
@@ -2,6 +2,7 @@
 using Neo.Network.P2P;
 using Neo.SmartContract.Native;
 using System.Net;
+using System.Threading;
 
 namespace Neo
 {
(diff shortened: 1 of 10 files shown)


neo-cli/Settings.cs
```diff
@@ -13,12 +14,30 @@ internal class Settings
...
-        static Settings()
+        static bool UpdateDefault(IConfiguration configuration)
         {
-            IConfigurationSection section = new ConfigurationBuilder().AddJsonFile("config.json").Build().GetSection("ApplicationConfiguration");
-            Default = new Settings(section);
+            var settings = new Settings(configuration.GetSection("ApplicationConfiguration"));
+            return null == Interlocked.CompareExchange(ref _default, settings, null);
+        }
+
+        public static bool Initialize(IConfiguration configuration)
```
shargon: Why this are changed?

Liaojinghui: This change is actually your previous modification, not my commit.  https://github.com/neo-project/neo-cli/pull/415/commits/626b1c8b319d735f7d880bda696306e4e52c1cd5

shargon: But why appear?

Liaojinghui: Guess this pull request is regarded as the child of my previous one, see, even the commit number is 11.

Liaojinghui: Figured out, I didn't merge the latest update of this repo, fixed now.

shargon:
Why we need `config.json` now?

Liaojinghui:
> Why we need `config.json` now?

I don't think it is necessary.

> Why we need `config.json` now?

Just reviewed https://github.com/neo-project/neo/blob/master/neo/ProtocolSettings.cs feels like there is totally no need to add `--mainnet` command

Liaojinghui:
Please ignore this pull request, the original code is already sound enough to handle all condition.

PR #433, opened 2019-08-13
Using project reference

lock9:
Referencing project instead of nuget

### DIFF
--- a/neo-cli/neo-cli.csproj
+++ b/neo-cli/neo-cli.csproj
@@ -32,6 +32,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\neo\neo.csproj" />
+    <ProjectReference Include="..\..\neo\neo\neo.csproj" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


Ashuaidehao:
Do you mean to add neo as a submodule? Only reference neo with your local machine's relative path won't work in others machine.

shargon:
I prefer with submodules too

neo-cli/neo-cli.csproj
```diff
@@ -28,8 +28,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Neo" Version="3.0.0-CI00130" />
     <PackageReference Include="System.ServiceProcess.ServiceController" Version="4.5.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\..\neo\neo\neo.csproj" />
```
vncoelho: This works, @lock9, since they are under the same `github` account?

shargon:
it works but is a little bit weird, i think that submodules will be better here

PR #443, opened 2019-08-21
Improve select menu

superboyiii:
Improve select menu for choosing "y" or "yes".


superboyiii:
Agree. Also exception message.

PR #444, opened 2019-08-21
Fix invoke failed

superboyiii:
Fix invoke failed.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -191,7 +191,7 @@ namespace Neo.Shell
                 Console.WriteLine($"Invoking script with: '{tx.Script.ToHexString()}'");
             }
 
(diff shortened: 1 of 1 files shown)


neo-cli/Shell/MainService.cs
```diff
@@ -191,7 +191,7 @@ private bool OnInvokeCommand(string[] args)
                 Console.WriteLine($"Invoking script with: '{tx.Script.ToHexString()}'");
             }
 
-            ApplicationEngine engine = ApplicationEngine.Run(tx.Script, tx);
+            ApplicationEngine engine = ApplicationEngine.Run(tx.Script, tx, extraGAS: 100_000_000);
```
shargon: Maybe we should use `testMode=true` ?

superboyiii: Agree.

PR #448, opened 2019-08-26
Check dotnet format in travis and adding .editorconfig

vncoelho:


shargon:
I don't know why we use here docker for travis, for me are different things, i prefer use travis without the docker file

vncoelho:
I agree. 
Lets rethink this.

lock9:
@shargon  any ideas why this PR is being blocked by Travis?

shargon:
because it doesn't have the dotnet framework

PR #451, opened 2019-08-27
Allow neo-cli to switch between RocksDB and LevelDB

shargon:
Require changes when https://github.com/neo-project/neo/pull/1063 was merged

If you want to test RocksDB PR, you can use this branch

### DIFF
--- a/neo-cli.sln
+++ b/neo-cli.sln
@@ -1,9 +1,11 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26430.15
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29215.179
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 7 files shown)


shargon:
Yes, we can do real benchmarks between them now

neo-cli.sln
```diff
@@ -1,22 +1,31 @@
-﻿
```
shargon: This changes will be removed on the last version

neo-cli/neo-cli.csproj
```diff
@@ -28,8 +28,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Neo" Version="3.0.0-CI00151" />
     <PackageReference Include="System.ServiceProcess.ServiceController" Version="4.5.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\..\neo\neo\neo.csproj" />
+  </ItemGroup>
```
shargon: This changes will be removed on the last version

PR #452, opened 2019-08-29
Fix inconsistency in plugin name

superboyiii:
Fix inconsistency between Plugin name and Plugin class name. For example: ```SimplePolicy``` is using ```SimplePolicyPlugin``` as class name, but using ```SimplePolicy``` as plugin name. This issue also could be seen on ```RpcSystemAssetTracker```. I think it's also necessary to unify plugin class name on neo3 to avoid this issue.


neo-cli/Shell/MainService.cs
```diff
@@ -1193,7 +1193,7 @@ private bool OnUnInstallCommand(string[] args)
 
             var pluginName = args[1];
 
-            if (!Plugin.Plugins.Any(u => u.Name == pluginName))
+            if (!Plugin.Plugins.Any(u => u.Name.Contains(pluginName)))
```
erikzhang: I think that it will fail in the next steps.

superboyiii: Why? It works well on mine.
image
image
image
image
image

superboyiii: If in previous version, if input ```uninstall RpcSystemAssetTracker```, it won't pass this check since it need plugin name and plugin class name be the same. If I input ```uninstall RpcSystemAssetTrackerPlugin```, it could pass this check but can't find the folder and file. That's all because of this difference. Like this:
image
image
image
image
image
image

superboyiii:
> But the name of the class is used?

Hi, Shargon. What do you mean?

neo-cli/Shell/MainService.cs
shargon: try with uninstall "a"

superboyiii: If we have to match exactly, I think we should unified the class name and project name.

shargon:
For reproduce the error, what are the steps?

`
install plugin RpcSystemAssetTracker
uninstall plugin RpcSystemAssetTracker
`
?

superboyiii:
> For reproduce the error, what are the steps?
> 
> `install plugin RpcSystemAssetTracker uninstall plugin RpcSystemAssetTracker`
> ?

Now only ```SimplePolicy``` and ```RpcNep5Tracker``` have this issue. 
For example ``` install SimplePolicy```, ```uninstall SimplePolicy```.

shargon:
Check my last changes, it works for me.

shargon:
mmmm... i see two solutions, as you said. 

- 1 Remove `plugin.Name`, and use the library name for the plugin
  - This avoid create two plugins in the same library
- 2 Use a file like `plugins.json` for store what plugins are installed or not.
  - This help when we need to remove hot-plugins (currently used)

superboyiii:
> mmmm... i see two solutions, as you said.
> 
> * 1 Remove `plugin.Name`, and use the library name for the plugin
>   
>   * This avoid create two plugins in the same library
> * 2 Use a file like `plugins.json` for store what plugins are installed.
>   
>   * This help when we need to remove hot-plugins (currently used)

I think solution 1 makes sense for making plugins more regular.

vncoelho:
@superboyiii, how is this PR? Should we merge this?

shargon:
1 is good for me

shargon:
If we want to remove the plugin name, this should be done in `neo` first

erikzhang:
@superboyiii Can you test it again?

superboyiii:
Sure

superboyiii:
Here's my test steps and result.

1. install Plugins
image
2. Plugins are loaded successfully
image
3. Uninstall plugins
image
4. Plugins are deleted successfully.
image
5. Rename one of the plugin class name, make it different as type name.
image
6. build it and install it
image
7. restart neo-cli and uninstall it.
image
8. Plugin is deleted successfully. Test Pass.
image
9. By the way, Neo v3.0.0-CI00815 isn't compatible as a dependency in neo-plugins, the code of plugins should be adjusted.
image
10. And if neo-cli is using dependency of Neo v3.0.0-CI00815, RpcNep5Tracker and SystemLog will break.
image
image
Only these are OK.
image

Summary: Test Pass. 
Warning: RpcNep5Tracker and SystemLog need adjust if apply the latest neo on neo-cli or neo-plugins.

PR #463, opened 2019-09-17
Update deploy command following NEO3 schema

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/91
Require https://github.com/neo-project/neo-devpack-dotnet/pull/109
Related to https://github.com/neo-project/neo-devpack-dotnet/issues/91

Should follow this logic
 https://github.com/neo-project/neo/blob/e1fdde307153879f0a7ae64bd3e11d042284896e/neo/SmartContract/InteropService.NEO.cs#L248

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -222,7 +222,7 @@ namespace Neo.Shell
         {
             if (string.IsNullOrEmpty(manifestFilePath))
             {
-                manifestFilePath = Path.ChangeExtension(nefFilePath, ".manifest" + Path.GetExtension(nefFilePath));
+                manifestFilePath = Path.ChangeExtension(nefFilePath, ".manifest.json");
             }
 
             // Read manifest
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -142,7 +142,7 @@ namespace Neo.Shell
             if (NoWallet()) return true;
             byte[] script = LoadDeploymentScript(
                 /* filePath */ args[1],
-                /* manifest */ args.Length >= 2 ? "" : args[2],
+                /* manifest */ args.Length == 2 ? "" : args[2],
                 /* scriptHash */ out var scriptHash);
 
             Transaction tx;
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -227,6 +227,7 @@ namespace Neo.Shell
             {
                 throw new ArgumentException(nameof(manifestFilePath));
             }
+
             var manifest = ContractManifest.Parse(File.ReadAllText(manifestFilePath));
 
             // Read nef
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -269,7 +269,7 @@ namespace Neo.Shell
 
                                 if (!InteropService.SupportedMethods().ContainsKey(ci.TokenU32))
                                 {
-                                    throw new FormatException($"Syscall not found {ci.TokenU32}. Are you using a NEO2 smartContract?");
+                                    throw new FormatException($"Syscall not found {ci.TokenU32.ToString("x2")}. Are you using a NEO2 smartContract?");
                                 }
                                 break;
                             }
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -142,7 +142,7 @@ namespace Neo.Shell
             if (NoWallet()) return true;
             byte[] script = LoadDeploymentScript(
                 /* filePath */ args[1],
-                /* manifest */ args[2],
+                /* manifest */ args.Length == 2 ? "" : args[2],
                 /* scriptHash */ out var scriptHash);
 
             Transaction tx;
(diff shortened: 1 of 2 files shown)


superboyiii:
> Require [neo-project/neo-devpack-dotnet#109](https://github.com/neo-project/neo-devpack-dotnet/pull/109)
> Related to [neo-project/neo-devpack-dotnet#91](https://github.com/neo-project/neo-devpack-dotnet/issues/91)
> 
> Should follow this logic
> https://github.com/neo-project/neo/blob/e1fdde307153879f0a7ae64bd3e11d042284896e/neo/SmartContract/InteropService.NEO.cs#L248

Great! Let me have a look.

superboyiii:
Hi, Shargon, I've made a test based on my simple sc. The content is：
```using Neo.SmartContract.Framework.Services.Neo;
using System;


namespace Neo.SmartContract
{
    public class ICO_Template : Framework.SmartContract
    {
        //Token Settings
        public static string Name() => "ZZH";
        public static string Symbol() => "ZZH";


        public static Object Main(string operation, params object[] args)
        {
            if (Runtime.Trigger == TriggerType.Application)
            {

                if (operation == "name") return Name();
                if (operation == "symbol") return Symbol();

            }
            return false;
        }
    }
}
```
It was compiled successfully within neon.exe
image
You could see there're manifest in the root path.
image
Then I deploy it on neo-cli-3.0.0-preview1, meet an issue on invoking ```ContractMethodDescriptor```.
image
image
image

shargon:
Thanks for the test, i will review it

shargon:
You are right, we need some changes included in https://github.com/neo-project/neo-devpack-dotnet/pull/93 , i will copy these fixes

neo-cli/Shell/MainService.cs
```diff
@@ -232,14 +243,48 @@ private byte[] LoadDeploymentScript(string nefFilePath, bool hasStorage, bool is
...
+
+                    switch (ci.OpCode)
+                    {
+                        case OpCode.SYSCALL:
+                            {
+                                // Check bad syscalls (NEO2)
+
+                                if (!InteropService.SupportedMethods().ContainsKey(ci.TokenU32))
+                                {
+                                    throw new FormatException($"Syscall not found {ci.TokenU32.ToString("x2")}. Are you using a NEO2 smartContract?");
```
vncoelho: Is this needed, @shargon?

The compiler is going to be different.
Maybe we can just say that the syscall is not valid. I think there is not need to really explicit that it is NEO2.

superboyiii: @vncoelho Hey, my friend, since we've not deployed a new nuget for NEO3 SmartContract Framework, and the NEO3 one is also using the version number ```2.9.3```, it can make developers greatly confused, including me, hahaha. So I think this error message is necessary because if you're using the NEO2 nuget of Framework 2.9.3 to deploy a sc on NEO3, it's incompatible.

shargon: Yes, If you use the old version of SmartContract.Framework, you can get a wrong NEF, so I think that this checks could be very useful. Is possible that some invalid contract pass, it is a basic filter for prevent some human errors.

lock9:
@shargon  we just have to replace ".nef" with "manifest.json", or equivalent (I don't recall the exact file extension)

shargon:
Is not tge expected behavior , tell me one linux command that works as you said?

lock9:
@shargon  all of them, you tell me a program that you have to point to all files 😂 
What if it is optional? If you don't send any, it will just use the same file name with a different extension?

shargon:
Then i don't understand you very well, you are asking for a file the first parameter and for a name in the second? give me an example please

neo-cli/Shell/MainService.cs
```diff
@@ -140,9 +140,9 @@ private bool OnBroadcastCommand(string[] args)
         private bool OnDeployCommand(string[] args)
         {
             if (NoWallet()) return true;
-            byte[] script = LoadDeploymentScript(
-                /* filePath */ args[1],
-                /* manifest */ args[2],
+            byte[] script = LoadDeploymentScript(
+                /* filePath */ args[1],
+                /* manifest */ args[2],
```
lock9: @shargon 
`manifestPath = filePath.replace(".nef", ".manifest.json")`
We don't need ` /* manifest */ args[2],`

shargon: I understand you something like this
`deploy c:\myNef.nef myManifest`

> c:\myNef.nef
> c:\myManifest.json

lock9: Only `deploy c:\myNef.nef`,  we can assume `myNef.manifest.json` exists

shargon:
@lock9 please review it again.

neo-cli/Shell/MainService.cs
```diff
@@ -220,6 +220,11 @@ private bool OnInvokeCommand(string[] args)
 
         private byte[] LoadDeploymentScript(string nefFilePath, string manifestFilePath, out UInt160 scriptHash)
         {
+            if (string.IsNullOrEmpty(manifestFilePath))
+            {
+                manifestFilePath = Path.ChangeExtension(nefFilePath, ".manifest" + Path.GetExtension(nefFilePath));
```
lock9: The extension is not json? It is not nef, is it?

shargon: Take a look again

superboyiii:
I'll let @nicolegys to have a test.

lock9:
@shargon 
image

Maybe the manifest is broken? I used the latest master and you can see that the file exists.

nicolegys:
Successfully deploy a smart contract without manifest.json
image

nicolegys:
In debug mode, it will throw an exception when I deploy a NEO2 smartContract.
image
But it will only return 'error' when I use a release version neo-cli.
image
Because of this:
image

image

Should we add 'Console.WriteLine' directly instead of throwing an exception?

shargon:
we can remove the Debug directive

PR #466, opened 2019-09-20
Revert initial fee 

superboyiii:
Revert initial fee because it's necessary for checking ```owner``` witness in ```deploy``` function of nep-5 contract unless adding attributes manually.


superboyiii:
https://github.com/neo-project/neo-cli/issues/465

PR #470, opened 2019-09-29
Printint error message when "UnlockWallet" wallet is not found

### LINKED ISSUE #471: An non-existent wallet which is set to config.json, but neo-cli doesn't throw any exception while start and open wallet automatically.
An not existed wallet which is set to config.json, but neo-cli doesn't throw any exception while start and open wallet automatically. 
For example, set the **path** in  **UnlockWallet** to an non-existent wallet.
image
Then I start neo-cli.
image
Then input "show state" command.
image
The "Index height" appears!!! It's too weird!

lock9:
Closes #471

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -1070,6 +1070,10 @@ namespace Neo.Shell
                 {
                     Program.Wallet = OpenWallet(Settings.Default.UnlockWallet.Path, Settings.Default.UnlockWallet.Password);
                 }
+                catch (FileNotFoundException)
+                {
+                    Console.WriteLine($"Warning: wallet file \"{Settings.Default.UnlockWallet.Path}\" not found.");
+                }
                 catch (CryptographicException)
                 {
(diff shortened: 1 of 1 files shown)


lock9:
@shargon that is a different error. When you use a wrong password, the message appears, it just doesn't appear when you are using a wallet that doesn't exist.

lock9:
@shargon  any questions?

shargon:
> When you use a wrong password, the message appears, it just doesn't appear when you are using a wallet that doesn't exist.

Yes as you said, is printed here 

https://github.com/neo-project/neo-cli/blob/7a7d961fbfa04a039c871311e1b35d82827c2451/neo-cli/Shell/MainService.cs#L825

We should unify the messages, you don't think so?

lock9:
@shargon  I think they are different, one is a startup function, the other is user drive command. I think it is ok the way it is, but let's wait for a second approval.

nicolegys:
Tested
image
@lock9

PR #474, opened 2019-10-16
Prevent Invalid ip format on connected in neo-cli

shargon:
Close https://github.com/neo-project/neo-cli/issues/473

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -1021,10 +1021,11 @@ namespace Neo.Shell
 
                     maxLines = Math.Max(maxLines, linesWritten);
 
-                    while (++linesWritten < maxLines)
+                    while (linesWritten < maxLines)
+                    {
                         WriteLineWithoutFlicker("", Console.WindowWidth - 1);
-
-                    maxLines = 0;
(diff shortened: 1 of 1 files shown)


neo-cli/Shell/MainService.cs
```diff
@@ -1004,20 +1004,28 @@ private bool OnShowStateCommand(string[] args)
...
 
-                    while (++linesWritten < Console.WindowHeight)
-                        WriteLineWithoutFlicker();
+                    maxLines = Math.Max(maxLines, linesWritten);
+
+                    while (linesWritten < maxLines)
+                    {
+                        WriteLineWithoutFlicker("", Console.WindowWidth - 1);
+                        maxLines--;
+                    }
```
superboyiii: Why this? ```maxlines``` will never smaller than ```linesWritten``` since ```maxLines = Math.Max(maxLines, linesWritten)```.

superboyiii: @shargon I think the format issue has been solved, but there still seems someting incorrect on remote display.
image

shargon: I will take a look

PR #426, opened 2019-08-07
Fix control+c

shargon:
Close https://github.com/neo-project/neo-cli/issues/425

PR #475, opened 2019-10-17
Revert "Shutdown Gracefully on Kill or Interrupt."

vncoelho:
Reverts neo-project/neo-cli#327

closes #476 

image

Application takes a lot of time to be finished (3 - 5 min). This is bad during developing and testing phase.

I think that ctrl+c should work properly (exiting quick), even corrupting files...aehauheauea
Exit should be the nice one.

### DIFF
--- a/neo-cli/Shell/MainService.cs
+++ b/neo-cli/Shell/MainService.cs
@@ -1128,6 +1128,10 @@ namespace Neo.Shell
                 {
                     Program.Wallet = OpenWallet(Settings.Default.UnlockWallet.Path, Settings.Default.UnlockWallet.Password);
                 }
+                catch (FileNotFoundException)
+                {
+                    Console.WriteLine($"Warning: wallet file \"{Settings.Default.UnlockWallet.Path}\" not found.");
+                }
                 catch (CryptographicException)
                 {
                     Console.WriteLine($"failed to open file \"{Settings.Default.UnlockWallet.Path}\"");
@@ -1304,6 +1308,11 @@ namespace Neo.Shell
 
         private static Wallet OpenWallet(string path, string password)
         {
+            if (!File.Exists(path))
+            {
+                throw new FileNotFoundException();
+            }
+
             if (Path.GetExtension(path) == ".db3")
             {
                 return UserWallet.Open(path, password);


vncoelho:
But I am not sure, if the majority wants to keep it like this...For me, I never like to wait, it makes no sense.
If a corruption database is detected we should manage that on the Database special plugin (leveldb, rocksdb).
For memory we can just finish the application.

lock9:
@vncoelho  please check #426

vncoelho:
Did you try to close when crashed the actors, as described?

superboyiii:
@vncoelho Here it sometimes closes in 1 seconds, sometimes takes about more than 10 seconds, seems not too different from the previous one on ```exit```.  But I don't have Debian OS. I tried it on Winodws10, CentOS 7.6, Ubuntu 16.04.

vncoelho:
@superboyiii, it is being terrible for linux users here. 

The client is never finishing, taking more than 1-2 minutes.
This should be revert asap or fixed.

lock9:
Hello @vncoelho, 
I will release docker images and a new testing process next Monday so this problem doesn't happen again. 
I don't know, but if the stable version is working, why should we revert this change?

Also: if we have an incompatibility between Windows and Linux, which version are we choosing to support? 
Again, sorry for the inconvenience, I will make sure next tests cover all supported platforms (Except Windows 7 😅)

shargon:
> I think that ctrl+c should work properly (exiting quick), even corrupting files...aehauheauea

No corruption, clean exit

shargon:
Could you introduce me the problem, i was in vacations

vncoelho:
@shargon, when we have errors, such as initialization `Akka` problems, the client is not closing.
It is taking more than 3-5min to exit the client, which is terrible.

Can you take a look at the comments of the issue?

vncoelho:
Talking about corruption, was there clear cases in which the database was getting corrupted? @lock9 @shargon

lock9:
Hi @vncoelho, on Windows, running on regular conditions, if you press control-c the program would hang and only be closed using a force close. Shouldn't we focus on having it working properly on the latest stable version? 
I didn't see if the database was being corrupted, but force closing neo can definitely cause an issue. 

One known problem that I had in the past, was the file lock. If you close it using force close, LeveDB would remain locked and wouldn't work until you restart your pc or do some equivalent "cleaning" 

TL;DR:
If this error affects a particular branch, and not the master, why should we revert it? Can't we fix this without reverting the code? Also, if you are testing, can't you use "stop"  in the debugger?

vncoelho:
> I didn't see if the database was being corrupted, but force closing neo can definitely cause an issue.

Now the forcing closing is happening on LINUX, @lock9. That is the point.

This error affects the master, @lock9.
The correct thing to do is to Revert and design the correct solution.

lock9:
Is this affecting the master version? Isn't this happening only in the branch you mentioned?
I tested it using multiple Linux distributions, they all close in seconds. I forced exceptions in code, and I got the same behavior. Owen also tested it and didn't find the problem either.
Are you using which versions of the code? And what OS?

vncoelho:
No, you can test in the master as well.
You need to try to initialize a private network with Consensus and you might see it.

Running on `microsoft/dotnet:2.2-sdk-bionic` docker container.

lock9:
@vncoelho can you write the steps to reproduce it? How are you running your private network? 1 node? Can you write the steps to configure that too? I'm testing using Testnet

lock9:
I think we might have a  problem with the actor system, so I'm afraid that reverting it would only hide a problem. Please send me the instructions to setup the same environment so I can debug it for you.

lock9:
@vncoelho do you have the branch with the current state/configurations? I think that this might be an option

PR #479, opened 2019-10-23
Make testnet setting as default on neo-cli(3.X)

superboyiii:
Since mainnet doesn't exist on NEO3 now, it's better to make testnet setting as default.


erikzhang:
I don't think testnet should be default in any case.

superboyiii:
> I don't think testnet should be default in any case.

We could turn back to mainnet whenever we have it. Since there's no mainnet now, this setting is no use but make new developers confused a lot.

PR #482, opened 2019-11-17
Update protocol.json to use testnet as default

lock9:
I think it is reasonable to make testnet configuration the default for now. There is no reason to keep Neo 2 main net configuration as default because it is not compatible.

### DIFF
--- a/ci/run-tests-in-docker.sh
+++ b/ci/run-tests-in-docker.sh
@@ -18,7 +18,7 @@ sleep 3
 # Test a RPX smart contract query
 JSONRPC_RES=$( curl --silent \
   --request POST \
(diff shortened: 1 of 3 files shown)


erikzhang:
I don't think it should use TESTNET configurations by default.

lock9:
Hi Erik, I understand and kind of agree, the issue that I see is that we are using the protocol for Neo 2, does that makes more sense than using testnet? Testnet is at least compatible. So, between neo 2 protocol and testnet, I believe testnet is more reasonable.

lock9:
Can we use a single node consensus instead? I think that using a Neo 2 protocol file makes less sense than using testnet or even a single node private blockchain.

PR #484, opened 2019-11-21
Memorystore test

lock9:
Hi @erikzhang. I used this code to test https://github.com/neo-project/neo/pull/1249. Maybe some update is wrong, but if that is the case, we need to be more clear which class or methods we should use.
This is just a draft for you to check.

### DIFF
--- a/neo-cli.sln
+++ b/neo-cli.sln
@@ -1,9 +1,11 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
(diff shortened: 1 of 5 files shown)


erikzhang:
I can't check it...

lock9:
@erikzhang  why is that? In any case, you should be able to run your version of neo-cli with that PR.
What can I do to make this testable for you? I only did 2 or 3 changes in mainservice (other changes are just my current branch state).

erikzhang:
Yes. It works fine for me after installing the `SystemLog` plugin.

image

PR #487, opened 2019-11-26
Remove Travis and use Github Actions

erikzhang:


erikzhang:
Ready to merge.

.github/workflows/rpc-tests.sh
```diff
@@ -1,16 +1,9 @@
 #!/bin/bash
-#
-# This script is run inside the Docker container and tests neo-cli
-#
-set -e
-
-cd /opt/neo-cli
 
-# Run tests with expect
-expect /opt/ci/test-neo-cli.expect
```
shargon: this line was removed?

erikzhang: It is moved to here:

https://github.com/neo-project/neo-cli/blob/1dc5c6071eafcfb585efbed45d78cdeadf6ac430/.github/workflows/dotnetcore.yml#L27-L28

PR #489, opened 2019-11-28
Remove address version config

belane:
Change NEO 3 version address to 53, this will produce that neo address starts with N.

Needs https://github.com/neo-project/neo/pull/1299
Closes https://github.com/neo-project/neo/issues/993 https://github.com/neo-project/neo/issues/1017


erikzhang:
We can just remove the field, and it will be set to the default value from Neo.dll.

shargon:
If mainnet and testnet have the same value, i am agree with @erikzhang

PR #491, opened 2019-12-01
Allow users to select storage engine

erikzhang:


shargon:
LevelDB will be distributed by default as a plugin in releases?

erikzhang:
All the storage plugins will be included.

vncoelho:
@shargon, perhaps we distribute all as it has been done and keep leveldb as default for now.

shargon:
Require https://github.com/neo-project/neo/pull/1312

PR #521, opened 2019-12-19
Showing version of neo and neo-vm on startup

melanke:
screenshot

@igormcoelho opened this issue https://github.com/neo-project/neo/issues/668 about printing neo versions on the log and @vncoelho suggested to print the git commit hash aswell.

I made a NerdBank.GitVersioning setup on the project to use a `version.json` file to define the version code (as suggested by @devhawk) and the git commit hash and save on AssemblyInformation.

This setup was made on neo and neo-vm aswell, if this PR is approved I can send PR for them so all versions will be printed with the commit hash.

I don't know if this solution satisfy all the needs presented by Igor, let me know if there is anything else I could do.

### DIFF
--- a/neo-cli/Services/ConsoleServiceBase.cs
+++ b/neo-cli/Services/ConsoleServiceBase.cs
@@ -352,7 +352,7 @@ namespace Neo.Services
 
             Console.ForegroundColor = ConsoleColor.DarkGreen;
 
-            var cliV = Assembly.GetEntryAssembly().GetVersion();
+            var cliV = Assembly.GetAssembly(typeof(Program)).GetVersion();
             var neoV = Assembly.GetAssembly(typeof(NeoSystem)).GetVersion();
             var vmV = Assembly.GetAssembly(typeof(ExecutionEngine)).GetVersion();
             Console.WriteLine($"{ServiceName} v{cliV}  -  NEO v{neoV}  -  NEO-VM v{vmV}");
--- a/neo-cli/Services/ConsoleServiceBase.cs
+++ b/neo-cli/Services/ConsoleServiceBase.cs
@@ -1,3 +1,4 @@
+using Neo.VM;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -350,7 +351,11 @@ namespace Neo.Services
                 catch { }
 
             Console.ForegroundColor = ConsoleColor.DarkGreen;
-            Console.WriteLine($"{ServiceName} Version: {Assembly.GetEntryAssembly().GetVersion()}");
+
+            var cliV = Assembly.GetEntryAssembly().GetVersion();
+            var neoV = Assembly.GetAssembly(typeof(NeoSystem)).GetVersion();
+            var vmV = Assembly.GetAssembly(typeof(ExecutionEngine)).GetVersion();
+            Console.WriteLine($"{ServiceName} v{cliV}  -  NEO v{neoV}  -  NEO-VM v{vmV}");
             Console.WriteLine();
 
             while (_running)


shargon:
What about put all together in the same line (3x-3x-3x)?

melanke:
> What about put all together in the same line (3x-3x-3x)?

sure, it's better, I will do this

neo-gui/neo-gui.csproj
```diff
@@ -1,9 +1,8 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
 
   <PropertyGroup>
     <Copyright>2016-2019 The Neo Project</Copyright>
     <AssemblyTitle>Neo.GUI</AssemblyTitle>
-    <Version>3.0.0-preview1</Version>
```
erikzhang: Why remove this line?

version.json
```diff
@@ -0,0 +1,12 @@
...
+  "publicReleaseRefSpec": [
+    "^refs/heads/master$",
+    "^refs/heads/v\\d+(?:\\.\\d+)?$"
+  ],
+  "cloudBuild": {
+    "buildNumber": {
+      "enabled": true
+    }
+  }
+}
```
erikzhang: What is this?

neo-gui/neo-gui.csproj
melanke: The version is comfigured on the version.json file now. So this line is not used anymore. This is the behaviour provided by NerdBank.GitVersioning, I think it is not possible to keep it on csproj and drop version.json

version.json
melanke: This file is used by NerdBank.GitVersioning. All this boilerplate was generated by its cli, but maybe we can erase everything and keep only the "version", but I am not sure about the consequences. I will try it.

erikzhang: What is the benefit of using `NerdBank.GitVersioning`?

superboyiii:
@melanke Good idea. It's good for locating issues when making intergration test.  But I think these changes for `ConsoleServiceBase.cs` are enough, we don't need to add addtional json files which makes it with trifles. And I also have another idea that maybe we could make it into `help` menu or create a new commend like `show dependency` which will not make users confused with what they're running when startup. I think only neo-cli-[version] should be showed at beginning since it's neo-cli.

superboyiii:
Oh, sorry for my misunderstanding, you'd like to get git commit specific version as well, right? Good idea.

version.json
superboyiii: > What is the benefit of using `NerdBank.GitVersioning`?

It could get git commit specific version as well.

erikzhang: I think we don't need it.

superboyiii: > I think we don't need it.

Maybe we can implement it to be a commend since it's useful in test purpose.

melanke: @erikzhang NerdBank.GitVersioning is used to put the git commit hash on the solution version. I believe it is good to be sure which specific version we are using, so we don't need to have a workflow to change the git tag for every push to master.

melanke:
@superboyiii I don't mind to change for showing the version only when running a specific command, but one problem Igor mentioned on the issue is that when you have a node with many log files is useful to know which version the problem is occuring and we can prevent this problem printing the version automatically on startup.

version.json
erikzhang: Hi, @melanke. We already have another mechanism for automatic version numbering:

https://github.com/neo-project/neo/blob/37136d6a1af97a6e679be9bee317f78a8f6f5f4f/.github/workflows/main.yml#L50

And I think we should not need to use `NerdBank.GitVersioning`. Otherwise conflicts may occur. But I think the modification of the display version is good. Could you please modify only the part that displays the version number?

melanke: Oh I didn't know about that. I removed the NerdBank.GitVersioning

neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -350,7 +351,11 @@ public void RunConsole()
                 catch { }
 
             Console.ForegroundColor = ConsoleColor.DarkGreen;
-            Console.WriteLine($"{ServiceName} Version: {Assembly.GetEntryAssembly().GetVersion()}");
+
+            var cliV = Assembly.GetEntryAssembly().GetVersion();
```
shargon: Better with typeof, neo-gui use this assembly as a reference

melanke: Done

vncoelho:
@melanke, in the current version can we see the commit version?

Where can we identify the commit hash?

melanke:
@vncoelho I removed NerdBank.GitVersioning because @erikzhang told me there is already a machanism for this purpose: https://github.com/neo-project/neo/blob/37136d6a1af97a6e679be9bee317f78a8f6f5f4f/.github/workflows/main.yml#L50

vncoelho:
But are you using this mechanism for showing the version on startup? I think it is useful.

Many people want to know, precisely, the commits of the client they are using. @erikzhang @shargon 

I will merge this PR for moving forward. However, I believe this information with `NerkBank` looks to be useful and of great importance for the image of those who use neo-cli.
In this sense, @melanke, perhaps you could open an issue or directly another PR.

superboyiii:
Tested. It works correctly. 
Dependency from nuget.
image
Dependency from internal existed project.
image

melanke:
Thank you @superboyiii

vncoelho:
@melanke, perhaps we you can open another PR with the Nerd bank and commit version and let it open for us to think and discuss.

I believe it is important to show commits as well.

melanke:
Ok, I will do this

PR #526, opened 2019-12-24
fix dead link

### LINKED ISSUE #522: Broken links in README
Here are several links from the main page which don't work:

> Bootstrapping the network.
> In order to synchronize the network faster, please check here.

here -- doesn't work

> Usage
> See documentation. E.g. try show state or create wallet wallet.json.

documentation -- doesn't work

> Download pre-compiled binaries
> See also official docs. Download and unzip latest release.

official docs -- doesn't work and latest release as well

Tommo-L:
close #522

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,21 @@
 <p align="center">
-<img
-    src="https://neo-cdn.azureedge.net/images/neo_logo.svg"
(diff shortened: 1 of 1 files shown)


shargon:
Github actions are not working?

PR #529, opened 2020-01-02
Add dependencies on fedora

corollari:
Trivial change, adding these dependencies so others don't have to dig through the package repositories to find the equivalent packages.


superboyiii:
@vncoelho Have you tested it on Fedora? We don't have this OS.

vncoelho:
I created a docker, @superboyiii.
```bash
sudo rpm --import https://packages.microsoft.com/keys/microsoft.asc
sudo wget -q -O /etc/yum.repos.d/microsoft-prod.repo https://packages.microsoft.com/config/fedora/30/prod.repo
sudo dnf install dotnet-sdk-3.1
```

Issue #531, opened 2020-01-08
Neo3: New address is added to address.txt twice when creating address.


nicolegys:
When I `create address`, the new address will be added to address.txt twice.

<img width="491" alt="Capture" src="https://user-images.githubusercontent.com/43694095/71968226-70803400-323f-11ea-9520-4bb2be2cee1c.PNG">

address.txt:
<img width="249" alt="Capture2" src="https://user-images.githubusercontent.com/43694095/71968242-783fd880-323f-11ea-984d-5c2c5c7634da.PNG">

OnCreateAddressCommand function:
![image](https://user-images.githubusercontent.com/43694095/71968467-f3a18a00-323f-11ea-9eda-dde241b845fd.png)

PR #532, opened 2020-01-08
Fix add address twice

shargon:
Close https://github.com/neo-project/neo-node/issues/531


nicolegys:
Test pass:

image

image

PR #455, opened 2019-08-30
Utilitary methods for Neo-cli

lock9:
Hi,
I've been testing neo3 and I added several methods that are useful for our users.
Note: this is _draft_ because I can't actually compile with the latest nuget package.

I hope that we decide to move to a single repository in the future and this problem is solved forever :pray:

**Changes**
show block:
showblock

show transaction:
showtransaction

show last-transactions:
showlast

show contract:
showcontract

tool hexToString:
tool stringToHex:
hextostring

tool numberToHex:
tool hexToNumber:
hextonumber

tool scriptToAddr:
tool addressToScript:
addresstoscript

Other improvements:
showing notifications when invoking an script:
image

script "decoding"

PR #535, opened 2020-01-16
Conversion utility methods for Neo-cli

meevee98:
It's an update from #455.

This pull request has only the conversion utility methods because I'm still unable to connect to Neo 3 testnet to test some of ``show`` methods.

`parse <value>`
image


shargon:
What do you think about `tool parse`, and accept all: address, pubkey or scripthash, and show that it can resolve at same time?

melanke:
@shargon if you put a string hex you will need to specify if you want to transform it to string or to number. And it can get a little to verbose:
```
tool parse hex number 0x2a
```

erikzhang:
I prefer to add these functions to GUI.

superboyiii:
@meevee98 @shargon @erikzhang I've already made a simple tool to support most of conversion method. https://github.com/neo-ngd/Neo3-Tool. If you'd like, could adjust and merge it into your code.

meevee98:
> What do you think about `tool parse`, and accept all: address, pubkey or scripthash, and show all that it can resolve at same time?

I agree that it's better than having a different command for each conversion. I will modify the code in this way.

> @meevee98 @shargon @erikzhang I've already made a simple tool to support most of conversion method. https://github.com/neo-ngd/Neo3-Tool. If you'd like, could adjust and merge it into your code.

Nice. If I may, I will include the other conversions to this implementation.

> I prefer to add these functions to GUI.

I think that these commands are also useful in neo-cli.
When we deploy a invoke a smart contract using the neo-cli, the return result is showed encoded in Base64, so a converting tool is needed anyway. If it's in the same application, the work gets faster, doesn't it?

melanke:
I would love to have this commands on Neo-Cli, thank you Meevee

meevee98:
I've updated this pull request with @shargon's [suggestion](https://github.com/neo-project/neo-node/pull/535#issuecomment-575260355) of `tool parse` command.
I've also included the other conversion methods from @superboyiii's [tool](https://github.com/neo-ngd/Neo3-Tool).

It's a unique command now that accepts any value and returns the possible conversions

neo-cli/CLI/MainService.cs
```diff
@@ -741,9 +1286,12 @@ private bool OnHelpCommand(string[] args)
             Console.WriteLine("\tplugins");
             Console.WriteLine("\tinstall <pluginName>");
             Console.WriteLine("\tuninstall <pluginName>");
+            Console.WriteLine("Tool Commands:");
+            Console.WriteLine("\ttool parse <value>");
```
erikzhang: Remove `tool` and keep only `parse`?

shargon: Only `parse` or `convert` it's good to me

neo-cli/CLI/MainService.cs
```diff
@@ -211,11 +211,545 @@ protected override bool OnCommand(string[] args)
...
+            {
+                Console.WriteLine("Invalid Parameters");
+            }
+            else
+            {
+                var input = args[1];
+                var parseFunctions = new Dictionary<string, Func<string, string>>()
+                {
+                    { "Address to BigEnd ScriptHash", AddressToBigEndScripthash },
+                    { "Address to LittleEnd ScriptHash", AddressToLittleEndScripthash },
```
shargon: We should have only one, the official one.

erikzhang: Agree.

meevee98: I've changed to convert only to big end script hash because that's the script hash that is shown with transactions and blocks.
What do you think? Is it okay or should I change it to little end?

neo-cli/CLI/MainService.cs
```diff
@@ -211,11 +211,448 @@ protected override bool OnCommand(string[] args)
...
+            {
+                throw new ArgumentException();
+            }
+            catch (ArgumentNullException)
+            {
+                throw new ArgumentException();
+            }
+        }
+
+
```
shargon: Remove this blank like please (I can't edit the file)

neo-cli/CLI/MainService.cs
```diff
@@ -211,11 +211,447 @@ protected override bool OnCommand(string[] args)
...
+                {
+                    try
+                    {
+                        var parseMethod = pair.Value;
+                        var result = parseMethod(input);
+
+                        Console.WriteLine($"{pair.Key,-30}\t{result}");
+                    }
+                    catch (ArgumentException)
+                    {
```
shargon: We should show the exception?

meevee98: That exception is thrown when the value cannot be parsed.

To convert the value, it tries to parse it in each conversion method implemented.
For example, if we enter ``parse 100000``, it will try to parse the 100000 to a number, to a string, to a hex string, to an address, and to a script hash and it will fail with the latter two. When it fails is when the exception is thrown.

I don't think we should show the exception unless no conversion is successful. If we show it, with the last example it would show six successful conversions and six parse error messages and I don't think that makes any sense.

melanke: Maybe we could store each parsing method that threw an exception in a array and show a message saying it wasn't possible to parse on the following methods: X, Y, Z, in a single line.

shargon: Something generic for all of then it's good, can't find a valid parse mechanism, or something like that.

shargon: Could you allow me to edit the file?

nicolegys:
Tested.

Do we still need to support converting a little end script hash to address?
There is only the big end script hash in your comment, but, in your code, the little one is also included.
If we support both of them, users must pay close attention to the format. `0x61ce772f7a725f93d8311dba9994dd8c50e3e64b` and `61ce772f7a725f93d8311dba9994dd8c50e3e64b` are quite different.

image

neo-cli/CLI/MainService.cs
melanke: Hi @shargon, I gave you write access to simplitech/neo-cli. Is that enought?

shargon:
From my side I think that it's ready to merge but I prefer to merge it after https://github.com/neo-project/neo-node/pull/536 It's good for you @meevee98 ?

meevee98:
It's no problem for me @shargon

shargon:
image

Maybe we should filter in `base64ToString` to only ASCII chars. And HexString to String to "non-empty"

shargon:
@meevee98 I am not sure if it's my computer, but it spend a lot of time in something trivial, could you check it?

image

meevee98:
> Maybe we should filter in `base64ToString` to only ASCII chars. And HexString to String to "non-empty"
Also it's not possible to be Hex if there odd letters

@shargon I agree with the empty strings, but I think that the conversions to a string with non-ascii characters should be kept.
I've seen in many smart contracts the concatenation between a string prefix and a bytearray been used and the result usually can't be converted to a string with only ascii characters, so it is possible to be a valid hex string even if there are some odd letters. 

For example, this hex string is converted to a string with many non-ascii characters
image
But that is because part of the hex string is the scripthash of an address
image


> @meevee98 I am not sure if it's my computer, but it spend a lot of time in something trivial, could you check it?
image

I'm not having this delay 🤔
image

shargon:
>  but I think that the conversions to a string with non-ascii characters should be kept.

Ok :)

shargon:
I think that 0.2 seconds in your computer it's a lot of time.... we need to find the issue

meevee98:
@shargon I was able to reduce the time to 0.1 seconds by removing the exception throwing in the conversion methods and replacing them with return null.
image
Could you check if the time is reduced on your computer as well?

superboyiii:
@meevee98 It's out of date, could you update it to the latest?

PR #536, opened 2020-01-23
Refactor node commands

shargon:
This pull request aims to improve the next things:

- Reorder the node commands.
- Ensure that all commands will have help for users.
- Unify the validation entries around all commands. 
- Provide auto-help.
- Make easier the creation of new commands for nodes and plugins.
- **Make easier the implementation of autocomplete feature (it will come after merge this).**

image


neo-cli/CLI/CommandParser/CommandStringToken.cs
```diff
@@ -0,0 +1,88 @@
...
+            RequireQuotes = requireQuotes || value.Contains("\"");
+        }
+
+        /// <summary>
+        /// Parse command line spaces
+        /// </summary>
+        /// <param name="commandLine">Command line</param>
+        /// <param name="index">Index</param>
+        /// <returns>CommandSpaceToken</returns>
+        internal static CommandStringToken Parse(string commandLine, ref int index)
```
erikzhang: Can we use a third-party parser?

shargon: Of course, do you have any recommendation?

shargon: The problem that I find is that for this state (without autocomplete) it's easy to find a method for parse the command line arguments  to `string[]` but I want to implement an autocomplete feature later, and I will need to know the offset of this argument, because I will need to know in wich argument is the cursor located.

shargon: I tried with https://github.com/dotnet/command-line-api but it doesn't preserve the space, also the escape characters, so for autocomplete could be worst with this lib

image

neo-cli/CLI/MainService.Blockchain.cs
```diff
@@ -0,0 +1,42 @@
+using Neo.CLI.CommandParser;
+using Neo.Ledger;
+using Neo.Services;
+using System;
+
+namespace Neo.CLI
+{
+    public partial class MainService : ConsoleServiceBase
```
erikzhang:
```suggestion
    partial class MainService
```

erikzhang: And other places.

neo-cli/CLI/MainService.Consensus.cs
```diff
@@ -0,0 +1,30 @@
+using Akka.Actor;
+using Neo.CLI.CommandParser;
+using Neo.Consensus;
+using Neo.Services;
+
+namespace Neo.CLI
+{
+    partial class MainService : ConsoleServiceBase
```
erikzhang:
```suggestion
    partial class MainService
```

shargon:
@superboyiii could your team test that all the commands still working as before?

nicolegys:
@shargon I'll give you the test results later.

nicolegys:
Here are my test results. Kindly let me know if I get something wrong.

### Blockchain Commands:
**`export block[s] <start> [count=4294967295]`
`export block[s] <path> <start>`**
image
image
1. not support export block[s] 
image
In this pr, `<start>` is a required parameter, but I think we need to support `export block[s]`.
2. export block[s] start count
image
I try to debug and find that `availableCommands` has 2 counts. 
image
Because export block[s] command has two entries, and when trying the second, the program recognizes `100` as a string.
3.export blocks (specify path chain.n.acc, and set start = n), then restart neo-cli, will fail
image
It seems that value of `writeStart` should be true? The start is not 0.
https://github.com/shargon/neo-cli/blob/unify-commands/neo-cli/CLI/MainService.Blockchain.cs#L34

### Node Commands:
**`show pool [verbose=False]`**
Enter `show pool true` `show pool 1`, neo-cli will crash.
image
image

### Wallet Commands:
**`create address [count=1]`**
image
This error will disappear after the first issue of `export key` being fixed.

**`export key <path> [scriptHash=null]`**
1. A small mistake in its ConsoleCommand.
image
2.`help export key` throw exception.
image
because the DefaultValue of second param is null, change https://github.com/shargon/neo-cli/blob/unify-commands/neo-cli/Services/ConsoleServiceBase.cs#L216 to the following, will pass.
```
.Select(u => u.HasDefaultValue ? $"[{u.Name}={(u.DefaultValue == null ? "null" : u.DefaultValue.ToString())}]" : $"<{u.Name}>"))
```
3.Here path is a required parameter, but I think we need to support export to the console.
4.How about changing the second param to address, to make it more user friendly?

**`import key <wif> <file>`**
I think `wif` and `file` should be two options, but now I have to enter both of them. 
image

**`import multisigaddress <m> <publicKeys>`**
One pubkey, two pubkeys, three pubkeys, all get the same multisigaddress. And It will return error when m > 1.
image
I find that, the function allows us to enter several pubkeys,
image
but here, it will only read the first 2 params.
image

**`send <assetName> <to> <amount>
send <assetId> <to> <amount>`**
1.Asking users to enter a scriptHash is not user friendly. How about changing the second param to address, or parsing the address to scriptHash?
image
image
2. When I choose the second send function, it returns error.
image
I try to debug and find that `availableCommands` has 2 counts. Because send command has two entries, and when trying the first, the program recognizes `0x8c23f196d8a1bfd103a9dcb1f9ccf0c611377d3b` as a string.
image

### Contract Commands:
**`deploy <filePath> <manifestPath>`**
We used to support not entering `manifestPath`.
**`invoke <scriptHash> <operation> <contractParameters>`**
Return fault.
image
Here should be `i = 0`
image

### Network Commands:
**`broadcast <command> <payload>`
`broadcast <command> <payload> <hashes>`
`broadcast <command> <payload> <port>`**
Also exists the same problem as other commands.
image

neo-cli/CLI/CommandParser/CommandTokenType.cs
```diff
@@ -0,0 +1,8 @@
+namespace Neo.CLI.CommandParser
+{
+    public enum CommandTokenType : byte
+    {
+        String = 0x00,
+        Space = 0x01
+    }
```
erikzhang: `Quote`?

shargon:
@nicolegys Thanks for your amazing review, please take a look with the last changes :)

neo-cli/CLI/CommandParser/CommandTokenType.cs
shargon: Done

neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -27,134 +31,251 @@ public abstract class ConsoleServiceBase
...
-                default:
-                    Console.WriteLine("error: command not found " + args[0]);
-                    return true;
-            }
+            public ConsoleCommandAttribute Command { get; set; }
+            public object[] Arguments { get; set; }
         }
 
-        protected internal virtual void OnStart(string[] args)
+        protected virtual bool OnCommand(string commandLine)
```
erikzhang: This can be `private` now, right?

shargon: Yes, it could be

erikzhang: neo-gui doesn't call it. I can't find where it calls it.

shargon: You are right, I thought that we use this method...

shargon:
Thanks @nicolegys ! Check it again :)

nicolegys:
**`export blocks <start> <count> [path=null]`**
ambiguous calls
image

shargon:
@nicolegys I think that we can have only one method for export, it's good for you?

nicolegys:
@shargon I agree with you, it is good for me.

neo-cli/CommandParser/CaptureWholeArgumentAttribute.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+
+namespace Neo.CommandParser
+{
+    /// <summary>
+    /// It's required if we don't require to use quotes for capture the argument.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
+    public class CaptureWholeArgumentAttribute : Attribute { }
```
erikzhang: Is it used?

shargon: It's used when you want to receive all the line without the command and without quotes.

For example

`help broadcast addr` otherwise should be `help "broadcast addr"`
`invoke hash method ["123",123]` otherwise should be `invoke hash method "[\"123\",123]"`

shargon: If we want to follow the "command standard" we can remove it, but all arguments must be quoted

erikzhang: I mean it is never used in the code.

shargon: Here you have an example
https://github.com/neo-project/neo-node/pull/536/files#diff-43d574f37fbdff4dd81543ee087c6172R174

nicolegys:
@shargon I found that `help + plugin name` disappered, like this:
image
could you fix it?

neo-cli/CommandParser/CommandTokenType.cs
```diff
@@ -0,0 +1,9 @@
+namespace Neo.CommandParser
+{
+    public enum CommandTokenType : byte
+    {
+        String = 0x01,
+        Space = 0x02,
+        Quote = 0x03,
```
erikzhang: It is not necessary to set the values.

shargon:
@nicolegys 

Plugin must change the commands, but it should show now the help of all available commands in this plugin

nicolegys:
> @nicolegys
> 
> Plugin must change the commands, but it should show now the help of all available commands in this plugin

@shargon Did you mean you will change the plugin commands in another PR in the future?

shargon:
Yes, the modules will need to upgrade to this mechanism, i will change all of them asap.

nicolegys:
Thank you @shargon , now this PR is ok for me.

neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -165,11 +165,31 @@ protected void OnHelpCommand([CaptureWholeArgument] string key = "")
...
+            var plugin = Plugins.Plugin.Plugins
+                            .Where(u => u.Name.Equals(key.Trim(), StringComparison.InvariantCultureIgnoreCase))
+                            .FirstOrDefault();
+
+            // Fetch commands
+
             foreach (var commands in _verbs.Values.Select(u => u))
             {
                 withHelp.AddRange(commands.Where(u => !string.IsNullOrEmpty(u.HelpCategory)));
             }
```
erikzhang: These line should be executed only when `plugin` is `null`, right?

shargon: I will optimize it

neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -27,134 +32,275 @@ public abstract class ConsoleServiceBase
...
+            // Fetch commands
+
+            if (plugin != null)
             {
-                int c = reader.Peek();
-                switch (c)
+                // Filter only the help of this plugin
+
+                key = "";
+                foreach (var commands in _verbs.Values.Select(u => u))
```
erikzhang: Why traverse `_verbs` here?

shargon: Because this collection already have loaded all the information of all plugins

neo-cli/CLI/MainService.Blockchain.cs
```diff
@@ -0,0 +1,32 @@
...
+    partial class MainService
+    {
+        /// <summary>
+        /// Process "export blocks" command
+        /// </summary>
+        /// <param name="start">Start</param>
+        /// <param name="count">Number of blocks</param>
+        /// <param name="path">Path</param>
+        [ConsoleCommand("export", "blocks", HelpCategory = "Blockchain Commands")]
+        private void OnExportBlocksStartCountCommand(uint start, uint count, string path = null)
```
bettybao1209: Previously we are allowed to export all blocks without speficying `start` and `count`. I think maybe it should not be disabled.

neo-cli/CLI/MainService.Blockchain.cs
```diff
@@ -0,0 +1,32 @@
...
+            count = Math.Min(count, Blockchain.Singleton.Height - start + 1);
+
+            if (string.IsNullOrEmpty(path))
+            {
+                path = $"chain.{start}.acc";
+            }
+
+            var writeStart = true;
+
+            WriteBlocks(start, count, path, writeStart);
```
bettybao1209: Here the `writeStart` is always `true`, it seems https://github.com/neo-project/neo-node/blob/b1ba72a9b022490d24ca8c52c696a226b2b88f2c/neo-cli/CLI/MainService.cs#L372 is redundant.

neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -0,0 +1,506 @@
...
+            File.WriteAllLines(path, addresses);
+        }
+
+        /// <summary>
+        /// Process "export key" command
+        /// </summary>
+        /// <param name="path">Path</param>
+        /// <param name="scriptHash">ScriptHash</param>
+        [ConsoleCommand("export", "key", HelpCategory = "Wallet Commands")]
+        private void OnExportKeyCommand(string path, UInt160 scriptHash = null)
```
bettybao1209: Here `path` should also be an optional parameter (allowed to be null ) to allow display the keys in console.

neo-cli/Services/ConsoleServiceBase.cs
erikzhang: `ConsoleCommandAttribute` is defined in neo-cli. How can a plugin use this attribute to register the commands?

neo-cli/CLI/MainService.Network.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            OnBroadcastCommand(MessageCommand.Block, Blockchain.Singleton.GetBlock(hash));
+        }
+
+        /// <summary>
+        /// Process "broadcast block" command
+        /// </summary>
+        /// <param name="block">Block</param>
+        [ConsoleCommand("broadcast", "block", HelpCategory = "Network Commands")]
+        private void OnBroadcastGetBlocksByHeightCommand(uint block)
```
bettybao1209: Here maybe it's better to use `index` or 'height' instead of block to provide more clear info.

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        /// <summary>
+        /// Process "invoke" command
+        /// </summary>
+        /// <param name="scriptHash">Script hash</param>
+        /// <param name="operation">Operation</param>
+        /// <param name="contractParameters">Contract parameters</param>
+        [ConsoleCommand("invoke", HelpCategory = "Contract Commands")]
+        private void OnInvokeCommand(UInt160 scriptHash, string operation, [CaptureWholeArgument] JArray contractParameters)
```
bettybao1209: `contractParameters` should have default value `null` to allow invoking functions without parameters, like name, decimals, etc.

neo-cli/CLI/MainService.Blockchain.cs
shargon: We decided to unify in one command because otherwise you can't distinguish between a valid path "0" and integer, it produce ambiguous calls

neo-cli/CLI/MainService.Blockchain.cs
shargon: When you don't want to `writeStart`?

neo-cli/Services/ConsoleServiceBase.cs
shargon: Good point, we can move this attribute to neo?

neo-cli/CLI/MainService.Blockchain.cs
bettybao1209: Now WriteBlocks is only called here, but it's ok for me.

neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -0,0 +1,506 @@
...
+        {
+            if (NoWallet()) return;
+
+            using (var snapshot = Blockchain.Singleton.GetSnapshot())
+            {
+                foreach (Contract contract in CurrentWallet.GetAccounts().Where(p => !p.WatchOnly).Select(p => p.Contract))
+                {
+                    var type = "Nonstandard";
+
+                    if (contract.Script.IsMultiSigContract(out _, out _))
```
nicolegys: `IsMultiSigContract()` was updated in neo, we should make it compatible.
https://github.com/neo-project/neo/commit/f26ac1be3b7ab421e8c88d03a72a42deacae3a05

neo-cli/Services/ConsoleServiceBase.cs
```diff
@@ -27,134 +32,275 @@ public abstract class ConsoleServiceBase
...
+
+                                if (TryProcessValue(arg.ParameterType, args, arg.GetCustomAttribute<CaptureWholeArgumentAttribute>() != null, out var value))
+                                {
+                                    arguments.Add(value);
+                                }
+                                else
+                                {
+                                    if (arg.HasDefaultValue)
+                                    {
+                                        arguments.Add(arg.DefaultValue);
```
Tommo-L: If the current parameter hasn't been set, the rest of the parameters should use the default value or null. In this case, we can support overloading.
```c#
private void OnExportBlocksStartCountCommand(uint start = 0, uint count = 4294967295, string path = null)
```

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        /// <summary>
+        /// Process "invoke" command
+        /// </summary>
+        /// <param name="scriptHash">Script hash</param>
+        /// <param name="operation">Operation</param>
+        /// <param name="contractParameters">Contract parameters</param>
+        [ConsoleCommand("invoke", HelpCategory = "Contract Commands")]
+        private void OnInvokeCommand(UInt160 scriptHash, string operation, [CaptureWholeArgument] JArray contractParameters = null)
```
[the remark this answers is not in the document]
shargon: if they are between quotes, i think that there are no problem, but [CaptureWholeArgument] must be at the end

neo-cli/CLI/MainService.Wallet.cs
shargon: Done

shargon:
@Tommo-L 

>export blocks
export blocks start
export blocks start count 
export blocks start count path
I think the first is the most common use.

Maybe with this it's enough?

`private void OnExportBlocksStartCountCommand(uint start, uint count = uint.MaxValue, string path = null)`

neo-cli/Services/ConsoleServiceBase.cs
shargon: > we can support overloading.

we already did, but if this overloading is not possible to distinguish, we have an ambiguous call

Example:

```
Call(int start);
Call(string arg);
Call(int start, string arg=null);
```

shargon: Is the case of export blocks, "0" it's a valid path

Tommo-L: Do we support both `export blocks path` and `export blocks start` ?

shargon: How do you difference between "0" as string and 0 as number?

Tommo-L: As the first argument is `start`, why 0 is the value of  `path`?  If we support the following format. And we use the type of argument, convert 0 to uint.
```
export blocks
export blocks start
export blocks start count
export blocks start count path
```

neo-cli/CLI/MainService.Contracts.cs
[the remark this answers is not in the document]
Tommo-L: 😂  I'm worried that we won't be able to support `cosiginer` here.

neo-cli/CLI/MainService.Blockchain.cs
vncoelho: I agree, perhaps the export blocks should still be valid.

neo-cli/CLI/MainService.Consensus.cs
```diff
@@ -0,0 +1,29 @@
...
+using Neo.Consensus;
+
+namespace Neo.CLI
+{
+    partial class MainService
+    {
+        /// <summary>
+        /// Process "change view" command
+        /// </summary>
+        [ConsoleCommand("change", "view", HelpCategory = "Consensus Commands")]
```
vncoelho: In this command still needed? Has this been used after dBFT 2.0?

shargon: Maybe we can have it only under debug directive

neo-cli/Services/ConsoleServiceBase.cs
shargon: Take a look again please

neo-cli/CLI/MainService.Consensus.cs
erikzhang: I think it is not needed.

neo-cli/CLI/MainService.Contracts.cs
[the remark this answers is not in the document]
shargon: We will need to remove the CaptureWholeArgument form contractParameters, in order to add cosigners

neo-cli/Services/ConsoleServiceBase.cs
erikzhang: I think it shouldn't be moved to neo. We have to find another way to get the help message.

shargon: For help or category could be easier, but for the command key... i am not sure

https://github.com/neo-project/neo-node/pull/536/files#diff-c07317dd01f102f456eb88d332c567f1R15

erikzhang: Maybe we should turn `ConsoleServiceBase` and `CommandParser` into a separate library.

neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -373,38 +386,14 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
         }
 
-        /// <summary>
-        /// Process "send" command
-        /// </summary>
-        /// <param name="assetName">Asset name</param>
-        /// <param name="to">To</param>
-        /// <param name="amount">Amount</param>
-        [ConsoleCommand("send", HelpCategory = "Wallet Commands", ExcludeIfAmbiguous = true)]
-        private void OnSendCommand(string assetName, UInt160 to, string amount)
```
nicolegys: Why remove `assetName`?  I think it's inconvenient.

nicolegys:
I think we should let users know that they can also use `assetName`.
image

neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -0,0 +1,496 @@
...
+                Console.WriteLine("cancelled");
+                return;
+            }
+            string password2 = ReadUserInput("password", true);
+            if (password != password2)
+            {
+                Console.WriteLine("error");
+                return;
+            }
+            CreateWallet(path, password);
```
superboyiii: I think we should add a check here, like:
```
if (!File.Exists(path))
            {
                CreateWallet(path, password);
            } else
            {
                Console.WriteLine("This wallet already exists, please create another one.");
            }
```

neo-cli/CLI/MainService.Contracts.cs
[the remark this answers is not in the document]
superboyiii: > We will need to remove the CaptureWholeArgument form contractParameters, in order to add cosigners

Seems command check should also be adjusted. I tried to make invoke command like this, but can't pass the check. However if I exchange the position of `contractParameters` and `witnessAddress`, it will execute normally.
```
        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, string witnessAddress = null)
        {
            List<ContractParameter> parameters = new List<ContractParameter>();
            List<Cosigner> signCollection = new List<Cosigner>();

            if (!NoWallet() && witnessAddress != null)
            {
                try
                {
                    witnessAddress.ToScriptHash();
                }
                catch
                {
                    witnessAddress = null;
                }

                using (SnapshotView snapshot = Blockchain.Singleton.GetSnapshot())
                {
                    UInt160[] accounts = CurrentWallet.GetAccounts().Where(p => !p.Lock && !p.WatchOnly).Select(p => p.ScriptHash).Where(p => NativeContract.GAS.BalanceOf(snapshot, p).Sign > 0).ToArray();
                    foreach (var signAccount in accounts)
                    {
                        if (witnessAddress is null)
                        {
                            break;
                        }
                        if (witnessAddress.Equals(signAccount.ToAddress()))
                        {
                            signCollection.Add(new Cosigner() { Account = signAccount });
                        }
                    }
                }
            }
            
            if (contractParameters != null)
            {
                foreach (var contractParameter in contractParameters)
                {
                    parameters.Add(ContractParameter.FromJson(contractParameter));
                }
            }

            Transaction tx = new Transaction
            {
                Sender = UInt160.Zero,
                Attributes = Array.Empty<TransactionAttribute>(),
                Witnesses = Array.Empty<Witness>(),
                Cosigners = signCollection.ToArray()
            };

            using (ScriptBuilder scriptBuilder = new ScriptBuilder())
            {
                scriptBuilder.EmitAppCall(scriptHash, operation, parameters.ToArray());
                tx.Script = scriptBuilder.ToArray();
                Console.WriteLine($"Invoking script with: '{tx.Script.ToHexString()}'");
            }

            using (ApplicationEngine engine = ApplicationEngine.Run(tx.Script, tx, testMode: true))
            {
                Console.WriteLine($"VM State: {engine.State}");
                Console.WriteLine($"Gas Consumed: {engine.GasConsumed}");
                Console.WriteLine($"Evaluation Stack: {new JArray(engine.ResultStack.Select(p => p.ToParameter().ToJson()))}");
                Console.WriteLine();
                if (engine.State.HasFlag(VMState.FAULT))
                {
                    Console.WriteLine("Engine faulted.");
                    return;
                }
            }

            if (NoWallet()) return;
            try
            {
                tx = CurrentWallet.MakeTransaction(tx.Script);
            }
            catch (InvalidOperationException)
            {
                Console.WriteLine("Error: insufficient balance.");
                return;
            }
            if (!ReadUserInput("relay tx(no|yes)").IsYes())
            {
                return;
            }
            SignAndSendTx(tx);
        }
    }
}
```
image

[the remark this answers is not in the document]
superboyiii: Oops! It must be like this. It's normal now.
image
:laughing:

shargon:
> I think we should let users know that they can also use assetName.

Change it to `asset`?

nicolegys:
Now only the first pubkey can be transferred to `OnImportMultisigAddress` method.
image

shargon:
> Now only the first pubkey can be transferred to OnImportMultisigAddress method.

It should be separate with ','  or with spaces between quotes.

nicolegys:
```
import multisigaddress 1 034783f18f1b73929169168362f81d37b5e16d3f28d93a3392af87684bdf353bca 03a897f69fd5b77943c9f5c3be48d7035b077273387233ad51a2e5414e5e5165de
```
When `TryProcessValue`, the args[0] is space.
image
So, here `ret` string starts with a space.
image
So, here str[0] is "".
image
Then it will throw exception when trying to parse ""
image

The same cause for the failure of invoke command (with witness).

nicolegys:
Thank you@shargon  Now it works well.

shargon:
Any idea about how to use `ConsoleCommand` in modules?

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// Process "invoke" command
+        /// </summary>
+        /// <param name="scriptHash">Script hash</param>
+        /// <param name="operation">Operation</param>
+        /// <param name="contractParameters">Contract parameters</param>
+        /// <param name="witnessAddress">Witness address</param>
+        [Category("Contract Commands")]
+        [ConsoleCommand("invoke")]
+        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160[] witnessAddress = null)
+        {
```
superboyiii: Shall we change witnessAddress to string[]? I think it's more convinence to use standard address. ScriptHash of address need one more round for conversion, and this need utility method https://github.com/neo-project/neo-node/pull/535 to merge.

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -0,0 +1,137 @@
...
+                {
+                    Console.WriteLine("Engine faulted.");
+                    return;
+                }
+            }
+
+            if (NoWallet()) return;
+            try
+            {
+                tx = CurrentWallet.MakeTransaction(tx.Script);
```
Tommo-L: Lost parameters
```suggestion
                tx = CurrentWallet.MakeTransaction(tx.Script, null, tx.Attributes, tx.Cosigners);
```

neo-cli/CLI/MainService.Contracts.cs
shargon: A valid address will be converted to UInt160

superboyiii: I mean like this: `invoke 0x214d27c05adac40a56d0c5d7b5001d9de5c72ec2 deploy [] NPvKVTGZapmFWABLsyvfreuqn73jCjJtN1`
I saw you're using `UInt160.Parse(u.Trim())).ToArray()` when input witness is UInt160[]`, but only `TryParse` can converted address to UInt160.

shargon: Good point! done!

shargon:
What do you think if plugin require to call a method for register the command? @erikzhang @neo-project/ngd-shanghai

erikzhang:
I think it's okay. But how does a plugin call a method from neo-cli?

neo-cli/CLI/MainService.Contracts.cs
superboyiii: It works well now.

ProDog:
Have you considered adding migrate to the cli?etc.
image

shargon:
> I think it's okay. But how does a plugin call a method from neo-cli?

neo-cli can call his method with a method as argument

shargon:
@erikzhang please, take a look to https://github.com/neo-project/neo-node/pull/536/commits/6b8c68156e25987fcaf3bb061a4b710d3eba2a64 and https://github.com/neo-project/neo/pull/1470

erikzhang:
Then I would prefer to create a separate library that contains `ConsoleServiceBase` and `CommandParser`. Then neo-cli and plugins both reference it.

superboyiii:
@shargon  Since https://github.com/neo-project/neo-devpack-dotnet/pull/187 contains `Migrate` method, we're trying to invoke it during our test. However, we found it's very convenient to convert nefFile to byte[], so finally @ProDog has made a new command `migrate` into cli. But we think it's not very reasonable to put it at this position since `migrate` is one of invoke methods? So what's your opinion. Anyway, we think we should provide an interface for developer to input nef script byteArray like `deploy` method.

shargon:
@erikzhang Could you create this intermediary repository?
@superboyiii I agree, but I think that we should add commands after merge this PR

superboyiii:
> @erikzhang Could you create this intermediary repository?
> @superboyiii I agree, but I think that we should add commands after merge this PR

Agree.

ProDog:
> @shargon Since [neo-project/neo-devpack-dotnet#187](https://github.com/neo-project/neo-devpack-dotnet/pull/187) contains `Migrate` method, we're trying to invoke it during our test. However, we found it's very inconvenient to convert nefFile to byte[], so finally @ProDog has made a new command `migrate` into cli. But we think it's not very reasonable to put it at this position since `migrate` is one of invoke methods. So what's your opinion? Anyway, we think we should provide an interface for developer to input nef script byteArray like `deploy` method.

In addition,if don't have a special migrate way, using cli's invoke to migrate can easily exceed the Console App‘s maximum input limit.

Suggested add:
`Console.SetIn(new StreamReader(Console.OpenStandardInput(maxnum)));`

Neo.ConsoleService/ConsoleServiceBase.cs
```diff
@@ -0,0 +1,643 @@
...
+        /// </summary>
+        [Category("Base Commands")]
+        [ConsoleCommand("help")]
+        protected void OnHelpCommand(string key)
+        {
+            var withHelp = new List<ConsoleCommandMethod>();
+
+            // Try to find a plugin with this name
+
+            var plugin = Plugins.Plugin.Plugins
```
erikzhang: All NEO related stuff must be moved from this project.

shargon: Done

nicolegys:
Works well, shall we merge it, then go on with https://github.com/neo-project/neo-modules/pull/187?

shargon:
Is good for me, but first, do you prefer only one attribute, or multiple like now?

erikzhang:
I prefer to use a single attribute.

shargon:
Done

Neo.ConsoleService/CommandToken.cs
```diff
@@ -0,0 +1,219 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Neo.ConsoleService
+{
+    public abstract class CommandToken
```
erikzhang: Why does it have to be `public`?

shargon: Was used because the handlers receive the list
https://github.com/neo-project/neo-node/pull/536/files#diff-3c91c28f4dea60d157e514145e03f575R423

Please take a look, I made some improves in order to change it to internal

tests/neo-cli.Tests/neo-cli.Tests.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RootNamespace>neo_cli.Tests</RootNamespace>
```
erikzhang: Maybe it should test the project "Neo.ConsoleService" directly.

erikzhang:
Don't forget to modify the workflow so that it can automatically publish `Neo.ConsoleService` to MyGet.

tests/Neo.ConsoleService.Tests/Neo.ConsoleService.Tests.csproj
```diff
@@ -0,0 +1,18 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+    <PackageReference Include="MSTest.TestAdapter" Version="2.1.0" />
+    <PackageReference Include="MSTest.TestFramework" Version="2.1.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\neo-cli\neo-cli.csproj" />
```
erikzhang: Need to update the reference.

shargon:
I think that `System.ServiceProcess.ServiceController` has problems in linux :S

> Don't forget to modify the workflow so that it can automatically publish Neo.ConsoleService to MyGet.

@erikzhang could you help me with this, i never did it in the past and I don't want to do a mistake with the nuget.

.github/workflows/dotnetcore.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     - name: Run Neo.ConsoleService Tests
       run: |
         find tests -name *.csproj | xargs -I % dotnet add % package coverlet.msbuild
-        dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=lcov /p:CoverletOutput=${GITHUB_WORKSPACE}/coverage/lcov
+        dotnet test tests/Neo.ConsoleService.Tests/Neo.ConsoleService.Tests.csproj /p:CollectCoverage=true /p:CoverletOutputFormat=lcov /p:CoverletOutput=${GITHUB_WORKSPACE}/coverage/lcov
```
erikzhang: Why add the project name?

erikzhang:
> I think that System.ServiceProcess.ServiceController has problems in linux :S

Yes. It is for Windows only.

> could you help me with this, i never did it in the past and I don't want to do a mistake with the nuget.

I will do it after the tests pass.

.github/workflows/dotnetcore.yml
shargon: It doesn't work without the name
https://github.com/neo-project/neo-node/runs/513334056

shargon: I think that otherwise it use neo-gui by default

erikzhang:
> Don't forget to modify the workflow so that it can automatically publish Neo.ConsoleService to MyGet.

I think that `Neo.ConsoleService` may not be modified often, so I plan to publish it directly to NuGet after the PR merges.

neo-cli/CLI/MainService.Blockchain.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <summary>
+        /// Process "export blocks" command
+        /// </summary>
+        /// <param name="start">Start</param>
+        /// <param name="count">Number of blocks</param>
+        /// <param name="path">Path</param>
+        [ConsoleCommand("export blocks", Category = "Blockchain Commands")]
+        private void OnExportBlocksStartCountCommand(uint start, uint count = uint.MaxValue, string path = null)
+        {
+            if (Blockchain.Singleton.Height < start)
```
Tommo-L: So we still don't support `start = 0` as the default value. 😂

shargon: We can change the logic of some commands later

superboyiii:
@erikzhang Seems the Git Check stops, could you fix it?

shargon:
@erikzhang could you merge this PR?

erikzhang:
Is it tested?

shargon:
> Is it tested?

Yes, it's tested from my side, we can wait until @superboyiii and @nicolegys approvals.

superboyiii:
@shargon @erikzhang  From my side, it passed our manual test. We're now trying to use automatic test tools to verify it more seriously. Give me three days to get final result. If it's OK, we could release it on Friday.

shargon:
@superboyiii take your time

superboyiii:
@shargon @erikzhang Test result shows it works well.

PR #545, opened 2020-02-26
Add cosigner

doubiliu:
Currently,in `OnInvokeCommand`,we use `new cosigner[0] ` as a inital paramter.This will caluse falut in `CheckWitness`.

We need to add the relevant cosigner

### DIFF
--- a/neo-cli/CLI/MainService.cs
+++ b/neo-cli/CLI/MainService.cs
@@ -281,25 +281,34 @@ namespace Neo.CLI
         private bool OnInvokeCommand(string[] args)
         {
             var scriptHash = UInt160.Parse(args[1]);
-            var specifyAddress = new String[args.Length - 2];
-            Array.ConstrainedCopy(args, 2, specifyAddress, 0, args.Length - 2);
-
             List<Cosigner> signCollection = new List<Cosigner>();
-            using (SnapshotView snapshot = Blockchain.Singleton.GetSnapshot())
+            if (CurrentWallet != null)
             {
-                UInt160[] accounts = CurrentWallet.GetAccounts().Where(p => !p.Lock && !p.WatchOnly).Select(p => p.ScriptHash).Where(p => NativeContract.GAS.BalanceOf(snapshot, p).Sign > 0).ToArray();
-                foreach (var signAccount in accounts)
+                var witnessAddress = args.Last();
+                try
                 {
-                    if (specifyAddress.Contains(signAccount.ToAddress()))
+                    witnessAddress.ToScriptHash();
+                }
+                catch
+                {
+                    witnessAddress = null;
+                }
+
+                using (SnapshotView snapshot = Blockchain.Singleton.GetSnapshot())
+                {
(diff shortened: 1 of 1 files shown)


neo-cli/CLI/MainService.cs
```diff
@@ -282,6 +282,16 @@ private bool OnInvokeCommand(string[] args)
         {
             var scriptHash = UInt160.Parse(args[1]);
 
+            List<Cosigner> signCollection = new List<Cosigner>();
```
shargon: We should be able to choose wich cosigner we want, imagine a wallet with 1.000 of address, we can't add all of them

doubiliu: ｍｍｍ．I have thought about this.But generally only those exchanges will create a lot of address and they will not use this node to invoke command.
@superboyiii  Can you help explain it together?

shargon: Then with only the first one with balance it's enough?

doubiliu: So we need to use `manual witness` first to make the pre-execution pass, calculate the contract's execution consumption, and then choose an account with a sufficient amount from the available accounts to pay the fees. Is this equal to implementing the logic about UTXO in NEO 2 again?

shargon: I think that the user should specify which account and with which scope want to sign

Lichen9618: > Then with only the first one with balance it's enough?

Take first one as default is dangerous,
If the contract is not source-open, it can steal asset from user easily.

I vote for the user can manually specify which account and with which scope.
But we need a method to make it easy for using

doubiliu: Great proposal.I think I can implement the specified account first, the scope is a bit complicated, and in general, the scope of this invoke script should be fixed

shargon: Maybe duplicating the methods it's easier InvokeGlobal, InvokeCalledByEntry...

doubiliu: I think `CalledByEntry` is enough.It just a little script to invoke contract function.No other script will call this script.

neo-cli/CLI/MainService.cs
```diff
@@ -281,6 +281,26 @@ private bool OnDeployCommand(string[] args)
         private bool OnInvokeCommand(string[] args)
         {
             var scriptHash = UInt160.Parse(args[1]);
+            var specifyAddress = new String[args.Length - 2];
+            Array.ConstrainedCopy(args, 2, specifyAddress, 0, args.Length - 2);
+
+            List<Cosigner> signCollection = new List<Cosigner>();
+            using (SnapshotView snapshot = Blockchain.Singleton.GetSnapshot())
+            {
+                UInt160[] accounts = CurrentWallet.GetAccounts().Where(p => !p.Lock && !p.WatchOnly).Select(p => p.ScriptHash).Where(p => NativeContract.GAS.BalanceOf(snapshot, p).Sign > 0).ToArray();
```
superboyiii: Open wallet isn't necessary for invoking, this will cause exception.

Tommo-L:
Recommend to fix this problem in/after #536

superboyiii:
I'll close this.

PR #546, opened 2020-02-26
Show utilitary methods for Neo-Cli

Thacryba:
This is an update from #455 and complementing #535.

Some changes were made:

- Adding the command show last-block:
image

- Adding the block index on show last-transactions:
image

- Changing the time format in show block:
image

It can be merged after #536.


superboyiii:
@Thacryba I think the second function is very useful. However, it hasn't decompile tx compeletely. So op codes haven't be recognized yet. And since the json output can be very long in cli, I think it's more suitable to be made into rpcserver plugin and the first one and the third one are already existed. I recommend to improve the second one and made txid as a param in RPC.

neo-cli/CLI/MainService.Show.cs
```diff
@@ -0,0 +1,325 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Prints the SmartContract ABI
+        /// </summary>
+        /// <param name="contractHash"></param>
+        /// <returns></returns>
+        [ConsoleCommand("show contract", Category = "Show Commands")]
+        private void OnShowContract(string contractHash)
```
vncoelho, line 75: Perhaps this could be on the Plugin and also be called through RPC.

neo-cli/CLI/MainService.Show.cs
```diff
@@ -0,0 +1,325 @@
...
+{
+    partial class MainService
+    {
+        /// <summary>
+        /// Prints the last transactions in the blockchain
+        /// </summary>
+        /// <param name="desiredCount"></param>
+        /// <returns></returns>
+        [ConsoleCommand("show last-transactions", Category = "Show Commands")]
+        private void OnShowLastTransactions(uint desiredCount)
```
vncoelho, line 20: Perhaps this could be on the Plugin and also be called through RPC.

PR #552, opened 2020-03-18
Change help for import multisigaddress

chenzhitong:


bettybao1209:
Greate job

shargon:
I will be replaced with #536

erikzhang:
So close it?

shargon:
I think that it will only will produce conflicts.

Issue #556, opened 2020-04-01
Cannot call plugin commands


nicolegys:
I failed to call the plugin commands when testing https://github.com/neo-project/neo-modules/pull/188. 
Then I found that, now in neo-cli, it try to register plugin commands before the plugins are loaded, so none of them could be registered.
![image](https://user-images.githubusercontent.com/43694095/78109091-27e5aa80-742b-11ea-92a0-2d20e4d6844a.png)
@shargon Could you help fix it?

shargon:
I will fix it

PR #557, opened 2020-04-01
Fix plugin commands

shargon:
Close https://github.com/neo-project/neo-node/issues/556

### DIFF
--- a/Neo.ConsoleService/CommandToken.cs
+++ b/Neo.ConsoleService/CommandToken.cs
@@ -54,14 +54,10 @@ namespace Neo.ConsoleService
                     case '"':
                     case '\'':
                         {
-                            if (lastToken is CommandQuoteToken quote)
+                            // "'"
+                            if (lastToken is CommandQuoteToken quote && quote.Value[0] != commandLine[index])
                             {
-                                // "'"
-
-                                if (quote.Value[0] != commandLine[index])
-                                {
-                                    goto default;
-                                }
+                                goto default;
                             }
 
                             lastToken = CommandQuoteToken.Parse(commandLine, ref index);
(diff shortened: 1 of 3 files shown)


nicolegys:
key.ToLower cannot get the value
image

shargon:
@nicolegys please take a look again

nicolegys:
Thank you @shargon . Now it works well.
Do we need to display all the command help information including the plugins when we call `help`?
Now it looks like this (`StatesDumper` and `PerformanceMonitor` are installed):
a bit messy
Or maybe we need to put all the plugin comands in Category `Plugin Commands`?
image

shargon:
Both options are good to me, what do you prefer? it's enough with change the category of the plugin command.

nicolegys:
I prefer to have plugin commands in the `Plugin Commands` catagory, to make the help information more clearly. And whether the plugins are installed or not, they will not affect the categories of neo-cli itself.
```
Base Commands:
...
Blockchain Commands:
...
Consensus Commands:
...
Contract Commands:
...
Network Commands:
...
Node Commands:
...
Plugin Commands:
...
Wallet Commands:
...
```

But maybe the plugin commands need to have their own catagories，like this 
```
neo> help PerformanceMonitor
Block Commands:
        block avgtime [blockCount=1000]
        block sync
        block time <blockIndexOrHash>
        block timesincelast
Check Commands:
        check cpu
        check memory
        check threads
Consensus Commands:
        commit time
        confirmation time
        payload time
Network Commands:
        connected
        ping [ipaddress=null]
        rpc time <url>
Transaction Commands:
        tx avgsize [txCount=1000]
        tx size <transactionHash>
```
I'm not sure if changing all of their categories to `Plugin Commands` will make the help information more friendly. Or maybe we need subcategories? What do you think?

PR #562, opened 2020-04-10
Fix Parse manifestFilePath

cloud8little:
https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Manifest/ContractManifest.cs#L120


cloud8little:
close since #563 fix the same issue.

PR #569, opened 2020-04-16
3.0.0-preview2

Tommo-L:


neo-cli/CLI/Helper.cs
```diff
@@ -19,5 +29,51 @@ public static bool IsYes(this string input)
 
             return input == "yes" || input == "y";
         }
+
+        public static JObject ToJson(this StackItem item)
```
shargon, line 33: Why don't move this to `neo` it's duplicated in neo-modules, and `neo-node`

Tommo-L: It was fixed in https://github.com/neo-project/neo/pull/1569, but not merged in neo-preview2.

Tommo-L:
Neo-preview2 has been released.

neo-cli/CLI/Helper.cs
mutrofin517: AQtk8ckQwnZTowVgs6zgCZEJyUQyM1z3ew

PR #577, opened 2020-05-01
Remove log logic

shargon:
Required for https://github.com/neo-project/neo/issues/1245
Require https://github.com/neo-project/neo/pull/1626


neo-cli/Logger.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Neo
+{
+    class Logger : Plugin, ILogPlugin
+    {
+        private static readonly object _lock = new object();
+
+        static Logger()
+        {
+            AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
```
erikzhang, line 13: Maybe we can handle this event in neo core?

shargon: save the errors in neo? I like it

shargon: If we merge https://github.com/neo-project/neo/pull/1626 I will remove the logic from this PR

erikzhang: No. I mean we handle this event in neo core, and send the message to the log stream. Then persist it in plugins.

erikzhang: Then we can remove the `error.log`. The logs are unified.

PR #561, opened 2020-04-08
Update Neo & enable change password

Qiao-Jin:
Close #558

PR #578, opened 2020-05-03
Backup Wallet on change password

belane:
Password change (#561) should be harmless but I think it is advisable to make a backup before to avoid losses due to disk corruption, power failures, ...
* Needs https://github.com/neo-project/neo/pull/1627

### DIFF
--- a/neo-cli/CLI/MainService.Wallet.cs
+++ b/neo-cli/CLI/MainService.Wallet.cs
@@ -482,6 +482,15 @@ namespace Neo.CLI
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


superboyiii:
I've merged https://github.com/neo-project/neo/pull/1627 on my local neo, and tested this PR, it works as expect.
1. Create a nep6 format wallet and change password, expect [*.bak appears]
2. Open bak wallet and list related info(pubkey, address)
3. Rechange password, expect [Wallet backup fail]
Test Result: `PASS`

shargon:
could you update the nuget?

PR #582, opened 2020-05-12
Move SystemLog plugin into neo-cli as a native logger with on/off functionalities.

joeqian10:
This fixes neo-modules issue [228](https://github.com/neo-project/neo-modules/issues/228).

### DIFF
--- a/neo-cli/CLI/Logger.cs
+++ b/neo-cli/CLI/Logger.cs
@@ -84,7 +84,15 @@ namespace Neo.CLI
                     var path = Combine(Settings.Default.Logger.Path, sb.ToString());
                     Directory.CreateDirectory(path);
                     path = Combine(path, $"{now:yyyy-MM-dd}.log");
-                    File.AppendAllLines(path, new[] { $"[{level}]{log}" });
+                    try
+                    {
+                        File.AppendAllLines(path, new[] { $"[{level}]{log}" });
+                    }
+                    catch (IOException)
+                    {
+                        Console.WriteLine("Error writing the log file: " + path);
+
+                    }
                 }
             }
         }
--- a/neo-cli/CLI/Logger.cs
+++ b/neo-cli/CLI/Logger.cs
@@ -1,10 +1,8 @@
 using Neo.Plugins;
 using System;
-using System.Collections.Generic;
 using System.Text;
 using System.IO;
 using static System.IO.Path;
-using Microsoft.Extensions.Configuration;
 using System.Reflection;
 using Neo.ConsoleService;
 
@@ -91,7 +89,6 @@ namespace Neo.CLI
                     catch (IOException)
                     {
                         Console.WriteLine("Error writing the log file: " + path);
-
                     }
                 }
             }
--- a/neo-cli/CLI/Logger.cs
+++ b/neo-cli/CLI/Logger.cs
@@ -6,8 +6,9 @@ using System.IO;
 using static System.IO.Path;
 using Microsoft.Extensions.Configuration;
 using System.Reflection;
+using Neo.ConsoleService;
 
-namespace Neo.SystemLog
+namespace Neo.CLI
 {
     public class Logger : Plugin, ILogPlugin
     {
@@ -18,7 +19,7 @@ namespace Neo.SystemLog
 
         public Logger() : base()
         {
-            Started = true; // default is started to log
+            Started = Settings.Default.Logger.Started; // default is started to log
         }
 
         private static void GetErrorLogs(StringBuilder sb, Exception ex)
--- a/neo-cli/CLI/MainService.Node.cs
+++ b/neo-cli/CLI/MainService.Node.cs
@@ -86,23 +86,5 @@ namespace Neo.CLI
             Console.WriteLine();
             Console.CursorVisible = true;
         }
-
-        /// <summary>
-        /// Start the default ILogPlugin
-        /// </summary>
-        [ConsoleCommand("log on", Category = "Node Commands")]
-        private void OnLogOnCommand()
-        {
-            this.logger.Started = true;
-        }
-
-        /// <summary>
-        /// Stop the default ILogPlugin
-        /// </summary>
-        [ConsoleCommand("log off", Category = "Node Commands")]
-        private void OnLogOffCommand()
-        {
-            this.logger.Started = false;
-        }
     }
 }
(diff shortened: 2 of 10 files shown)


neo-cli/CLI/MainService.Node.cs
```diff
@@ -86,5 +86,23 @@ private void OnShowStateCommand()
             Console.WriteLine();
             Console.CursorVisible = true;
         }
+
+        /// <summary>
+        /// Start the default ILogPlugin
+        /// </summary>
+        [ConsoleCommand("log on", Category = "Node Commands")]
+        private void OnLogOnCommand()
```
erikzhang, line 94: Why do we need these commands?

neo-cli/CLI/MainService.cs
```diff
@@ -65,6 +66,8 @@ private set
         protected override string Prompt => "neo";
         public override string ServiceName => "NEO-CLI";
 
+        private readonly Logger logger = new Logger();
```
erikzhang, line 69: The plugin should be initialized in the `Start()` method. (After the initialization of `NeoSystem`)

neo-cli/SystemLog/Logger.cs
```diff
@@ -0,0 +1,89 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+using static System.IO.Path;
+using Microsoft.Extensions.Configuration;
+
+namespace Neo.SystemLog
+{
+    public class Logger : Plugin, ILogPlugin
```
erikzhang, line 11: You also need to override `ConfigFile` and `Path`.

neo-cli/SystemLog/ConsoleColorSet.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Neo.SystemLog
+{
+    internal class ConsoleColorSet
```
shargon, line 7: Move this to https://github.com/neo-project/neo-node/tree/master/Neo.ConsoleService ?

neo-cli/CLI/MainService.Node.cs
joeqian10: We need these to toggle log function when neo-cli is running as @shargon mentioned. Do you have any better ideas?

neo-cli/CLI/MainService.cs
joeqian10: I have moved the initialization into `Start()` method.

neo-cli/CLI/MainService.cs
```diff
@@ -65,6 +66,8 @@ private set
         protected override string Prompt => "neo";
         public override string ServiceName => "NEO-CLI";
 
+        private Logger logger;
```
shargon, line 69:
```suggestion
        private readonly Logger logger;
```

neo-cli/SystemLog/Logger.cs
joeqian10: I have overriden `ConfigFile` but `Path` is not marked as `virtual`.

neo-cli/CLI/MainService.cs
erikzhang: It can't be `readonly` because it is not initialized in constructor.

joeqian10: If you mark it as `readonly`, a readonly field cannot be assigned to (except in a constructor of the class in which the field is defined or a variable initializer). And I have put the initialization in the `Start()` method.

neo-cli/CLI/MainService.Node.cs
erikzhang: I mean why do we need to toggle log function when neo-cli is running? Maybe a command-line argument is enough. Example: `--log-on` or `--log-off`.

shargon: If we have the config inside the json, I think that it's not needed

neo-cli/SystemLog/Logger.cs
erikzhang: It should be. We need to modify the `Path` to be `virtual`.

neo-cli/SystemLog/ConsoleColorSet.cs
joeqian10: Sure, no problem.

neo-cli/CLI/MainService.Node.cs
joeqian10: OK, I will add a configuration section inside `config.json`.

neo-cli/CLI/Logger.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                if (!string.IsNullOrEmpty(Settings.Default.Logger.Path))
+                {
+                    StringBuilder sb = new StringBuilder(source);
+                    foreach (char c in GetInvalidFileNameChars())
+                        sb.Replace(c, '-');
+                    var path = Combine(Settings.Default.Logger.Path, sb.ToString());
+                    Directory.CreateDirectory(path);
+                    path = Combine(path, $"{now:yyyy-MM-dd}.log");
+                    File.AppendAllLines(path, new[] { $"[{level}]{log}" });
```
superboyiii, line 87:
```suggestion
                    try
                    {
                        File.AppendAllLines(path, new[] { $"[{level}]{log}" });
                    }
                    catch (IOException)
                    {
                        Console.WriteLine("Error writing the log file: " + path);

                    }
```

superboyiii:
@joeqian10 Unhandled exceptions are not recorded in System log.
image
Now:
image
Before:
image

neo-cli/CLI/Logger.cs
```diff
@@ -0,0 +1,100 @@
+using Neo.Plugins;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+using static System.IO.Path;
+using Microsoft.Extensions.Configuration;
+using System.Reflection;
+using Neo.ConsoleService;
```
superboyiii, line 9:
```suggestion
using Neo.Plugins;
using System;
using System.Text;
using System.IO;
using static System.IO.Path;
using System.Reflection;
using Neo.ConsoleService;
```
Delete unused dependency.

erikzhang:
Why merge it so quickly? I didn't finish my review.

erikzhang:
Reverted.

Issue #583, opened 2020-05-12
Some params don't have default values in settings


superboyiii:
Some params don't have default values in Settings.cs. For exmaple, `Port` and `WsPort`. We should set default settings to make it work even if config.json is missing in workspace.

PR #584, opened 2020-05-12
Fill default settings

superboyiii:
Close https://github.com/neo-project/neo-node/issues/583

### DIFF
--- a/neo-cli/Settings.cs
+++ b/neo-cli/Settings.cs
@@ -67,7 +67,7 @@ namespace Neo
         public P2PSettings(IConfigurationSection section)
         {
             this.Port = ushort.Parse(section.GetValue("Port", "10333"));
-            this.WsPort = ushort.Parse(section.GetValue("WPort", "10334"));
+            this.WsPort = ushort.Parse(section.GetValue("WsPort", "10334"));
             this.MinDesiredConnections = section.GetValue("MinDesiredConnections", Peer.DefaultMinDesiredConnections);
             this.MaxConnections = section.GetValue("MaxConnections", Peer.DefaultMaxConnections);
             this.MaxConnectionsPerAddress = section.GetValue("MaxConnectionsPerAddress", 3);
--- a/neo-cli/Settings.cs
+++ b/neo-cli/Settings.cs
@@ -85,10 +85,10 @@ namespace Neo
         {
             if (section.Exists())
             {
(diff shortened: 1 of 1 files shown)


shargon:
Could you change also

https://github.com/neo-project/neo-node/pull/584/files#diff-0327b59f4dd09986f49db7500602d9a9R88-R91

superboyiii:
@shargon Change for what?

shargon:
Add the default value, if the section exists, but the values don't

superboyiii:
@shargon I've already add the default values in this PR :) Or which param else do you think is need to have default value as well?

neo-cli/Settings.cs
```diff
@@ -52,7 +52,7 @@ public class StorageSettings
 
         public StorageSettings(IConfigurationSection section)
         {
-            this.Engine = section.GetSection("Engine").Value;
+            this.Engine = section.GetValue("Engine", "LevelDBStore");
```
erikzhang, line 55: The default value should be `MemoryStore`. Because `LevelDBStore` may be not installed.

superboyiii: But due to config.json's default settings, it's using `LevelDBStore` but not `MemoryStore`, or shall  I adjust config.json as well?
https://github.com/neo-project/neo-node/blob/605fccbef4150118b78f112785eda40de7fc336b/neo-cli/config.json#L4

shargon:
>  I've already add the default values in this PR :) Or which param else do you think is need to have default value as well?

Lines 88-91 https://github.com/neo-project/neo-node/pull/584/files#diff-0327b59f4dd09986f49db7500602d9a9R88-R91

superboyiii:
> > I've already add the default values in this PR :) Or which param else do you think is need to have default value as well?
> 
> Lines 88-91 https://github.com/neo-project/neo-node/pull/584/files#diff-0327b59f4dd09986f49db7500602d9a9R88-R91

The initial value of `public bool IsActive { get; }` is always false. So that it will not influence cli running if the params of `UnlockWalletSettings` is null.
image
But add default value is more safe. I'll add it.

shargon:
dotnet format please

superboyiii:
@shargon Could you fix this? I don't know much about it.
image

shargon:
> Could you fix this? I don't know much about it.

Done, re-run and it works

PR #594, opened 2020-05-27
Fix relay tx

Tommo-L:
Affected by https://github.com/neo-project/neo/pull/1667, we need to change from `LocalNode` to `Blockchain` to relay tx.


cloud8little:
Smoke test passed.
1. Single CN consensus
2. send neo/gas
3. deploy nep5 contract.
4. sign and relay.

PR #596, opened 2020-05-28
[neox-2.x]Upgrade to Neo 2.10.3-neox-preview1

superboyiii:
Upgrade to **Neo 2.10.3-neox-preview1**


superboyiii:
@shargon Will upgrade testnet on June 2nd. I set `StateRootEnableIndex` on 4420000 so that it could have more than 5 days buffer after next Tuesday.

PR #604, opened 2020-06-08
Add description and uninstall restriction for "SystemLog"

### LINKED ISSUE #601: An exception when uninstall SystemLog
After https://github.com/neo-project/neo-node/pull/587, `SystemLog` plugin's file is neo-cli.dll, `uninstall SystemLog`  command will throw an exception because can't deleted itself.

exception:
```
System.UnauthorizedAccessException
  HResult=0x80070005
  Message=Access to the path 'D:\neo-node\neo-cli\bin\Debug\netcoreapp3.0\neo-cli.dll' is denied.
  Source=System.IO.FileSystem
  StackTrace:
   at System.IO.FileSystem.DeleteFile(String fullPath)
   at System.IO.File.Delete(String path)
   at Neo.CLI.MainService.OnUnInstallCommand(String pluginName) in D:\neo-node\neo-cli\CLI\MainService.Plugins.cs:line 81
```

joeqian10:
Close #601 

This may look stupid.🤣 Maybe we should not implement "SystemLog" as a plugin, but which needs modification from neo-core: 

```c#
        public static void Log(string source, LogLevel level, object message)
        {
            foreach (ILogPlugin plugin in Plugin.Loggers)
                plugin.Log(source, level, message);
        }
```

### DIFF
--- a/neo-cli/CLI/Logger.cs
+++ b/neo-cli/CLI/Logger.cs
@@ -16,6 +16,7 @@ namespace Neo.CLI
         private static readonly ConsoleColorSet FatalColor = new ConsoleColorSet(ConsoleColor.Red);
 
         public override string Name => "SystemLog";
+        public override string Description => "Prints consensus log and is a built-in plugin which cannot be uninstalled";
         public override string ConfigFile => Combine(GetDirectoryName(Path), "config.json");
         public override string Path => GetType().Assembly.Location;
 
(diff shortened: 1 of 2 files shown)


shargon:
Maybe you want to save the logs in a syslog server or similar, I think that it's not bad.

ProDog:
Now `SystemLog` is not a plugin, so should we remove it from plugins?

erikzhang:
I think you should not list it as a plugin. Because it is a part of the cli, not a plugin, although it implements the plugin interface.

PR #613, opened 2020-06-24
Fix invoke command

Tommo-L:
Close https://github.com/neo-project/neo-node/issues/614 Close https://github.com/neo-project/neo/issues/1735

Don't have to check the balance of cosiginers.

@cloud8little plz, have a check.

PR #616, opened 2020-06-29
Remove dead code

shargon:
Already checked on line 56


Tommo-L:
Already fixed in https://github.com/neo-project/neo-node/pull/613

PR #628, opened 2020-07-20
Allow smart contract verification

shargon:
As seen in the talk with @lock9 

- Allow to see watch only address
- Allow to use smart contract as sender
- Remove unused method.

Require https://github.com/neo-project/neo/pull/1800


neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -53,7 +53,7 @@ private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contra
         {
             Signer[] signers = Array.Empty<Signer>();
             if (signerAccounts != null && !NoWallet())
-                signers = CurrentWallet.GetAccounts().Where(p => !p.Lock && !p.WatchOnly && signerAccounts.Contains(p.ScriptHash)).Select(p => new Signer() { Account = p.ScriptHash, Scopes = WitnessScope.CalledByEntry }).ToArray();
```
superboyiii, line 56: Why `WatchOnly` address can be signer?

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -53,7 +48,7 @@ private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contra
...
+                signers = CurrentWallet.GetAccounts().Where(p => !p.Lock && signerAccounts.Contains(p.ScriptHash)).Select(p => new Signer() { Account = p.ScriptHash, Scopes = WitnessScope.CalledByEntry }).ToArray();
```
superboyiii, line 51: Why WatchOnly address can be signer?

shargon: Withdrawal from a contract require a watch only address.

shargon:
@lock9 could you confirm that it solve your problem?

neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -284,6 +284,57 @@ private void OnImportKeyCommand(string wifOrFile)
...
+        [ConsoleCommand("import watchonly", Category = "Wallet Commands")]
+        private void OnImportWatchOnlyCommand(string addressOrFile)
+        {
+            UInt160 address = null;
+            try
+            {
+                address = StringToAddress(addressOrFile);
+            }
+            catch (FormatException) { }
+            if (address == null)
```
Tommo-L, line 299:
```suggestion
            if (address is null)
```

Tommo-L:
If we want to allow to use smart contract as sender, we must modify neo-core first.

https://github.com/neo-project/neo/blob/master/src/neo/Wallets/Wallet.cs#L252

https://github.com/neo-project/neo/blob/master/src/neo/Wallets/Wallet.cs#L320

https://github.com/neo-project/neo/blob/master/src/neo/Wallets/Wallet.cs#L396

https://github.com/neo-project/neo/blob/master/src/neo/Wallets/Wallet.cs#L402

lock9:
@shargon  I can only test this tomorrow morning, by only checking the code I can't ensure it works

shargon:
> If we want to allow to use smart contract as sender, we must modify neo-core first.

You should add as your smart contract as watch only first.

ProDog:
I have tested and it throw an exception, if I `import watchonly` a smart contract hash, because it's key is null.
```
neo> import watchonly 0xb6770c896f96f3d85381ec45a403f37d18f6d714
Address: NMpBQr7KMPoVd82LeJbVYnH35m8jxWvhZh
```

image

shargon:
Yes, we can't know the pubkey

meevee98:
I've tried to withdraw gas from a smart contract by using the transfer command.
The validation is working well, until the moment the cli asks for relaying the transaction. If I relay, it fails to send the transaction.

image

As @Tommo-L [has said](https://github.com/neo-project/neo-node/pull/628#issuecomment-663335254), the neo-core must be changed as well. I've been researching and this fail is because the ``Wallet.Sign``, that only has validation for multi signature or accounts with public key. Including a smart contract as a watch-only address doesn't include a public key, so a transaction signing fails.

https://github.com/neo-project/neo/blob/daa6e7a3b95524534107cfb6e174d9a52f3cd617/src/neo/Wallets/Wallet.cs#L414-L440

shargon:
@meevee98 @Tommo-L please, could you check https://github.com/neo-project/neo/pull/1800 ?

ProDog:
I have tested it OK.
But I have a question, why must `import watchonly` first if add contract as verification? For developers, they would like to make it easier to use this, such as using the SDK to construct this transaction.

shargon:
@ProDog now with https://github.com/neo-project/neo/pull/1800#discussion_r463935448 (https://github.com/neo-project/neo/pull/1800/commits/ab971cede24d08dca66986ecc201ca7c0208023d) it's not required to add the contract as `watch-only`.

ProDog:
@erikzhang this need to merge, please review it at sometime.

erikzhang:
Compilation error.

shargon:
merge?

PR #630, opened 2020-07-20
Get innerException message Recursively

### LINKED ISSUE #629: Deploy an existed contract only return "engine faulted"
Prepare a wallet with gas, deploy a contract twice.
8959a4053c9ce5c0ec578f6442f94d9

cloud8little:
fix #629


neo-cli/CLI/MainService.cs
```diff
@@ -578,7 +578,7 @@ static string GetExceptionMessage(Exception exception)
 
             if (exception.InnerException != null)
             {
-                return exception.InnerException.Message;
+                return GetExceptionMessage(exception.InnerException);
```
shargon, line 581: Require both changes? or it's enough with the other?

cloud8little: It returns "Exceptions has been thrown by the target of an invocation" if calling current GetExceptionMessage, because Deploy (catch) -> MakeTransaction (throw) -> ApplicationEngine.Run (throw) -> createContract (throw)
b3d93acdeb6f2e17360978c1f3bca78

shargon: We can move `GetExceptionMessage` to `ConsoleServiceBase` and change https://github.com/neo-project/neo-node/blob/dec52fbd25850113f5372bf8d2aa7c97a6ffcced/Neo.ConsoleService/ConsoleServiceBase.cs#L615

Issue #627, opened 2020-07-20
Remove neo-gui from NEO3


shargon:
I know that @neo-project/ngd-shanghai are working on a new version of neo-gui, currently neo-gui is not adapted for neo3 changes, so I propose to remove this code until @neo-project/ngd-shanghai finish the new gui version.

PR #631, opened 2020-07-22
Remove neo-gui

shargon:
Close https://github.com/neo-project/neo-node/issues/627


shargon:
@erikzhang what do you think?

erikzhang:
Why remove GUI?

shargon:
I think that it's not updated. For example deploy a contract doesn't work, it doesn't compose the manifest https://github.com/neo-project/neo-node/blob/dec52fbd25850113f5372bf8d2aa7c97a6ffcced/neo-gui/GUI/DeployContractDialog.cs#L19 But we can wait until finish the ngd version.

PR #637, opened 2020-07-31
update list address and list key

ProDog:
update list address and list key


ProDog:
PR for https://github.com/neo-project/neo-node/pull/628.

PR #646, opened 2020-08-10
Remove Travis

superboyiii:
Remove travis since it never works and matter merging.


superboyiii:
@shargon Could you do this again? https://github.community/t/expected-waiting-for-status-to-be-reported/18001/3

PR #659, opened 2020-09-07
[master-2.x] Add minimum network fee

Qiao-Jin:
This PR adds a adjustable MinimumNetworkFee into protocol setting.
Related to [neo-core #1901](https://github.com/neo-project/neo/pull/1901)

### DIFF
--- a/neo-cli/protocol.json
+++ b/neo-cli/protocol.json
@@ -5,6 +5,7 @@
     "SecondsPerBlock": 15,
     "LowPriorityThreshold": 0.001,
     "StateRootEnableIndex": 6016000,
+    "MinimumNetworkFee": 0,
     "StandbyValidators": [
       "03b209fd4f53a7170ea4444e0cb0a6bb6a53c2bd016926989cf85f9b0fba17a70c",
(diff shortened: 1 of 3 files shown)


erikzhang:
If the consensus nodes change the policy of `MinimumNetworkFee`, how do other nodes know it?

Qiao-Jin:
> If the consensus nodes change the policy of `MinimumNetworkFee`, how do other nodes know it?

Each time upon changing a public announce can be published
A `MinimumNetworkFee ` RPC call can also be used to know the `MinimumNetworkFee` of seed nodes.

erikzhang:
But the seed nodes are different from the consensus nodes.

Qiao-Jin:
> But the seed nodes are different from the consensus nodes.

That's true, but consensus node cannot expose network ID so they cannot broadcast in public by themselves... unless they publish some posts onchain in terms of transactions, but that's another story.

erikzhang (approved):

shargon (approved):

vncoelho (approved):

PR #662, opened 2020-09-12
Remove Get validators

shargon:
Close https://github.com/neo-project/neo/issues/1926

### DIFF
--- a/neo-cli/CLI/MainService.Vote.cs
+++ b/neo-cli/CLI/MainService.Vote.cs
@@ -107,19 +107,19 @@ namespace Neo.CLI
         }
 
         /// <summary>
-        /// Process "get validators"
+        /// Process "get next validators"
         /// </summary>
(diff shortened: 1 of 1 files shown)


erikzhang (approved):

cloud8little (commented):
there is already get next validators. I think we could remove one. https://github.com/neo-project/neo-node/blob/50324a1c5075ce71d55bca63ed78d43f8a6f6433/neo-cli/CLI/MainService.Vote.cs#L156-L161

shargon:
@cloud8little good point! it was removed because it's already defined here https://github.com/neo-project/neo-node/blob/5c329bbe23854f805a9914d18c626039a92cde45/neo-cli/CLI/MainService.Vote.cs#L113-L129

Tommo-L (approved):

cloud8little:
Looks good to me.

PR #663, opened 2020-09-15
Modify FreeGasChangeHeight

superboyiii:
Modify FreeGasChangeHeight


superboyiii:
We'd like to upgrade mainnet on Sep 19th so that the height is modified.

PR #676, opened 2020-10-22
Sync neo changes

superboyiii:
Related: https://github.com/neo-project/neo/pull/2044
Close https://github.com/neo-project/neo-node/issues/685
***Please dont't merge until related PRs are all merged.***

PR #680, opened 2020-11-21
Fix contract hash

Tommo-L:
wait https://github.com/neo-project/neo/pull/2044

### DIFF
--- a/neo-cli/CLI/MainService.Contracts.cs
+++ b/neo-cli/CLI/MainService.Contracts.cs
@@ -1,7 +1,9 @@
 using Neo.ConsoleService;
 using Neo.IO.Json;
 using Neo.Network.P2P.Payloads;
+using Neo.SmartContract;
 using Neo.SmartContract.Native;
+using Neo.VM;
 using System;
 using System.Linq;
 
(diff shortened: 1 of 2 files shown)


neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -30,7 +32,16 @@ private void OnDeployCommand(string filePath, string manifestPath = null)
                 Console.WriteLine("Error: " + GetExceptionMessage(e));
                 return;
             }
-            Console.WriteLine($"Script hash: {scriptHash.ToString()}");
+
+            // TODO need to replace by ContractHash method
+
+            using var sb = new ScriptBuilder();
```
shargon, line 38: `ContractCreate` push the contract to the stack, so we can read the hash from the execution result

Tommo-L: That's too bad, we should have a simple way to calculate contract hash, not by vm execution.

superboyiii:
I will merge this first, then we could review in https://github.com/neo-project/neo-node/pull/676.

PR #682, opened 2020-11-25
Max traceable blocks

chenzhitong:


shargon (approved):

neo-cli/protocol.mainnet.json
```diff
@@ -2,6 +2,7 @@
   "ProtocolConfiguration": {
     "Magic": 5195086,
     "MillisecondsPerBlock": 15000,
+    "MaxTraceableBlocks": 5760,
```
shargon, line 5: By default it's other value
https://github.com/neo-project/neo/blob/402e9b19d80bb9093601f5ac57ff0cdc3c6cf6ab/src/neo/ProtocolSettings.cs#L98

shargon (changes requested):

vncoelho (commented):
It should be something around `2102400` equals to 365 days.

neo-cli/protocol.json
```diff
@@ -2,6 +2,7 @@
   "ProtocolConfiguration": {
     "Magic": 5195086,
     "MillisecondsPerBlock": 15000,
+    "MaxTraceableBlocks": 5760,
```
vncoelho, line 5: This is one day currently, right?
If nodes miss syncronization it would require `import`

shargon (approved):

Issue #708, opened 2020-12-29
neo-cli doesn't compile against neo master


devhawk:
```
node\neo-cli\CLI\MainService.cs(285,20): error CS1061: 
'ScriptBuilder' does not contain a definition for 'EmitAppCall' and no accessible extension method 'EmitAppCall' accepting a first argument of type 'ScriptBuilder' could be found (are you missing a using directive or an assembly reference?) [node\neo-cli\neo-cli.csproj]
node\neo-cli\CLI\MainService.cs(532,31): error CS1061: 
'ScriptBuilder' does not contain a definition for 'EmitAppCall' and no accessible extension method 'EmitAppCall' accepting a first argument of type 'ScriptBuilder' could be found (are you missing a using directive or an assembly reference?) [node\neo-cli\neo-cli.csproj]
node\neo-cli\CLI\MainService.Vote.cs(46,31): error CS1061: 'ScriptBuilder' does not contain a definition for 'EmitAppCall' and no accessible extension method 'EmitAppCall' accepting a first argument of type 'ScriptBuilder' could be found (are you missing a using 
directive or an assembly reference?) [node\neo-cli\neo-cli.csproj]
node\neo-cli\CLI\MainService.Vote.cs(86,31): error CS1061: 'ScriptBuilder' does not contain a definition for 'EmitAppCall' and no accessible extension method 'EmitAppCall' accepting a first argument of type 'ScriptBuilder' could be found (are you missing a using 
directive or an assembly reference?) [node\neo-cli\neo-cli.csproj]
node\neo-cli\CLI\MainService.Vote.cs(110,31): error CS1061: 'ScriptBuilder' does not contain a definition for 'EmitAppCall' and no accessible extension method 
'EmitAppCall' accepting a first argument of type 'ScriptBuilder' could be found (are you missing a using directive or an assembly reference?) [node\neo-cli\neo-cli.csproj]
```

cloud8little:
should be resolved by #707

PR #709, opened 2021-01-11
Sync to Neo 3.0.0-CI01148

chenzhitong:
Close https://github.com/neo-project/neo-node/issues/708


shargon (approved):

neo-cli/CLI/MainService.cs
```diff
@@ -391,10 +391,6 @@ public async void Start(string[] args)
                 {
                     Console.WriteLine($"Failed to open file \"{Settings.Default.UnlockWallet.Path}\"");
                 }
-                if (Settings.Default.UnlockWallet.StartConsensus && CurrentWallet != null)
```
shargon, line 394: We should remove `StartConsensus` from settings, it will come in the plugin

vncoelho: It is already removed, yes?

vncoelho: Updated

shargon (approved):

vncoelho (approved):

neo-cli/CLI/MainService.cs
```diff
@@ -391,10 +391,6 @@ public async void Start(string[] args)
...
-                {
-                    OnStartConsensusCommand();
-                }
```
superboyiii, line 397: We should reserve this, it will implement `Plugin.GetService<IConsensusProvider>()?.Start(CurrentWallet);` from dBFT plugin, MainService.Consensus.cs should be reserved as well.

shargon: We should start the consensus with a new command, registered in the plugin

erikzhang: Agree with @shargon

shargon (dismissed):

erikzhang:
https://github.com/neo-project/neo-node/blob/3acb8aa42c42bc2817fd958444a7dab182c19082/neo-cli/Settings.cs#L99

This should be removed too.

erikzhang (approved):

shargon (approved):

PR #720, opened 2021-01-18
Sync with neo pr2215(LedgerContract)

chenzhitong:


neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -31,7 +31,7 @@ private void OnDeployCommand(string filePath, string manifestPath = null)
                 return;
             }
 
-            UInt160 hash = SmartContract.Helper.GetContractHash(tx.Sender, nef.Script);
+            UInt160 hash = SmartContract.Helper.GetContractHash(tx.Sender, nef.CheckSum, string.Empty);
```
cloud8little, line 34: string.Empty should be the name of the contract.

superboyiii, line 34:
```suggestion
            var manifest = ContractManifest.Parse(File.ReadAllBytes(manifestPath));
            UInt160 hash = SmartContract.Helper.GetContractHash(tx.Sender, nef.CheckSum, manifest.Name);
```

PR #727, opened 2021-01-25
Add test mode gas when invoking 

bettybao1209:
close https://github.com/neo-project/neo/issues/2260

### DIFF
--- a/Neo.ConsoleService/CommandToken.cs
+++ b/Neo.ConsoleService/CommandToken.cs
@@ -133,12 +133,18 @@ namespace Neo.ConsoleService
         /// <param name="args">Args</param>
         public static void Trim(List<CommandToken> args)
         {
-            for (int i = args.Count - 1; i >= 0; i--)
+            // Trim start
+
+            while (args.Count > 0 && args[0].Type == CommandTokenType.Space)
             {
-                if (args[i].Type == CommandTokenType.Space)
-                {
-                    args.RemoveAt(i);
-                }
+                args.RemoveAt(0);
+            }
+
+            // Trim end
+
+            while (args.Count > 0 && args[args.Count - 1].Type == CommandTokenType.Space)
+            {
+                args.RemoveAt(args.Count - 1);
             }
         }
 
--- a/Neo.ConsoleService/CommandToken.cs
+++ b/Neo.ConsoleService/CommandToken.cs
@@ -133,18 +133,12 @@ namespace Neo.ConsoleService
         /// <param name="args">Args</param>
         public static void Trim(List<CommandToken> args)
         {
-            // Trim start
-
-            while (args.Count > 0 && args[0].Type == CommandTokenType.Space)
+            for (int i = args.Count - 1; i >= 0; i--)
             {
-                args.RemoveAt(0);
-            }
-
-            // Trim end
-
-            while (args.Count > 0 && args[args.Count - 1].Type == CommandTokenType.Space)
-            {
-                args.RemoveAt(args.Count - 1);
+                if (args[i].Type == CommandTokenType.Space)
+                {
+                    args.RemoveAt(i);
+                }
             }
         }
 
--- a/neo-cli/CLI/MainService.Contracts.cs
+++ b/neo-cli/CLI/MainService.Contracts.cs
@@ -4,6 +4,7 @@ using Neo.Network.P2P.Payloads;
 using Neo.SmartContract.Native;
 using System;
 using System.Linq;
+using System.Numerics;
 
 namespace Neo.CLI
 {
@@ -34,7 +35,7 @@ namespace Neo.CLI
             UInt160 hash = SmartContract.Helper.GetContractHash(tx.Sender, nef.CheckSum, manifest.Name);
 
             Console.WriteLine($"Contract hash: {hash}");
-            Console.WriteLine($"Gas: {new BigDecimal(tx.SystemFee, NativeContract.GAS.Decimals)}");
+            Console.WriteLine($"Gas: {new BigDecimal((BigInteger)tx.SystemFee, NativeContract.GAS.Decimals)}");
             Console.WriteLine();
             SignAndSendTx(tx);
         }
--- a/neo-cli/CLI/MainService.Contracts.cs
+++ b/neo-cli/CLI/MainService.Contracts.cs
@@ -18,7 +18,7 @@ namespace Neo.CLI
         private void OnDeployCommand(string filePath, string manifestPath = null)
         {
             if (NoWallet()) return;
-            byte[] script = LoadDeploymentScript(filePath, manifestPath, out var nef);
+            byte[] script = LoadDeploymentScript(filePath, manifestPath, out var nef, out var manifest);
 
             Transaction tx;
             try
@@ -31,7 +31,7 @@ namespace Neo.CLI
                 return;
             }
 
-            UInt160 hash = SmartContract.Helper.GetContractHash(tx.Sender, nef.Script);
+            UInt160 hash = SmartContract.Helper.GetContractHash(tx.Sender, nef.CheckSum, manifest.Name);
 
             Console.WriteLine($"Contract hash: {hash}");
             Console.WriteLine($"Gas: {new BigDecimal(tx.SystemFee, NativeContract.GAS.Decimals)}");
--- a/neo-cli/CLI/MainService.Contracts.cs
+++ b/neo-cli/CLI/MainService.Contracts.cs
@@ -86,7 +86,7 @@ namespace Neo.CLI
             if (NoWallet()) return;
             try
             {
-                tx = CurrentWallet.MakeTransaction(tx.Script, sender, signers, gas: (long)value.Value);
+                tx = CurrentWallet.MakeTransaction(tx.Script, sender, signers, maxGas: (long)value.Value);
             }
             catch (InvalidOperationException e)
             {
--- a/neo-cli/CLI/MainService.Contracts.cs
+++ b/neo-cli/CLI/MainService.Contracts.cs
@@ -5,6 +5,7 @@ using Neo.SmartContract.Native;
 using System;
 using System.Linq;
 using System.Numerics;
+using Neo.Wallets;
 
 namespace Neo.CLI
 {
@@ -50,8 +51,19 @@ namespace Neo.CLI
         /// <param name="signerAccounts">Signer's accounts</param>
         /// <param name="gas">Max fee for running the script</param>
         [ConsoleCommand("invoke", Category = "Contract Commands")]
-        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null, long gas = TestModeGas)
+        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null, string gas = null)
         {
+            long gasAmount = TestModeGas;
+            if (!string.IsNullOrEmpty(gas))
+            {
+                AssetDescriptor descriptor = new AssetDescriptor(NativeContract.GAS.Hash);
+                if (!BigDecimal.TryParse(gas, descriptor.Decimals, out BigDecimal decimalAmount) || decimalAmount.Sign <= 0)
+                {
+                    Console.WriteLine("Incorrect Amount Format");
+                    return;
+                }
+                gasAmount = (long)decimalAmount.Value;
+            }
             Signer[] signers = Array.Empty<Signer>();
             if (signerAccounts != null && !NoWallet())
             {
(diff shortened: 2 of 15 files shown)


shargon (approved):

neo-cli/CLI/MainService.cs
```diff
@@ -514,7 +516,7 @@ private void SendTransaction(byte[] script, UInt160 account = null)
         /// <param name="verificable">Transaction</param>
         /// <param name="contractParameters">Contract parameters</param>
```
shargon, line 517: Add comments?

bettybao1209: Updated.

shargon: Please add for `showStack` too

bettybao1209: Done.

erikzhang (approved):

superboyiii (changes requested):
https://github.com/bettybao1209/neo-node/blob/a85b767e76a62a0343c1b276743f1b4199bd1041/neo-cli/CLI/MainService.cs#L488
Please change here like: `using (ApplicationEngine engine = ApplicationEngine.Run(tx.Script, container: tx, gas:1010_00000000))`, it was called by `register candidates` method straightly. This is what I need.

shargon (dismissed):

bettybao1209:
> Please change here like: `using (ApplicationEngine engine = ApplicationEngine.Run(tx.Script, container: tx, gas:1010_00000000))`, it was called by `register candidates` method straightly. This is what I need.

There are two ways to register a candidate, so I updated both the `invoke` and `register candidate` commands. Also, I have to add the `TestModeGas` in `MakeTransaction` method since it also runs the script. Please review again.

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -50,8 +51,19 @@ private void OnDeployCommand(string filePath, string manifestPath = null)
         /// <param name="signerAccounts">Signer's accounts</param>
         /// <param name="gas">Max fee for running the script</param>
         [ConsoleCommand("invoke", Category = "Contract Commands")]
-        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null, long gas = TestModeGas)
+        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null, string gas = null)
```
shargon, line 54: Use `BigDecimal`?

bettybao1209: I parsed the string to BigDecimal in the next few lines.

shargon: I mean as argument, instead of string BigDecimal, and allow them as a new type

neo-cli/CLI/MainService.cs
```diff
@@ -463,8 +465,10 @@ private static void WriteLineWithoutFlicker(string message = "", int maxWidth =
         /// </summary>
         /// <param name="script">script</param>
         /// <param name="account">sender</param>
-        private void SendTransaction(byte[] script, UInt160 account = null)
+        /// <param name="gas">Max fee for running the script</param>
+        private void SendTransaction(byte[] script, UInt160 account = null, long gas = TestModeGas)
```
superboyiii, line 469: I think no need to add extra TestModeGas as an optional param here. It's not invokefunction. This made it more complex. The system fee of registering candidates is a a certain value -> about 1000 gas.
https://github.com/neo-project/neo/blob/677bd9648d40b31076ba78c88c8a48bc0e919f66/src/neo/SmartContract/Native/NeoToken.cs#L212
So 1000 gas + 20 test gas is enough. If you do it optional here, users will not know how much to input. We should not throw the problem to user.

bettybao1209: SendTransaction is also called by other commands，of which the test mode gas are different. Also，this method should be apparent to users.

superboyiii: > SendTransaction is also called by other commands，of which the test mode gas are different. Also，this method should be transparent to users.

Only vote related methods are using `SendTransaction`, 1020 gas will not cause any malicious attack since it's a user behaviour . But if we make it optional here, it will not be transparent but just confusion.

bettybao1209:
1. Insufficient gas (two signers + default gas):
 `invoke 0xf61eebf573ea36593fd43aa150c055ad7906ab83 registerCandidate [{"type":"PublicKey","value":"03d922481a6190c0afdc200600abec944d5b588a6fe6ef9639cc28be3493319b0d"}] NaqFR6Qv3VM3XWLLqbuK6PwRi9uYGGDk1e NMbT8LVG9aLdKs3HBzf5LvV2XBqjjjjpeR,NUjGMLwcYZuhUSTeNcZSvYkrQRciAhnFET`
 
 2. succeed (+ 1010 gas) :
` invoke 0xf61eebf573ea36593fd43aa150c055ad7906ab83 registerCandidate [{"type":"PublicKey","value":"03d922481a6190c0afdc200600abec944d5b588a6fe6ef9639cc28be3493319b0d"}] NaqFR6Qv3VM3XWLLqbuK6PwRi9uYGGDk1e NMbT8LVG9aLdKs3HBzf5LvV2XBqjjjjpeR 1010`
 
 3. succeed (+ null) ：
 `export key null 0x75b75932a1451cc0c56a95eff7fcc01de45aa5a3`

bettybao1209:
wait for https://github.com/neo-project/neo-node/pull/733

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -48,9 +50,11 @@ private void OnDeployCommand(string filePath, string manifestPath = null)
         /// <param name="contractParameters">Contract parameters</param>
         /// <param name="sender">Transaction's sender</param>
         /// <param name="signerAccounts">Signer's accounts</param>
+        /// <param name="gas">Max fee for running the script</param>
         [ConsoleCommand("invoke", Category = "Contract Commands")]
-        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null)
+        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null, decimal maxGas = 20)
         {
+            var value = BigDecimal.Parse(maxGas.ToString(CultureInfo.InvariantCulture), NativeContract.GAS.Decimals);
```
Tommo-L, line 57:
```suggestion
            var fee = BigDecimal.Parse(maxGas.ToString(CultureInfo.InvariantCulture), NativeContract.GAS.Decimals);
```

neo-cli/CLI/MainService.cs
```diff
@@ -465,8 +467,10 @@ private static void WriteLineWithoutFlicker(string message = "", int maxWidth =
...
         {
+
```
Tommo-L, line 473:
```suggestion
```

neo-cli/CLI/MainService.Contracts.cs
bettybao1209: Done.

Tommo-L (dismissed):

Neo.ConsoleService/CommandToken.cs
```diff
@@ -133,18 +133,12 @@ public static string ToString(IEnumerable<CommandToken> tokens)
...
-
-            while (args.Count > 0 && args[0].Type == CommandTokenType.Space)
-            {
-                args.RemoveAt(0);
-            }
-
-            // Trim end
-
-            while (args.Count > 0 && args[args.Count - 1].Type == CommandTokenType.Space)
+            for (int i = args.Count - 1; i >= 0; i--)
```
shargon, line 136: This will move the arguments and it could be a little random

neo-cli/CLI/MainService.Vote.cs
```diff
@@ -5,18 +5,23 @@
 using Neo.VM.Types;
 using Neo.Wallets;
 using System;
+using System.Globalization;
 
 namespace Neo.CLI
 {
     partial class MainService
     {
+        public const uint RegisterGas = 1010;
```
shargon, line 14:
```suggestion
        public const uint RegisterGas = 1010;

```

Neo.ConsoleService/CommandToken.cs
bettybao1209: Here I remove all space items in the args list at one time so we don't have to trim the string for each parameter. But the side effect is that we should use the `comma` to separate the string array as defined in https://github.com/neo-project/neo-node/blob/master/Neo.ConsoleService/ConsoleServiceBase.cs#L414.

bettybao1209: Maybe we should focus on adding the gas feature in this pr, so I revert the change.

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -48,9 +50,11 @@ private void OnDeployCommand(string filePath, string manifestPath = null)
         /// <param name="contractParameters">Contract parameters</param>
         /// <param name="sender">Transaction's sender</param>
         /// <param name="signerAccounts">Signer's accounts</param>
+        /// <param name="gas">Max fee for running the script</param>
         [ConsoleCommand("invoke", Category = "Contract Commands")]
-        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null)
+        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null, decimal maxGas = 20)
         {
+            var gas = BigDecimal.Parse(maxGas.ToString(CultureInfo.InvariantCulture), NativeContract.GAS.Decimals);
```
erikzhang, line 57: We shouldn't use `BigDecimal.Parse()` here. Use the constructor directly.

bettybao1209: Done.

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -48,9 +48,11 @@ private void OnDeployCommand(string filePath, string manifestPath = null)
         /// <param name="contractParameters">Contract parameters</param>
         /// <param name="sender">Transaction's sender</param>
         /// <param name="signerAccounts">Signer's accounts</param>
+        /// <param name="gas">Max fee for running the script</param>
         [ConsoleCommand("invoke", Category = "Contract Commands")]
-        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null)
+        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null, decimal maxGas = 20)
```
erikzhang, line 53: I think maybe we should use `long` uniformly. I see that we currently use `long` in some places and `decimal` in other places. This is easy to confuse.

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -48,9 +48,11 @@ private void OnDeployCommand(string filePath, string manifestPath = null)
         /// <param name="contractParameters">Contract parameters</param>
         /// <param name="sender">Transaction's sender</param>
         /// <param name="signerAccounts">Signer's accounts</param>
+        /// <param name="gas">Max fee for running the script</param>
         [ConsoleCommand("invoke", Category = "Contract Commands")]
-        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null)
+        private void OnInvokeCommand(UInt160 scriptHash, string operation, JArray contractParameters = null, UInt160 sender = null, UInt160[] signerAccounts = null, long maxGas = 20)
         {
+            var gas = new BigDecimal((decimal)maxGas, NativeContract.GAS.Decimals);
```
erikzhang, line 55: This variable is no more needed.

bettybao1209: You mean we should take the gas input like `2000000000`  as `20`?

erikzhang: Yes.

erikzhang: Maybe we use `decimal` for user input?

shargon: I think that it's easier to use decimal for users

erikzhang: So, for the user interface, we use `decimal`. For the program interface, we use long.

erikzhang (approved):

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -48,9 +48,11 @@ private void OnDeployCommand(string filePath, string manifestPath = null)
         /// <param name="contractParameters">Contract parameters</param>
         /// <param name="sender">Transaction's sender</param>
         /// <param name="signerAccounts">Signer's accounts</param>
+        /// <param name="gas">Max fee for running the script</param>
```
shargon, line 51:
```suggestion
        /// <param name="masGas">Max fee for running the script</param>
```

shargon (approved):

erikzhang (approved):

superboyiii (approved):

PR #737, opened 2021-02-03
Show header height when show state

ProDog:
Show header height when show state


ProDog:
Wait https://github.com/neo-project/neo/pull/2259.

erikzhang (approved):

PR #741, opened 2021-02-08
Sync with neo (remove singletons #2302)

chenzhitong:


chenzhitong:
Can't run after the change

Issue #762, opened 2021-03-16
show state immediately after neo-node start error


cloud8little:
![693f6484b79cc5ba0e9412e90545b98](https://user-images.githubusercontent.com/34291844/111279077-4319ec00-8675-11eb-885c-f6c0c10a59a3.png)

![a806ba04ee24181d0d74fe57bd80fbf](https://user-images.githubusercontent.com/34291844/111279259-7a889880-8675-11eb-8873-feb873b71bc6.png)

It is caused by the change https://github.com/neo-project/neo-node/pull/758, localnode is null at this time.

after several seconds later, it shows numbers.

shargon:
Revert #758?

ZhangTao1596:
> Revert #758?

If you have better solution to solve https://github.com/neo-project/neo/issues/2380.
I found that if you start node with websockt port occupied by another application, you will get different localnodes before and after call `NeoSystem.StartNode`.

PR #763, opened 2021-03-16
fix localnode null

bettybao1209:
close https://github.com/neo-project/neo-node/issues/762

### DIFF
--- a/neo-cli/CLI/MainService.Node.cs
+++ b/neo-cli/CLI/MainService.Node.cs
@@ -71,7 +71,6 @@ namespace Neo.CLI
                     uint headerHeight = NeoSystem.HeaderCache.Last?.Index ?? height;
 
                     Console.SetCursorPosition(0, 0);
-                    LocalNode ??= await NeoSystem.LocalNode.Ask<LocalNode>(new LocalNode.GetInstance());
                     WriteLineWithoutFlicker($"block: {height}/{headerHeight}  connected: {LocalNode.ConnectedCount}  unconnected: {LocalNode.UnconnectedCount}", Console.WindowWidth - 1);
 
                     int linesWritten = 1;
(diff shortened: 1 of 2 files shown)


neo-cli/CLI/MainService.Node.cs
```diff
@@ -71,6 +71,7 @@ private void OnShowStateCommand()
                     uint headerHeight = NeoSystem.HeaderCache.Last?.Index ?? height;
 
                     Console.SetCursorPosition(0, 0);
+                    LocalNode ??= await NeoSystem.LocalNode.Ask<LocalNode>(new LocalNode.GetInstance());
```
shargon, line 74: How it could be null if it's asigned during `Start`?

ZhangTao1596: Because `Start` is `async`.

cloud8little:
Test passed. When immediately input show state after starting neo, it shows height after 1~2 seconds.

cloud8little (approved):

erikzhang:
Please wait for my real fix.

ZhangTao1596 (approved):

neo-cli/CLI/MainService.Node.cs
ZhangTao1596: We should remove `LocalNode` ask in `Start`.

shargon (approved):

ZhangTao1596 (commented):
It is a revert of #758. 
It can't fix https://github.com/neo-project/neo/issues/2380.
We can't trust global `LocalNode` as I said https://github.com/neo-project/neo-node/pull/763#discussion_r595070102.

superboyiii:
Yes, if make it like this, we will remeet the issue of https://github.com/neo-project/neo/issues/2380.
image

erikzhang:
Running two different nodes on a single server with the same port is not a normal usage. And the second node should throw exception and be terminated. I think the fix of neo-project/neo#2380 should be in the core.

PR #776, opened 2021-05-08
add unvote function

cloud8little:


erikzhang (approved):

erikzhang:
Test?

shargon:
Is not better to allow null on `publicKey` and use `vote`?

cloud8little:
> Is not better to allow null on `publicKey` and use `vote`?

If you mean set the publicKey as null by default when user did not input the publicKey, it is feasible, but I think it is not straightforward to express the meaning of unvote. User may forget to input the publickey field and actually it will cancel the user's vote, it is very confused.

nicolegys (approved):
tested

shargon (approved):

PR #787, opened 2021-06-08
Fix incomplete signature make invoke failed.

superboyiii:


neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -615,7 +615,11 @@ private void SignAndSendTx(DataCache snapshot, Transaction tx)
                 Console.WriteLine($"Signed and relayed transaction with hash={tx.Hash}");
                 return;
```
erikzhang, line 616: This line can be removed.

superboyiii: Done

erikzhang (approved):

shargon (approved):

PR #796, opened 2021-07-15
Fix logger path

erikzhang:


shargon:
But then testnet and mainnet will collide using the same binary

erikzhang:
`_{0}` is not used in the code.

shargon (approved):

PR #807, opened 2021-08-06
Install plugin and dependencies.

Liaojinghui:
Automatically install plugin depandencies.


neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -17,6 +18,30 @@ partial class MainService
...
+        {
+            CheckAndInstall(pluginName);
+        }
+
+        private void CheckAndInstall(string pluginName)
+        {
+            if (pluginName == "ApplicationLogs" ||
+            pluginName == "RpcNep17Tracker" ||
+            pluginName == "StateService" ||
+            pluginName == "OracleService"
```
erikzhang, line 30: I think hard coding is not a good idea.

Liaojinghui: Then we have to add it to the config

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -49,18 +85,86 @@ private void OnInstallCommand(string pluginName)
...
-                try
+
+                var output = new MemoryStream();
+                while ((read = stream.Read(buffer, 0, buffer.Length)) > 0)
                 {
-                    zip.ExtractToDirectory(".");
-                    Console.WriteLine($"Install successful, please restart neo-cli.");
+                    output.Write(buffer, 0, read);
+                    totalRead += read;
+                    Console.Write($"\rDownloading {pluginName}.zip {totalRead / 1024}KB/{response.ContentLength / 1024}KB {(totalRead * 100) / response.ContentLength}%");
```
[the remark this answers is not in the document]
Liaojinghui: As I have added the sha256 of the file to the console, I don't think that link is necessary.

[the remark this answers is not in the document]
cloud8little: it is necessary, if the user account into network issue,  and the download will be stuck and paused, they will never know where to download the plugin file. I think the old behavior is more user-friendly.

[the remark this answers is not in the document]
Liaojinghui: added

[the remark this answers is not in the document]
Liaojinghui: image

Liaojinghui:
@erikzhang

shargon (commented):
tested?

Liaojinghui:
> tested?

@shargon it works well on my machine.

shargon (dismissed):

cloud8little:
Tested with two comments:

1. when installing OracleService, Grpc.xx dependency files will be added in the zip file, but when uninstall, they will not be removed. when install again, it will show plugin exist, users can use "reinstall OraceService" to solve the problem, while I doubt they will use the reinstall command. IMO, in this case, can we remind the user to use reinstall instead?

d7f0cbb0280d384d249c1baa13c38ff

2. This is not caused by this PR, but I think we should also solve this. Currently install <pluginname> is case-sensitive.  Can we make it case-insensitive? 

6e736fbfb8b91632130e02f8b64fbdc

Liaojinghui:
> Tested with two comments:
> 
> 1. when installing OracleService, Grpc.xx dependency files will be added in the zip file, but when uninstall, they will not be removed. when install again, it will show plugin exist, users can use "reinstall OraceService" to solve the problem, while I 


@cloud8little unfortunately, we can not uninstall the dependency along with the plugin, because we don't know whether that dependency is also relyed by other plugins.

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -56,21 +95,109 @@ private void OnInstallCommand(string pluginName)
...
+            {
+                zip.ExtractToDirectory(".", overWrite);
+                InstallDependency(pluginName, pluginToInstall);
+                Console.WriteLine($"Install successful, please restart neo-cli.");
+                pluginToInstall.Pop();
+            }
+            catch (IOException)
+            {
+                pluginToInstall.Pop();
+                Console.WriteLine($"Plugin already exist.");
```
cloud8little, line 156:
```suggestion
                Console.WriteLine($"Plugin already exist. Try reinstall command");
```

cloud8little:
> > Tested with two comments:
> > 
> > 1. when installing OracleService, Grpc.xx dependency files will be added in the zip file, but when uninstall, they will not be removed. when install again, it will show plugin exist, users can use "reinstall OraceService" to solve the problem, while I
> 
> @cloud8little unfortunately, we can not uninstall the dependency along with the plugin, because we don't know whether that dependency is also relyed by other plugins.

how about to remind the user to use reinstall command?

cloud8little:
@Liaojinghui please help solve the conflict.

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -92,8 +219,10 @@ private void OnUnInstallCommand(string pluginName)
                 return;
             }
 
-            File.Delete(plugin.Path);
-            File.Delete(plugin.ConfigFile);
+            DeleteFiles(plugin.Path,
+                plugin.ConfigFile,
+                $"{Plugin.PluginsDirectory}/{pluginName}/DEPENDENCY");
```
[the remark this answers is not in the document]
erikzhang: I think we don't have this file now.

shargon:
tested?

Liaojinghui:
@shargon

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -27,12 +30,48 @@ partial class MainService
...
+        /// <summary>
+        /// Force to install a plugin again. This will overwrite
+        /// existing plugin files, in case of any file missing or
+        /// damage to the old version.
+        /// </summary>
+        /// <param name="pluginName">name of the plugin</param>
+        [ConsoleCommand("reinstall", Category = "Plugin Commands", Description = "Overwrite existing plugin by force.")]
+        private void OnReinstallCommand(string pluginName)
+        {
+            Stack<string> pluginToInstall = new();
```
shargon, line 54: if we don't clean the old files, maybe reinstall it's the same

Liaojinghui: I don't think `reinstall` could do the same as `uninstall`, what if users just wanna remove the plugin.

Liaojinghui:
@shargon @erikzhang  This pr works together with https://github.com/neo-project/neo-modules/pull/624

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -30,8 +33,43 @@ partial class MainService
...
+        /// Download plugin from github release
+        /// The function of download and install are divided
+        /// for the consideration of `update` command that
+        /// might be added in the future.
+        /// </summary>
+        /// <param name="pluginName">name of the plugin</param>
+        /// <returns>Downloaded content</returns>
+        private async Task<MemoryStream> DownloadPlugin(string pluginName)
+        {
+            var url = $"https: //github.com/neo-project/neo-modules/releases/download/v{typeof(Plugin).Assembly.GetVersion()}/{pluginName}.zip";
```
cloud8little, line 70:
```suggestion
            var url = $"https://github.com/neo-project/neo-modules/releases/download/v{typeof(Plugin).Assembly.GetVersion()}/{pluginName}.zip";
```

cloud8little:
https://github.com/neo-project/neo-node/pull/807#discussion_r845920181 with the correct change, all tests passed.

superboyiii (approved):

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -54,20 +92,107 @@ private async Task OnInstallCommandAsync(string pluginName)
...
+            using (SHA256 sha256 = SHA256.Create())
+            {
+                ConsoleHelper.Info("SHA256: ", $"{sha256.ComputeHash(stream.ToArray()).ToHexString()}");
+            }
+            using ZipArchive zip = new(stream, ZipArchiveMode.Read);
+
+            try
+            {
+                zip.ExtractToDirectory(".", overWrite);
+                await InstallDependency(pluginName, pluginToInstall);
```
erikzhang, line 145: Maybe we should install the dependencies first?

Liaojinghui: How could we know the dependency without installing the plugin...

erikzhang: Download it first, then read the config.json file in memory.

Liaojinghui: How about install it first, then delete it if any error happens when install its dependency?

erikzhang: If you installed it first, but the process crashed abnormally, you have no chance to delete the files.

Liaojinghui: updated

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -54,20 +92,114 @@ private async Task OnInstallCommandAsync(string pluginName)
...
+
+            using (SHA256 sha256 = SHA256.Create())
+            {
+                ConsoleHelper.Info("SHA256: ", $"{sha256.ComputeHash(stream.ToArray()).ToHexString()}");
+            }
+            using ZipArchive zip = new(stream, ZipArchiveMode.Read);
+
+            try
+            {
+                foreach (var entry in zip.Entries.Where(p => p.Name == "config.json"))
```
erikzhang, line 144: If there are more than one `config.json` in the zip file, the plugin will be installed multiply times.

Liaojinghui: I know~ but will you add more than one `config.json` to the plugin?

cloud8little (approved):

erikzhang (approved):

erikzhang:
Need to test again.

cloud8little:
18c41be0c7c79b01086d821b815217f

reinstall StateService cause error:

3bd5561d40cdb4b028823c811e9725b

shargon (approved):

superboyiii:
@shargon @erikzhang Please have a look.

erikzhang (approved):

cloud8little:
install OracleService and then uninstall, show successful, then install OracleService, it shows error because the Grpc dll already exists, when uninstall, they are not deleted. Then it requires end-user to delete these files by themselves, then install can be successful. 

608588bd406b736a21507958173f438

erikzhang (approved):

PR #809, opened 2021-08-10
Add `restart` command for `neo-cli`

Liaojinghui:

Issue #808, opened 2021-08-08
neo-cli update system


Liaojinghui:
Currently, we need to manually download the new version `neo-cli` from GitHub to update the neo node, which is inconvenient and inefficient, especially when emergency fixes are released. In the meanwhile, Github only has full version binary available, this is redundant for an update while we only need those files that are changed. 

**Solution**
1. Create a patch release that only contains files that need to be overwritten or added every time a new version is released. That patch should contain a patch instruction file indicating how we should make that patch.
2. Add a system update command to the neo-cli, enabling users to patch the system manually from the neo console.
3. Provides an `automatically update` option, which is set to close by default, but that node will automatically update when a new version is available once opened by the node owner. (or even forced to stop consensus when a critical patch is released)
4. (Optional hotpatch) Create a wrapper program to  `neo-cli` to automatically reboot `neo-cli` when an update or a plugin installation happens.

Liaojinghui:
Full binary package: 42MB
Full binary folder after unzip: 95MB

**For ordinary update**:
Patch binary package: 300kb
Patch binary folder after unzip: 1MB

PR #810, opened 2021-08-11
`Update` command

Liaojinghui:
This command helps the node to update to the latest version fast and easily.
this will be very useful for some emergency patches.

This PR implements issue #808
Rely on #807 
Should work collaboratively with #809 and https://github.com/neo-project/neo/pull/2569 to do a fast node update.


Liaojinghui:
Need https://github.com/neo-project/neo/pull/2569

PR #811, opened 2021-08-13
User friendly `show state` command

Liaojinghui:
image


shargon (approved):

erikzhang:
What's the difference?

Liaojinghui:
> What's the difference?

@erikzhang  I created this pr first, this one solved two major issues:
- adding a block update progress, indicating the process of consensus (the dot in the second line). Otherwise, the screen looks frozen.
- give the tag colors making the state more clear to check. (This can be replaced by #812 )


After I committed this pr, I found that the whole cli outputs are not user-friendly. So I created #812 which focuses on the color of different level logs.

Issue #814, opened 2021-08-17
No need to create an address as default when creating a new wallet


superboyiii:
Someone just wants to create a wallet and import an existed privkey but has to delete the first address because it 's never used.

Liaojinghui:
why not allowing them to create a wallet by importing a private key?

Liaojinghui:
empty wallet just looks weird,,,,, though it is normal in real life.....

Qiao-Jin:
Moreover it think it should be enabled to delete address from wallet

roman-khimov:
> empty wallet just looks weird,,,,, though it is normal in real life.....

NeoGo allows to create empty wallet, that's what `wallet init` command does by default. Only with `-a` it actually also creates a new account.

> Moreover it think it should be enabled to delete address from wallet

Sure. `wallet remove` in NeoGo allows to do that.

PR #815, opened 2021-08-17
Enable private key import in initial wallet creation

superboyiii:
Close https://github.com/neo-project/neo-node/issues/814
~Require https://github.com/neo-project/neo/pull/2580~


erikzhang:
If there is no address in the wallet, the password is removed too.

superboyiii:
> If there is no address in the wallet, the password is removed too.

No, wallet need password, it's a container which is for later import.

erikzhang:
The password is not stored in the wallet, but is used to encrypt the private keys. If you remove all the addresses, then the password is removed from the wallet too.

shargon:
@superboyiii why you need an empty wallet?

superboyiii:
> @superboyiii why you need an empty wallet?

Someone imports Neo Legacy's privatekey to N3 wallet but have to create an unused address as the default address, it's strange.
Mostly, importing private key doesn't need any default address in wallet.

superboyiii:
@erikzhang @shargon Could you review again? I make a double pwd check when open an empty wallet in case someone imports privatekey but forget the previous password.

shargon:
@superboyiii maybe it's better to create a new command for import for neo legacy?

superboyiii:
@erikzhang @shargon Pls review again.

neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -187,11 +187,23 @@ private void OnExportKeyCommand(string path = null, UInt160 scriptHash = null)
...
-        private void OnCreateWalletCommand(string path)
+        private void OnCreateWalletCommand(string path, string wifOrFile = null)
         {
+            string password = ReadUserInput("password", true);
+            if (password.Length == 0)
+            {
+                Console.WriteLine("Cancelled");
+                return;
+            }
+            string password2 = ReadUserInput("password", true);
```
shargon:
```suggestion
            string password2 = ReadUserInput("repeat password", true);
```

neo-cli/CLI/MainService.cs
```diff
@@ -385,17 +388,6 @@ public void OpenWallet(string path, string password)
                 case ".json":
                     {
                         NEP6Wallet nep6wallet = new NEP6Wallet(path, NeoSystem.Settings);
-                        if (!nep6wallet.GetAccounts().Any())
```
shargon: We don't need to remove this code now, isn't it?

superboyiii: > We don't need to remove this code now, isn't it?

Then I still need `nep6wallet.UnlockEmpty(password)`

neo-cli/CLI/MainService.cs
```diff
@@ -230,8 +246,8 @@ private IEnumerable<Block> GetBlocksFromFile()
 
         private bool NoWallet()
         {
-            if (CurrentWallet != null) return false;
-            Console.WriteLine("You have to open the wallet first.");
+            if (CurrentWallet != null && CurrentWallet.GetAccounts().Any()) return false;
+            Console.WriteLine("No wallet or no address exists.");
```
erikzhang, line 250: We don't need these changes.

neo-cli/CLI/MainService.cs
superboyiii: PS: Unlock() is OK, I'll use it.

neo-cli/CLI/MainService.cs
superboyiii: Without this, if open an empty wallet we can still go ahead. This will waste time in most commands which signer needed like `vote`, `deploy`, `send`.

erikzhang: How to create an empty wallet with neo-cli?

superboyiii: > How to create an empty wallet with neo-cli?

When you delete an address which is the only address by https://github.com/neo-project/neo-node/pull/817

superboyiii: > How to create an empty wallet with neo-cli?

Otherwise, we should enable empty wallet to import privatekey.

superboyiii:
@shargon @erikzhang Ready for review.

neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -301,10 +302,13 @@ private void OnImportKeyCommand(string wifOrFile)
         [ConsoleCommand("import watchonly", Category = "Wallet Commands")]
         private void OnImportWatchOnlyCommand(string addressOrFile)
         {
+            if (CurrentWallet == null) return;
```
shargon, line 305: If we don't use NoWallet, the user won't receive a error message

neo-cli/CLI/MainService.cs
erikzhang: I think we don't need to think about empty wallet. It's not a normal usage.

neo-cli/CLI/MainService.Wallet.cs
superboyiii: > If we don't use NoWallet, the user won't receive a error message

fixed.

neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -111,7 +111,11 @@ private void OnUpgradeWalletCommand(string path)
         [ConsoleCommand("create address", Category = "Wallet Commands")]
         private void OnCreateAddressCommand(ushort count = 1)
         {
-            if (NoWallet()) return;
+            if (CurrentWallet == null)
```
shargon, line 114: Why not use `NoWallet` if it's almost the same?

superboyiii: > Why not use `NoWallet` if it's almost the same?

Empty wallet can also create address.

superboyiii: NoWallet() is reverted.

neo-cli/CLI/MainService.cs
superboyiii: > I think we don't need to think about empty wallet. It's not a normal usage.

NoWallet() is reverted.

neo-cli/CLI/MainService.Wallet.cs
```diff
@@ -301,10 +296,17 @@ private void OnImportKeyCommand(string wifOrFile)
...
             UInt160 address = null;
             try
             {
                 address = StringToAddress(addressOrFile, NeoSystem.Settings.AddressVersion);
+                WalletAccount currentAccount = CurrentWallet.GetAccount(address);
+                if (currentAccount != null)
+                {
+                    Console.WriteLine("This address is already in your wallet");
+                    return;
+                }
```
erikzhang, line 309: Move these lines out of the `try` block.

neo-cli/CLI/MainService.cs
```diff
@@ -372,6 +366,16 @@ public void OpenWallet(string path, string password)
...
+                        if (!nep6wallet.GetAccounts().Any())
+                        {
+                            Console.WriteLine("No address is found, please reconfirm your password.");
+                            string password2 = ReadUserInput("repeat password", true);
+                            if (password != password2)
+                            {
+                                Console.WriteLine("Two passwords are inconsistent.");
+                                return;
+                            }
+                        }
```
erikzhang, line 378: These lines are not necessary.

superboyiii: > These lines are not necessary.

Done.

erikzhang (approved):

PR #828, opened 2021-10-09
upgrade neo-cli and all the local plugins to the latest version

Liaojinghui:
This pr will automatically update both the `neo-cli` and all installed `plugins` to the most latest version.


Liaojinghui:
image

cloud8little:
why the plugin is still in version 3.0.1?

a32068d04ad430061032b1b88c25de2

b7cf20765e484d7fd6f62545f5db15a

Liaojinghui:
image
Fixed now, it is because I forgot to copy the .dll file as it is outside of the `PluginName` folder

cloud8little:
The plugin program file xx.dll will be replaced now. But before upgrade, I have oracleservice.dll in the Plugins file, while without OracleService folder(which contain config.json), and after upgrade, other dependencies such as grpc filestorage.dll is added, and OracleService folder new generated, but the folder is still empty without config.json.

c9550887ebb85d81495a4c9a42f461f

cloud8little:
> `g.json` too.

I think if there is no config.json, then you can unzip it from the upgrade file, if config.json already exist, no need to override it.

Liaojinghui:
> I think if there is no config.json, then you can unzip it from the upgrade file, if config.json already exist, no need to override it.

that is exactly the current logic.

cloud8little:
Regardless of win/linux, why the upgrader program must need a runtime folder to work?

fcb37e7e632cbf1b9ea2290295df9bf

Liaojinghui:
> win/linux, why the upgrader program must need a runtime folder to work?

It rely on Neo, I guess it is a dependency of Neo

superboyiii (dismissed):

superboyiii:
@shargon @erikzhang  Merge?

erikzhang:
Tested?

superboyiii:
> Tested?

Yes

update/Program.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                response = DownloadFromAPI(version, "neo-modules", pluginName);
+            }
+            using (response)
+            {
+                using Stream stream = response.GetResponseStream();
+                using ZipArchive zip = new(stream, ZipArchiveMode.Read);
+                try
+                {
+                    var temp = Path.Combine(Path.GetTempPath());
```
shargon, line 131: Combine with what?
```suggestion
                    var temp = Path.GetTempPath();
```

Liaojinghui:
@shargon please review

Liaojinghui:
image

Liaojinghui:
@shargon can you please review?

update/Update.cs
```diff
@@ -0,0 +1,195 @@
...
+            try
+            {
+                var targetFile = newPath.Replace(sourcePath, targetPath);
+                // Avoid config.json and other .json files
+                if (Path.GetExtension(newPath) != ".json" || !File.Exists(targetFile))
+                    File.Copy(newPath, targetFile, true);
+            }
+            catch
+            {
+                // ignored
```
shargon, line 191: I think that we need to log it

update/Update.cs
```diff
@@ -0,0 +1,195 @@
...
+            }
```
shargon, line 192:
```suggestion
            catch (Exception ex)
            {
                ConsoleHelper.Error(ex.ToString());
            }
```

shargon: I think that we need to log it

shargon (approved):

erikzhang:
I think the upgrade will be a little difficult. For example, automatic updating of configuration files (while keeping the original settings). Consider whether to resync blocks.

Liaojinghui:
> I think the upgrade will be a little difficult. For example, automatic updating of configuration files (while keeping the original settings). Consider whether to resync blocks.

This just makes the process of `downloading new version from github then unzip it and copy it to target folder` into a single command.....as to config and resync, user can take care them manually (they are doing these manually now, nothing has changed).

erikzhang:
Then I don't think this upgrader is useful. Because if an upgrader is used, the user will have a false sense of "it'll do everything for me", which often leads to problems.

PR #836, opened 2021-11-04
add libleveldb.dll

Qiao-Jin:


shargon:
I think that the plugin should provide it

Qiao-Jin:
> I think that the plugin should provide it

But plugins are in Plugin folder

erikzhang:
Agree with @shargon.

PR #847, opened 2022-01-20
typo, comment

Liaojinghui:


neo-cli/CLI/ConsolePercent.cs
```diff
@@ -46,10 +47,10 @@ public long Value
         /// <summary>
         /// Maximum value
         /// </summary>
-        public long MaxValue
+        private long MaxValue
```
shargon, line 50: Please don't change visiblities

Neo.ConsoleService/ConsoleColorSet.cs
```diff
@@ -14,8 +14,8 @@ namespace Neo.ConsoleService
 {
     public class ConsoleColorSet
     {
-        public ConsoleColor Foreground;
-        public ConsoleColor Background;
+        private readonly ConsoleColor _foreground;
```
shargon, line 17: Please don't change visiblities

shargon (changes requested):

Neo.ConsoleService/ConsoleServiceBase.cs
```diff
@@ -540,8 +529,8 @@ public void Run(string[] args)
                         RedirectStandardOutput = true,
                         UseShellExecute = false
                     });
-                    process.WaitForExit();
-                    Console.Write(process.StandardOutput.ReadToEnd());
+                    process?.WaitForExit();
```
erikzhang, line 532: I think we should throw the exception here instead of ignoring it. The easiest way is to remove the question mark.

neo-cli/CLI/Logger.cs
```diff
@@ -27,26 +27,31 @@ internal class Logger : Plugin, ILogPlugin
 
         public override string Name => "SystemLog";
         public override string Description => "Prints consensus log and is a built-in plugin which cannot be uninstalled";
-        public override string ConfigFile => Combine(GetDirectoryName(Path), "config.json");
+        public override string ConfigFile => Combine(GetDirectoryName(Path) ?? throw new InvalidOperationException(), "config.json");
```
erikzhang, line 30: Why do we need this?

neo-cli/CLI/Logger.cs
```diff
@@ -27,26 +27,31 @@ internal class Logger : Plugin, ILogPlugin
...
         public override string Path => GetType().Assembly.Location;
 
         private static void GetErrorLogs(StringBuilder sb, Exception ex)
         {
-            sb.AppendLine(ex.GetType().ToString());
-            sb.AppendLine(ex.Message);
-            sb.AppendLine(ex.StackTrace);
-            if (ex is AggregateException ex2)
+            while (true)
```
erikzhang, line 35: What's the difference?

neo-cli/CLI/Logger.cs
Liaojinghui: Cause the return value of function `GetDirectoryName` is nullable:

`public static string? GetDirectoryName(string? path)`

neo-cli/CLI/Logger.cs
Liaojinghui: Compared to loop, recursion is an inefficient algorithm, that is why we should avoid using recursion as much as possible.

Liaojinghui: almost removed all changes that has nothing to do with the typo and comments, realized that this pr contains too much that is not important at all.

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -26,6 +26,7 @@ partial class MainService
         /// </summary>
         /// <param name="filePath">File path</param>
         /// <param name="manifestPath">Manifest path</param>
+        /// <param name="data"></param>
```
shargon, line 29:
```suggestion
        /// <param name="data">Extra data for deploy</param>
```

neo-cli/CLI/MainService.Contracts.cs
```diff
@@ -57,8 +58,12 @@ private void OnDeployCommand(string filePath, string manifestPath = null, JObjec
         /// <summary>
         /// Process "update" command
         /// </summary>
+        /// <param name="scriptHash">Script hash</param>
         /// <param name="filePath">File path</param>
         /// <param name="manifestPath">Manifest path</param>
+        /// <param name="sender">Sender</param>
+        /// <param name="signerAccounts">Signer Accounts</param>
+        /// <param name="data"></param>
```
shargon, line 66:
```suggestion
        /// <param name="data">Extra data for update</param>
```

neo-cli/CLI/MainService.Node.cs
```diff
@@ -107,10 +107,15 @@ private void OnShowStateCommand()
...
                 }
-            });
+            }, cancel.Token);
             ReadLine();
             cancel.Cancel();
-            try { Task.WaitAll(task, broadcast); } catch { }
+            try { Task.WaitAll(task, broadcast); }
+            catch
+            {
+                // ignored
```
shargon, line 116: I think that this comment is not needed

Neo.ConsoleService/ConsoleServiceBase.cs
```diff
@@ -598,7 +586,10 @@ public virtual void RunConsole()
                 {
                     Console.Title = ServiceName;
                 }
-                catch { }
+                catch
+                {
+                    // ignored
+                }
```
Liaojinghui, line 592:
```suggestion
                catch { }
```

shargon (approved):

neo-cli/CLI/Logger.cs
erikzhang: But it makes no difference. If it is `null`, then `Combine` method will throw an exception.

neo-cli/CLI/Logger.cs
```diff
@@ -78,6 +78,8 @@ void ILogPlugin.Log(string source, LogLevel level, object message)
                         case LogLevel.Fatal: FatalColor.Apply(); break;
                         case LogLevel.Info: InfoColor.Apply(); break;
                         case LogLevel.Warning: WarningColor.Apply(); break;
+                        default:
+                            throw new ArgumentOutOfRangeException(nameof(level), level, null);
```
erikzhang, line 82: Why not just ignore it?

neo-cli/CLI/MainService.Node.cs
```diff
@@ -71,7 +71,7 @@ private void OnShowStateCommand()
                     NeoSystem.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(NativeContract.Ledger.CurrentIndex(NeoSystem.StoreView))));
                     await Task.Delay(NeoSystem.Settings.TimePerBlock, cancel.Token);
                 }
-            });
+            }, cancel.Token);
```
erikzhang, line 74: It's not necessary. Because it is already judged in the loop.

erikzhang (approved):

PR #848, opened 2022-01-22
featured log

Liaojinghui:
image
image

image

image

Original:

image


neo-cli/CLI/Logger.cs
```diff
@@ -11,97 +11,187 @@
...
 using Neo.Plugins;
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 using static System.IO.Path;
 
-namespace Neo.CLI
+namespace Neo.CLI;
```
erikzhang, line 20: For new files, we can use this new style. But for old files, we should not add unnecessary changes.

neo-cli/CLI/Logger.cs
```diff
@@ -65,43 +88,112 @@ void ILogPlugin.Log(string source, LogLevel level, object message)
...
+                                loglevel = "DEBUG";
+                                if (OperatingSystem.IsLinux() || OperatingSystem.IsMacOS())
+                                    messages[0] = $"🔨{messages[0]}";
+                                break;
+                            }
+
+                        case LogLevel.Error:
+                            logcolor = ErrorColor; loglevel = "ERROR";
+                            if (OperatingSystem.IsLinux() || OperatingSystem.IsMacOS())
+                                messages[0] = $"❌{messages[0]}";
```
shargon, line 115: If the software read the output, it could produce errors, it should be possible to disable it or detect if a process is reading the output

Liaojinghui: sorry i dont understand, why process reading the output could produce error here? its not yet an output but a string when i change it.

shargon: It will get colors and unicode icons

Liaojinghui: But i just added a few more extra operations here, the original logic still remains, if there is no issue in the past, why would be a problem here? I assume the `lock` already addresses your concern?

erikzhang:
I don't like this style. It adds a lot of unnecessary newlines. And I don't like the icons either.

Liaojinghui:
> I don't like this style. It adds a lot of unnecessary newlines. And I don't like the icons either.

Icon can be removed easily. new lines can be avoided. Formatted log makes it easier to address what is happening.

Liaojinghui:
Now icons are removed, long hash characters are abbreviated

neo-cli/CLI/Logger.cs
```diff
@@ -65,43 +89,117 @@ void ILogPlugin.Log(string source, LogLevel level, object message)
...
                     switch (level)
                     {
-                        case LogLevel.Debug: DebugColor.Apply(); break;
-                        case LogLevel.Error: ErrorColor.Apply(); break;
-                        case LogLevel.Fatal: FatalColor.Apply(); break;
-                        case LogLevel.Info: InfoColor.Apply(); break;
-                        case LogLevel.Warning: WarningColor.Apply(); break;
+                        case LogLevel.Debug:
+                            {
+#if RELEASE
```
erikzhang, line 103: Why this condition?

neo-cli/CLI/Logger.cs
```diff
@@ -1,34 +1,58 @@
...
+
+        /// <summary>
+        /// Process "log on" command
+        /// </summary>
+        [ConsoleCommand("log on", Category = "Log Commands")]
+        private void OnLogOnCommand()
+        {
+            if (Settings.Default.Logger.ConsoleOutput) showLog = true;
+            else ConsoleHelper.Warning("Please enable log first.");
+        }
```
erikzhang, line 54: Why do we need this method since it is already configured in `config.json`?

Liaojinghui: We need to check the log, but sometimes we also need to interact with the console, then `log` breaks our command....

neo-cli/CLI/Logger.cs
```diff
@@ -1,34 +1,58 @@
...
+        [ConsoleCommand("log off", Category = "Log Commands")]
+        private void OnLogOffCommand()
+        {
+            showLog = false;
+        }
+
+        /// <summary>
+        /// Process "log on" command
+        /// </summary>
+        [ConsoleCommand("log on", Category = "Log Commands")]
```
erikzhang, line 49: I think the command name is confused. It only control the `ConsoleOutput`, not to turn on or turn off the log actually.

Liaojinghui: `console log on` and `console log off`?

neo-cli/CLI/Logger.cs
```diff
@@ -1,34 +1,58 @@
...
         public override string Name => "SystemLog";
         public override string Description => "Prints consensus log and is a built-in plugin which cannot be uninstalled";
         public override string ConfigFile => Combine(GetDirectoryName(Path), "config.json");
         public override string Path => GetType().Assembly.Location;
+        private bool _showLog = Settings.Default.Logger.ConsoleOutput;
+
+        /// <summary>
+        /// Process "console log off" command to turn off console log
+        /// </summary>
+        [ConsoleCommand("console log off", Category = "Log Commands")]
```
erikzhang, line 41: You need to register the commands.

neo-cli/CLI/Logger.cs
```diff
@@ -65,43 +89,89 @@ void ILogPlugin.Log(string source, LogLevel level, object message)
             lock (typeof(Logger))
             {
                 DateTime now = DateTime.Now;
-                var log = $"[{now.TimeOfDay:hh\\:mm\\:ss\\.fff}] {message}";
-
-                if (Settings.Default.Logger.ConsoleOutput)
+                var log = $"[{now.TimeOfDay:hh\\:mm\\:ss\\.fff}]";
+                if (Settings.Default.Logger.ConsoleOutput && _showLog)
```
erikzhang, line 94:
```suggestion
                if (_showLog)
```

neo-cli/CLI/Logger.cs
```diff
@@ -65,43 +90,89 @@ void ILogPlugin.Log(string source, LogLevel level, object message)
...
-                        case LogLevel.Warning: WarningColor.Apply(); break;
+                        case LogLevel.Debug: logColor = DebugColor; logLevel = "DEBUG"; break;
+                        case LogLevel.Error: logColor = ErrorColor; logLevel = "ERROR"; break;
+                        case LogLevel.Fatal: logColor = FatalColor; logLevel = "FATAL"; break;
+                        case LogLevel.Info: logColor = KeyColor; logLevel = "INFO"; break;
+                        case LogLevel.Warning: logColor = WarningColor; logLevel = "WARN"; break;
+                        default: logColor = InfoColor; logLevel = "INFO"; break;
                     }
+                    logColor.Apply();
+                    Console.Write(logLevel + " ");
```
erikzhang, line 110: Since we've used colors (which are the most intuitive) to represent the different categories, is this line necessary?

neo-cli/CLI/Logger.cs
erikzhang: Wait for #865

erikzhang:
I'd say I like `console log on` and `console log off`. But I still prefer the original whole line in the same color, which looks more intuitive.

nicolegys:
The log message does not write to file now.

Liaojinghui:
@nicolegys  can you test again, i could not repeat the problem, everything works well on my machine.

nicolegys:
> @nicolegys can you test again, i could not repeat the problem, everything works well on my machine.

neo-cli/CLI/MainService.Logger.cs
```diff
@@ -71,43 +93,89 @@ private void OnLog(string source, LogLevel level, object message)
...
-                    path = Combine(path, $"{now:yyyy-MM-dd}.log");
-                    try
-                    {
-                        File.AppendAllLines(path, new[] { $"[{level}]{log}" });
-                    }
-                    catch (IOException)
-                    {
-                        Console.WriteLine("Error writing the log file: " + path);
-                    }
+                    File.AppendAllLines(path, new[] { $"[{level}]{log}" });
```
nicolegys, line 148:
```suggestion
                    File.AppendAllLines(path, new[] { $"[{level}]{log} {message}"});
```
`message` is missing here, how could everything work well on your machine:joy:

Liaojinghui: My bad, i must have checked the wrong log file

nicolegys (approved):
pass

Liaojinghui:
@erikzhang @shargon ready to merge.

neo-cli/CLI/MainService.Logger.cs
```diff
@@ -71,43 +93,89 @@ private void OnLog(string source, LogLevel level, object message)
...
+                        case LogLevel.Warning: logColor = WarningColor; logLevel = "WARN"; break;
+                        default: logColor = InfoColor; logLevel = "INFO"; break;
                     }
+                    logColor.Apply();
+                    Console.Write(logLevel + " ");
+                    InfoColor.Apply();
 
-                    Console.WriteLine(log);
+                    Console.Write($"{log} {messages[0]}");
+                    for (var i = 0; i < 35 - messages[0].Length - logLevel.Length; i++) Console.Write(' ');
```
shargon, line 117: Only one `Console.Write`?

neo-cli/CLI/MainService.Logger.cs
```diff
@@ -71,43 +93,89 @@ private void OnLog(string source, LogLevel level, object message)
...
+                    File.AppendAllLines(path, new[] { $"[{level}]{log} {message}" });
                 }
+                catch (IOException)
+                {
+                    Console.WriteLine("Error writing the log file: " + path);
+                }
+            }
+        }
+
+        private static string[] Parse(string message)
```
shargon, line 157: Add a comment about the expected format to parse?

shargon: I think we never follow any standard in messages.

Liaojinghui: > I think we never follow any standard in messages.

I will work on that in my next pr after this pr is merged. I have discussed with @nicolegys  and decided to make the log system more test friendly.

erikzhang (dismissed):
I like `console log on` and `console log off`. But I still prefer the original whole line in the same color, which looks more intuitive.

neo-cli/CLI/MainService.Logger.cs
```diff
@@ -71,43 +93,78 @@ private void OnLog(string source, LogLevel level, object message)
...
                     {
-                        case LogLevel.Debug: DebugColor.Apply(); break;
-                        case LogLevel.Error: ErrorColor.Apply(); break;
-                        case LogLevel.Fatal: FatalColor.Apply(); break;
-                        case LogLevel.Info: InfoColor.Apply(); break;
-                        case LogLevel.Warning: WarningColor.Apply(); break;
+                        case LogLevel.Debug: logColor = DebugColor; logLevel = "DEBUG"; break;
+                        case LogLevel.Error: logColor = ErrorColor; logLevel = "ERROR"; break;
+                        case LogLevel.Fatal: logColor = FatalColor; logLevel = "FATAL"; break;
+                        case LogLevel.Info: logColor = KeyColor; logLevel = "INFO"; break;
```
nicolegys, line 108:
```suggestion
                        case LogLevel.Info: logColor = InfoColor; logLevel = "INFO"; break;
```
Ok for me, except the green all over the window. Could you set the info log color as white?

neo-cli/CLI/MainService.Logger.cs
nicolegys: > Add a comment about the expected format to parse?

Agree, we'd better add a comment. @Liaojinghui

nicolegys (dismissed):
Test OK. @erikzhang

neo-cli/CLI/MainService.Logger.cs
```diff
@@ -23,8 +25,10 @@ partial class MainService
         private static readonly ConsoleColorSet WarningColor = new ConsoleColorSet(ConsoleColor.Yellow);
         private static readonly ConsoleColorSet ErrorColor = new ConsoleColorSet(ConsoleColor.Red);
         private static readonly ConsoleColorSet FatalColor = new ConsoleColorSet(ConsoleColor.Red);
+        private static readonly ConsoleColorSet KeyColor = new ConsoleColorSet(ConsoleColor.DarkGreen);
```
erikzhang, line 28: This field is never used.

neo-cli/CLI/MainService.Logger.cs
erikzhang: Now that it's all the same color, do we still need to parse the message?

Liaojinghui:
@shargon can you please review this pr?

neo-cli/CLI/MainService.Logger.cs
```diff
@@ -1,30 +1,33 @@
 // Copyright (C) 2016-2022 The Neo Project.
-// 
-// The neo-cli is free software distributed under the MIT software 
+//
+// The neo-cli is free software distributed under the MIT software
 // license, see the accompanying file LICENSE in the main directory of
-// the project or http://www.opensource.org/licenses/mit-license.php
```
vncoelho, line 5: These lines were format correctly? Is it needed?

vncoelho (commented):
It looks much better, specially for consensus logs. I am in favor of this feature.

neo-cli/CLI/MainService.Logger.cs
shargon: Ping @Liaojinghui

superboyiii:
@shargon Could you review again? Tested OK.

Liaojinghui:
@shargon Hey shargon, I am not sure what exactly is your concern, just answer the following two previously unaddressed questions:

> Only one `Console.Write`?

Updated into one line.

> Now that it's all the same color, do we still need to parse the message?

Indeed we still need to parse the message, it is a historic issue as the previous log has messages sticked together while we need to separate them apart.

shargon (approved):

superboyiii (approved):

vncoelho (approved):

vncoelho:
> I like `console log on` and `console log off`. But I still prefer the original whole line in the same color, which looks more intuitive.

@erikzhang, your review is blocking.
Can we proceed like this or you still want it to be off as standard?

vncoelho:
@Liaojinghui, I think there should be some command to remove colors and keep original compatibility.
Perhaps some people will prefer it clean as the original one.

Liaojinghui:
> @Liaojinghui, I think there should be some command to remove colors and keep original compatibility. Perhaps some people will prefer it clean as the original one.

It is already in the black/white mode for the most log, only errors will show an red [Error] and warning a yellow [warning], the rest remains black/white

vncoelho:
Let me try again. Last time I saw it differently.

neo-cli/CLI/MainService.Logger.cs
```diff
@@ -71,43 +92,80 @@ private void OnLog(string source, LogLevel level, object message)
...
-                        case LogLevel.Warning: WarningColor.Apply(); break;
+                        case LogLevel.Debug: logColor = DebugColor; logLevel = "DEBUG"; break;
+                        case LogLevel.Error: logColor = ErrorColor; logLevel = "ERROR"; break;
+                        case LogLevel.Fatal: logColor = FatalColor; logLevel = "FATAL"; break;
+                        case LogLevel.Info: logColor = InfoColor; logLevel = "INFO"; break;
+                        case LogLevel.Warning: logColor = WarningColor; logLevel = "WARN"; break;
+                        default: logColor = InfoColor; logLevel = "INFO"; break;
+                    }
+                    logColor.Apply();
+                    Console.Write($"{logLevel} {log} \t{messages[0],-20}");
```
vncoelho, line 112: What are these ones here and this if that checks Length > 20?

Maybe add some comment.

Liaojinghui: If they are longer than 20, the longer one will not align that well., the rest message will be pushed backwards. But in most (actually i would say all) existing situations, it aligns pretty good.

vncoelho:
Tested again with long lines and works fine.
Colors are just for warning and error and it is almost the same as default.
Easier for futures control and modifications as well.

PR #850, opened 2022-02-06
neo-cli: tune MaxTransactionsPerBlock for testnet

roman-khimov:
Allow to flush the mempool in just 10 blocks.

### DIFF
--- a/neo-cli/config.testnet.json
+++ b/neo-cli/config.testnet.json
@@ -26,7 +26,7 @@
     "Network": 877933390,
     "AddressVersion": 53,
     "MillisecondsPerBlock": 15000,
-    "MaxTransactionsPerBlock": 512,
+    "MaxTransactionsPerBlock": 5000,
     "MemoryPoolMaxTransactions": 50000,
     "MaxTraceableBlocks": 2102400,
     "InitialGasDistribution": 5200000000000000,


Liaojinghui:
Do we have an approximate maximum execution time of a single transction, just in case of Resource Exhaustion Attack?

roman-khimov:
It's strongly correlated with GAS cost and we have MaxBlockSystemFee limit (somewhat optimistic I'd say, but still). We also know that C# node can handle several thousands of NEO transfers per second, so with 15s block time 5K doesn't seem to be a lot of transactions.

shargon (approved):

Liaojinghui:
@roman-khimov I know the awesome benchmark u made on neo N3. Actually I always wanna know what exactly makes C# slower than go node. But anyway, benchmark won't stop resource exhaustion attack, strict gas restrictions may work, I will check it out.

roman-khimov:
DoS scenarios are always limited by GAS, sending lots of transactions costs something and it's not hard to calculate the cost of the attack. Actually, this change raises the cost of the attack because to cause some noticeable delays one would need to send more transactions and spend more GAS.

As for computational overhead of big blocks, `MaxBlockSystemFee` is big enough already to be noticeable and if someone has enough GAS to spend he might keep the nodes somewhat busy even with a single transaction (just ask @vang1ong7ang). But then again, the attacker is limited by the amount of GAS (money) he has for attack.

Liaojinghui (approved):

roman-khimov:
> I will do a thorough benchmark on opcodes in another work.

Take a look at neo-project/neo#1875 also, although both VMs have improved since then I think.

Liaojinghui:
> Take a look at [neo-project/neo#1875](https://github.com/neo-project/neo/issues/1875) 

Thank you.

superboyiii:
@roman-khimov Could you push it to `develop` branch? We'd like to keep `master` clear before next release.

roman-khimov:
> @roman-khimov Could you push it to `develop` branch?

Sure, done.

erikzhang (approved):

PR #851, opened 2022-03-08
Update seedlist for testnet

superboyiii:
seed1 and seed1t4 are using the same ip, the same to seed2 ~ seed5


shargon:
It's the same IP but you can change it in a future, I think that different dns is ok

PR #852, opened 2022-03-28
Add new testnet network id

superboyiii:
Since testnet data is incompatible now becasue of the modification of native contract, we could create a new testnet `N3T5`. Wait for a better solution on native contract update.

### DIFF
--- a/neo-cli/config.testnet.json
+++ b/neo-cli/config.testnet.json
@@ -23,7 +23,7 @@
     }
   },
   "ProtocolConfiguration": {
-    "Network": 877933390,
+    "Network": 894710606,
     "AddressVersion": 53,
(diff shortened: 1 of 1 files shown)


erikzhang (approved):

Liaojinghui:
Just curious, does the id has any special meaning? or its just randomly picked?

superboyiii:
> Just curious, does the id has any special meaning? or its just randomly picked?



> Just curious, does the id has any special meaning? or its just randomly picked?

BigInteger -> Hex -> UTF8 string -> `N3T5`

Liaojinghui:
> BigInteger -> Hex -> UTF8 string -> `N3T5`
LOOOOOOL, Impressive~~~~~way more complex than i thought

PR #853, opened 2022-04-03
Mode system

Liaojinghui:
updated to work under https://github.com/neo-project/neo-modules/pull/709

This pr implements neo-project/neo-node#863
Mode system is intended to help developers to save and load neo node configurations.

1. cli provides some default settings, such as mainnet, testnet, private net. Maybe Oracle net and FS net in the future.
2. user can load from these default settings.
3. user can have their own node setttings and save them as a mode
4. user can delete existing modes
5. if no mode is assigned from the console, load `mainnet`
6. Does **not** support runtime mode switch.

image

image

`mainnet` mode:

```shell
dotnet neo-cli.dll /mode mainnet
```

`testnet` mode:

```shell
dotnet neo-cli.dll /mode testnet
```

`privatenet` mode:

```shell
dotnet neo-cli.dll /mode privatenet
```

when user load into the privatenet mode, a single node blockchain network will be setup with the default primary account NbVj8GhwToNv4WF2gVaoco6hbkMQ8hrHWP. Please note that the secret key of this account is publicly available to everyone, 
### **do not send any asset to this account in main net**
you will lose your token and we will not be responsible for that.

**Create your own mode**
You can config your own network by modifying the `config.json` and `config.fs.json` in the root directory, as well as the `config.json` of Plugins. Then run `mode save [mode name]` to save your configuration as a mode. If the name you assigned is an existing mode, existing mode will be updated.

> Your configurations will not be updated into mode automatically, you must manually run `mode save [mode name]` to save them.


neo-cli/Modes/Mode
```diff
@@ -0,0 +1 @@
+mainnet
```
erikzhang, line 1: Why do we need this file? We can hardcode the default to be `mainnet`.

Liaojinghui: I was considering to set the user assigned mode as the default mode, which might change regularly. But its ok if we only set `mainnet` as default mode. has updated.

neo-cli/Modes/privatenet/private.json
```diff
@@ -0,0 +1 @@
+{"name":null,"version":"1.0","scrypt":{"n":16384,"r":8,"p":8},"accounts":[{"address":"NbNRXrkfMyrEyEv2euE63Xxxqkno7xvuk2","label":null,"isDefault":false,"lock":false,"key":"","contract":{"script":"","parameters":[{"name":"signature","type":"Signature"}],"deployed":false},"extra":null}],"extra":null}
```
erikzhang, line 1: Why do we need this file?

Liaojinghui: anyone who want to setup a private single node network. such that they can start the private network without modifying anyting.

Liaojinghui: Without this given account, they will need to create an account by themself, then config the config.json manually.

erikzhang: But there is no private key in the wallet. The user still have to create another wallet.

Liaojinghui: Forgot that, will update right now.

erikzhang: Can we move all the json files to a single folder?

Liaojinghui: We can not, they are of the same name~~~all of them are `config.json`

erikzhang: `config.xxx.json`?

Liaojinghui: emmmmmmmmm, good idea, will do.

Liaojinghui: `config.json` `config.fs.json` `DBFTPlugin.json` `ApplicationLog.json` under same folder now

nicolegys:
`mode delete`
Cannot delete a directory which is not empty.
Image

`save as an existing mode name`
There is no warning and the original mode configs will be overwritten.
Image

nicolegys:
The logic of `mode save` is a little strange for me.
If I want to save a mode, firstly I need to start neo-cli with some specified mode in `./Mode/` folder or with default mode `mainnet`, and corresponding config files will be generated in `./` folder. Then I modify or replace the config files in `./` folder , use `mode save <modeName>` cmd to copy the config files to `./Mode/` folder. Why don't I directly create a new mode in `./Mode/` folder?:joy:

Liaojinghui:
> The logic of `mode save` is a little strange for me. If I want to save a mode, firstly I need to start neo-cli with some specified mode in `./Mode/` folder or with default mode `mainnet`, and corresponding config files will be generated in `./` folder. Then I modify or replace the config files in `./` folder , use `mode save <modeName>` cmd to copy the config files to `./Mode/` folder. Why don't I directly create a new mode in `./Mode/` folder?😂

Cause `config.json` are spread everywhere, with `save mode` we can collect `config` file from the `node ` and `Plugins`. If user directly create mode under `Modes` folder, how could they deal with `Plugins`.

neo-cli/CLI/MainService.Mode.cs
```diff
@@ -0,0 +1,137 @@
+// Copyright (C) 2016-2021 The Neo Project.
```
Liaojinghui, line 1:
```suggestion
// Copyright (C) 2016-2022 The Neo Project.
```

neo-cli/Modes/privatenet/private.json
erikzhang: They are still in different folders.

Liaojinghui:
@nicolegys May you please test this?

nicolegys (changes requested):
@Liaojinghui :joy:

neo-cli/CLI/MainService.Mode.cs
```diff
@@ -0,0 +1,199 @@
...
+            var dir = new DirectoryInfo($"./Modes/{mode.ToLower()}");
+            if (!dir.Exists)
+                throw new DirectoryNotFoundException($"Mode not found: {dir.FullName}");
+
+            // Get the config files of the node
+            foreach (var file in dir.GetFiles())
+            {
+                var targetFilePath = Path.Combine("./", file.Name);
+                file.CopyTo(targetFilePath, true);
+            }
```
nicolegys, line 126:
```suggestion

```

neo-cli/CLI/MainService.Mode.cs
```diff
@@ -0,0 +1,199 @@
...
+            // Process the plugin
+            var modePlugins = dir.GetDirectories();
```
nicolegys, line 128: `dir.GetDirectories` cannot find plugins, because you use `{plugin.Name}.json` when saving a mode

nicolegys (changes requested):

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -141,6 +142,18 @@ private async Task<MemoryStream> DownloadPluginAsync(string pluginName)
                 await InstallDependenciesAsync(es, installed);
             }
             zip.ExtractToDirectory("./", true);
+            Console.WriteLine();
+            // Save the config.json to current mode
+            try
+            {
+                // what if the file already exists in the mode? OK, lets overwrite it then.
+                File.Copy($"Plugins/{pluginName}/config.json", $"Modes/{_currentMode}/{pluginName}.json", true);
```
nicolegys, line 150: > https://github.com/neo-project/neo-node/pull/853/files#diff-aa0b3721c73c0224f617d2fe705a28faccd60158bd47ba54b009007d9f23e690R125-R126

you are replacing one config file by itself.

nicolegys (changes requested):

Liaojinghui:
> After I install a new plugin, neo-cli will remind me to restart it. But if I restart it immediately without saving this mode, in this pr, the new installed plugin will be removed. Need to add some words in this warning to remind users saving mode before restarting neo-cli.
> 
> > https://github.com/neo-project/neo-node/blob/fd9203f1ce75f163b8c6a8d9870cd20c6abbc4c3/neo-cli/CLI/MainService.Plugins.cs#L43

This issue should have been solved in my last commit. The reason of that issue is the plugin folder name is kinda different from the plugin name.

neo-cli/CLI/MainService.Plugins.cs
nicolegys: > I have to do this to make sure a plugin uses the config of current mode.

your logic:
1.start neo-cli with `mode A`，in which config file `k.json` of `Plugin k` is included.
2.auto-install `Plugin k` because `k` is not in `Plugin` folder
3.overwrite `Modes/A/k.json` with default config file of `Plugin k` in `Plugins/k/config.json`
4.copy `Modes/A/k.json` (now it is the default config file) to `Plugins/k/config.json`

In fact, after all of the operations, the config file of `Plugin k` is still the default one, not the one you saved in `mode/A` folder.

nicolegys:
Need to remind users to restart/resync neo-cli if new plugins are auto-installed.

nicolegys:
exception.

Liaojinghui:
> Need to remind users to restart/resync neo-cli if new plugins are auto-installed.

Well, actually, it is supposed to be loaded automatically....will check this logic.

neo-cli/CLI/MainService.Mode.cs
```diff
@@ -0,0 +1,196 @@
...
+
+            modePlugins.ForEach(async p =>
+            {
+                var pluginName = Path.GetFileNameWithoutExtension(p.Name);
+                // if the plugin does not exist, maybe consider install it
+                if (p.Name.Contains("config")) return;
+                if (!Directory.Exists($"Plugins/{pluginName}/"))
+                {
+                    await InstallPluginAsync(pluginName, overWrite:true, saveConfig:false);
+                    _needRestart = true;
```
nicolegys, line 125: There is no way to ensure that this line is executed before `if (_needRestart)`, because it is executed asynchronously.

nicolegys:
async issue still exists.

neo-cli/CLI/MainService.cs
```diff
@@ -371,14 +373,45 @@ public async void Start(string[] args)
...
+                            }
+                        }
+                        break;
                 }
+            // Load the mode (network)
+            await LoadMode(mode);
+            if (_needRestart)
+            {
+                ConsoleHelper.Warning("Please restart the node to apply the changes.");
+                OnStop();
```
nicolegys, line 412: This line causes the following exception on exit.

superboyiii:
@Liaojinghui Can we move on? I think it's useful.

nicolegys:
1. If plugins are auto-installed/auto-uninstalled when starting `neo-cli`, type `exit` and `neo-cli` will throw an exception.


2. `reinstall plugin` will replace this plugin's config file in `mode\xxx` by the default config file.
3. e.g. Both `ApplicationLogs` and `RpcServer` are not installed but in `\mode\testnet`, now I start `neo-cli` with `testnet` mode, I could find that `RpcServer`'s config file in `\mode\testnet` be replaced by the default one. Because `RpcServer` is installed as a dependency of `ApplicationLogs`.

nicolegys:
> 1. If plugins are auto-installed/auto-uninstalled when starting `neo-cli`, type `exit` and `neo-cli` will throw an exception.
>    
> 2. `reinstall plugin` will replace this plugin's config file in `mode\xxx` by the default config file.
> 3. e.g. Both `ApplicationLogs` and `RpcServer` are not installed but in `\mode\testnet`, now I start `neo-cli` with `testnet` mode, I could find that `RpcServer`'s config file in `\mode\testnet` be replaced by the default one. Because `RpcServer` is installed as a dependency of `ApplicationLogs`.

`2. ...` hasn't been fixed.

nicolegys:
If use lower case to install plugins, it can't find folders like `Plugins/applicationlogs`.

nicolegys:
This issue still exists.
#880

nicolegys:
Threw an exception when reinstalling StateService, but not always.
Image

nicolegys:
> 1. If plugins are auto-installed/auto-uninstalled when starting `neo-cli`, type `exit` and `neo-cli` will throw an exception.
>    
> 2. `reinstall plugin` will replace this plugin's config file in `mode\xxx` by the default config file.
> 3. e.g. Both `ApplicationLogs` and `RpcServer` are not installed but in `\mode\testnet`, now I start `neo-cli` with `testnet` mode, I could find that `RpcServer`'s config file in `\mode\testnet` be replaced by the default one. Because `RpcServer` is installed as a dependency of `ApplicationLogs`.

3. ... hasn't been fixed.
**To Reproduce**
1.Make sure that `ApplicationLogs` is before `RpcServer` in `.PLUGINS` file, and both of them are not installed.

2.Start neo-cli.

Liaojinghui:
> Threw an exception when reinstalling StateService, but not always. Image

This is an issue in `neo core`, and has being addressed in https://github.com/neo-project/neo/pull/2569 . This pr can not and should not handle that.

nicolegys:
No more issues，need review.

superboyiii:
@Liaojinghui Please fix format.

neo-cli/neo-cli.csproj
```diff
@@ -13,14 +13,17 @@
...
+    <Content Include="Modes\**">
+        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        <TargetPath>Modes\%(RecursiveDir)\%(Filename)%(Extension)</TargetPath>
+    </Content>
+</ItemGroup>
+
+
 
   <ItemGroup>
     <PackageReference Include="Neo" Version="3.4.0" />
```
nicolegys, line 29: @Liaojinghui need to update Neo reference to 1391

neo-cli/Modes/privatenet/.PLUGINS
```diff
@@ -0,0 +1,4 @@
+﻿LevelDBStore
+RpcServer
+ApplicationLogs
+DBFTPlugin
```
nicolegys, line 4:
```suggestion
```

neo-cli/Modes/privatenet/config.json
```diff
@@ -0,0 +1,40 @@
...
+      "Engine": "LevelDBStore",
+      "Path": "Data_LevelDB_{0}"
+    },
+    "P2P": {
+      "Port": 10333,
+      "WsPort": 10334,
+      "MinDesiredConnections": 10,
+      "MaxConnections": 40,
+      "MaxConnectionsPerAddress": 3
+
```
nicolegys, line 18:
```suggestion
```

nicolegys:
It uses an error plugin path when starting neo-cli in the different directory with `neo-cli.dll`.

neo-cli/Modes/privatenet/.PLUGINS
Liaojinghui: I meant to make the private net more powerful

Liaojinghui:
>  It uses an error plugin path when starting neo-cli in the different directory with `neo-cli.dll`.

The path issue is not a problem should be addressed in this pr. it is a nature feature of C# on windows platform.

nicolegys:
> >  It uses an error plugin path when starting neo-cli in the different directory with `neo-cli.dll`.
> 
> The path issue is not a problem should be addressed in this pr. it is a nature feature of C# on windows platform.
> 
> Relative path will always be relative to the executing folder, unless you publish it in a self-contain mode.

It was published in the self-contain mode.
Actually, this issue is because method `PluginExists` uses the wrong path to find the specified plugin.

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -173,7 +208,8 @@ private async Task InstallDependenciesAsync(Stream config, HashSet<string> insta
         /// <returns></returns>
         private static bool PluginExists(string pluginName)
         {
-            return Plugin.Plugins.Any(p => p.Name.Equals(pluginName, StringComparison.InvariantCultureIgnoreCase));
+            return Plugin.Plugins.Any(p => p.Name.Equals(pluginName, StringComparison.InvariantCultureIgnoreCase)) ||
+                   new DirectoryInfo("Plugins").GetDirectories().Any(p => p.Name.Equals(pluginName, StringComparison.InvariantCultureIgnoreCase));
```
nicolegys, line 212:
```suggestion
                   new DirectoryInfo($"{StrExeFilePath}/Plugins").GetDirectories().Any(p => p.Name.Equals(pluginName, StringComparison.InvariantCultureIgnoreCase));
```

neo-cli/CLI/MainService.Plugins.cs
```diff
@@ -140,7 +142,40 @@ private async Task<MemoryStream> DownloadPluginAsync(string pluginName)
                 await using Stream es = entry.Open();
                 await InstallDependenciesAsync(es, installed);
             }
-            zip.ExtractToDirectory("./", true);
+            if (!Directory.Exists($"{StrExeFilePath}Plugins"))
+            {
+                Directory.CreateDirectory($"{StrExeFilePath}Plugins");
+            }
```
nicolegys, line 148:
```suggestion
            if (!Directory.Exists($"{StrExeFilePath}/Plugins"))
            {
                Directory.CreateDirectory($"{StrExeFilePath}/Plugins");
            }
```

neo-cli/Modes/privatenet/.PLUGINS
nicolegys: > I meant to make the private net more powerful

If one plugin is added in `.PLUGINS` without its config file in mode folder, then start neo-cli, it will be auto-installed but its config file won't be auto-saved. We need to check this logic to make it more accurate.

nicolegys:
In addtion, as you said, relative path is an old issue that also exists in master branch, and it happens not only in the places I mentioned in this pr. I'll create annother issue if needed.

Liaojinghui:
> In addtion, as you said, relative path is an old issue that also exists in master branch, and it happens not only in the places I mentioned in this pr. I'll create annother issue if needed.

You dont have to, cause this issue does not exist in the official publish package as it is published in the self-contain mode.

nicolegys:
> > In addtion, as you said, relative path is an old issue that also exists in master branch, and it happens not only in the places I mentioned in this pr. I'll create annother issue if needed.
> 
> You dont have to, cause this issue does not exist in the official publish package as it is published in the self-contain mode.

I have said that this issue also exists when it is published in the self-contain mode. I've tested it both on v3.4.0 and master branch, both on windows and linux. The cause of this issue is, in our code, files are read or generated based on the current execution directory, not the relative directory. And this issue was not raised before because of few usage scenarios, so it's not a high priority. Please be more careful.

superboyiii:
@shargon Could you review again?

.github/workflows/test-neo-cli.expect
```diff
@@ -14,6 +14,10 @@ expect {
     timeout { exit 1 }
 }
 
+sleep 5
```
shargon, line 17: 5 seconds is required?

Liaojinghui: > 

Yes, has to wait for 5 seconds

superboyiii:
@Liaojinghui Please solve conflicts

Liaojinghui:
> @Liaojinghui Please solve conflicts

Resolved

Issue #824, opened 2021-09-24
node: ? ?? New Release - a Docker image, plz


PureKrome:
It's 2021. Lets not ask people to download code to install the CLI or even download an exe _and then install_ an extra db, etc.

Lets make this _really simple_ for developers and offer a docker image with everything in it.

please.

thank you kindly! 🍰

vncoelho:
Agreed.

Liaojinghui:
Except for the docker image, i also think we need some scripts for different environment setup or tests.

PureKrome:
> Except for the docker image, i also think we need some scripts for different environment setup or tests.

And what would these scripts be doing? As in ... as a developer, i don't want scripts because this means i have to do stuff to setup my environment to experiment with Neo3. Too much of a barrier to entry. All i should do is

=> docker run -it neo3 

or whatever. and off i go.

Liaojinghui:
> And what would these scripts be doing? As in ... as a developer, i don't want scripts because this means i have to do stuff to setup my environment to experiment with Neo3. Too much of a barrier to entry. All i should do is

script to run in testnet, script to run in mainnet, script to run in private net, script to install everything needed in a new syste

PureKrome:
ah yes. misunderstood! yep, some simple CLI commands would definitely simplify things

hurui200320:
For those who just want to spin up a node, I really like a docker image so I can `docker run ...`. 

However, I do notice the modular design for N3 nodes. Like the plugin system, if you're on the testnet, when you download a plugin, it's likely your node is running on testnet, but your plugin is configured for mainnet (because of the default value). So I'm expecting some code refactor first, like allowing plugins to fetch network id (and other common settings) from an internal API, or something else (I'm not writing any C#, so I'm sorry if I was wrong).

Then we can have a better experience with the docker image.

Or we can have some shell script to build docker image locally, based on what you need. For example, I need the plugin A, B, and C, and a node for testnet. Then the script will clone the node's code, compile it, download the desired plugins, make the config, then pack them into a docker image. 

Or just offers different levels of docker images. Like `light` for bare nodes, `medium` for every plugin except consensus, and `full` for consensus nodes.

Liaojinghui:
Is this what you are looking for #853 ?

hurui200320:
[...]

PR #856, opened 2022-04-09
Get plugin network

Liaojinghui:
Allow users to check the plugin network configuration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [3.2.0]
+
+### Changed
+- ([#850](https://github.com/neo-project/neo-node/pull/850/)) Modify MaxTransactionsPerBlock for testnet
+- ([#852](https://github.com/neo-project/neo-node/pull/852/)) Add new testnet network id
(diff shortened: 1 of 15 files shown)


erikzhang:
Some plugins support multiply networks. For example, [RpcServer](https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServer/config.json).

Liaojinghui:
> Some plugins support multiply networks. For example, [RpcServer](https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServer/config.json).

OOOOOOK, need more work on them then~~~~

erikzhang:
And why do we need this command?

Liaojinghui:
> And why do we need this command?

Cause it is hard to figure out what is happening when you have everything right but the node never run correctly, just like what is mentioned here https://github.com/neo-project/neo-node/issues/824#issuecomment-1090464930

erikzhang:
If you can think of using this command, then you must have realized what the problem is, then this command is useless.

Liaojinghui:
I would say, locating issue with one command is way batter than "get into multiple plugins, open their config.json files, then compare the network cross multiple files". Cli is supposed to make things easier right?

PR #893, opened 2023-03-13
Validate before deploy update

Liaojinghui:
@shargon how about updating the script checking logic before we issue the transaction from cli.


shargon:
I prefer to do it in both, in core and in neo-cli

Liaojinghui:
The behavior of the docker test appears to be unpredictable, as it has shown inconsistent results after being run on the same code multiple times - sometimes passing and sometimes failing without any discernible pattern, despite being tested over 30 times.

shargon (approved):

PR #898, opened 2023-08-12
set the testnet hardfork height to 0

Liaojinghui:
Possible solution to https://github.com/neo-project/neo/pull/2886

If we set up a new testnet or a private net, all hardforks are enabled by default since the genesis.


shargon:
it was no conflicts?

PR #901, opened 2023-09-28
Cli: Adding Script System for neo-cli

Liaojinghui:
This amazing idea is from 17v.

By enabling developers to write and run script in the neo-cli, they are empowerd to directly interact with the leder and smart contract directly with C# code. 

What's the benefit?

- Typed variables.
- Code recommendation.
- Combined commands.
- Loop
- Event monitor.
- And more.

### DIFF
--- a/neo-cli/template.cs
+++ b/neo-cli/template.cs
@@ -1,0 +1,76 @@
+// Copyright (C) 2016-2023 The Neo Project.
+//
+// The neo-cli is free software distributed under the MIT software
+// license, see the accompanying file LICENSE in the main directory of
+// the project or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+
+using System;
+using System.Linq;
+using Neo;
+using Neo.SmartContract;
+using Neo.SmartContract.Native;
+using Neo.Network.P2P.Payloads;
+using Neo.Wallets;
+using Neo.VM;
+using Neo.ScriptHelper;
+
+#if DEBUG
+NeoSystem NeoSystem = new(null);
+#endif
+
+// Create an instance of ScriptHelper
+ScriptHelper helper = new ScriptHelper(NeoSystem);
+
+// loading key pair
+var keyPair = new KeyPair(Wallet.GetPrivateKeyFromWIF("xxxxxxxxxxxxxxxxxxxxxxxx-your private key"));
+var scriptHash = Contract.CreateSignatureContract(keyPair.PublicKey).ScriptHash;
+Console.WriteLine($"ME: {scriptHash} {scriptHash.ToAddress(NeoSystem.Settings.AddressVersion)}");
+
+// get snapshot
+var snapshot = NeoSystem.GetSnapshot();
+// get latest block index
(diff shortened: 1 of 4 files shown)


Liaojinghui:
@shargon  how do you think about this one, interact with the neo node via script

shargon:
> @shargon  how do you think about this one, interact with the neo node via script

Is not the same as `invokeScript` in rpc? Or what kind of script it is?

shargon:
> @shargon  it is C# script, where you can write c# code to construct transactions and interact with the neo blockchain, then run the script in the `cli`, it is painful to use cli since there is no type system right? now this is the solution, you can use this to conveniently do complex operations however you like. Airdrop, event monitoring, all can be done in script.
> 
> take a look at the example here: [neo-cli/template.cs](https://github.com/neo-project/neo-node/blob/2ee16b84fc41e87cd482064094d655398be89beb/neo-cli/template.cs)

Can we avoid to use reflection in that script?

Liaojinghui:
> Can we avoid to use reflection in that script?

i am afrait this is how it works.

shargon:
> > Can we avoid to use reflection in that script?
> 
> i am afrait this is how it works. You have security concerns?

I would not like the user to be able to access to the Storage or internal methods with this script.

Liaojinghui:
@shargon how do you think?

shargon:
I'm worried about people who make scripts to supposedly do something, and people who don't know how to program execute this code.

Liaojinghui:
this can be disabled by default tough, only guys who need it and konws how to use it can able it.

shargon:
> this can be disabled by default tough, only guys who need it and konws how to use it can able it.

Then, It's better to do it as a module.

neo-cli/template.cs
```diff
@@ -0,0 +1,76 @@
...
+    ValidUntilBlock = blockIndex + 1,
+};
+
+// sign the transaction
+var signature = transaction.Sign(keyPair, NeoSystem.Settings.Network);
+var invocationScript = new byte[] { ((byte)OpCode.PUSHDATA1), 64 }.Concat(signature).ToArray();
+var verificationScript = Contract.CreateSignatureContract(keyPair.PublicKey).Script;
+transaction.Witnesses = new Witness[] { new Witness { InvocationScript = invocationScript, VerificationScript = verificationScript } };
+Console.WriteLine($"HASH: {transaction.Hash}");
+Console.WriteLine($"{transaction.Verify(NeoSystem.Settings, snapshot, new(), new Transaction[] { })}");
```
cschuchardt88, line 63: Maybe should be `Debug.Assert`?

neo-cli/template.cs
```diff
@@ -0,0 +1,76 @@
...
+var transaction = new Transaction
+{
+    Version = 0,
+    Nonce = (uint)new Random().NextInt64(),
+    SystemFee = 1000000,
+    NetworkFee = 0x3e500,
+    Attributes = new TransactionAttribute[],
+    Script = sb.ToArray(),
+    Signers = new Signer[] { new Signer { Account = scriptHash, Scopes = WitnessScope.CalledByEntry } },
+    ValidUntilBlock = blockIndex + 1,
```
cschuchardt88, line 54: Plus `NeoSystem.Settings.MaxValidUntilBlockIncrement`?

PR #916, opened 2025-11-12
Required changes for Exec factor

shargon:
Required for https://github.com/neo-project/neo/pull/4278
- It require a release of neo nuget first


shargon:
> test fails. Could you check it?

It require #4278 from neo to be released first

PR #917, opened 2025-11-17
Add DNS oracle protocol

Jim8y:
## Summary
- add DNS-over-HTTPS oracle protocol and config wiring
- document dns:// usage and add tests for certificate parsing

## Testing
- not run (not requested)


shargon (commented):
I think that it's better to define the dns server, and make a dns query, instead of http query

erikzhang:
> I think that it's better to define the dns server, and make a dns query, instead of http query

We need DNS over HTTPS.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,448 @@
...
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Web;
+
+namespace Neo.Plugins.OracleService.Protocols;
+
+class OracleDnsProtocol : IOracleProtocol
```
shargon, line 41: Rename to OracleDoHProtocol ?

docs/oracle-dns-protocol.md
```diff
@@ -0,0 +1,128 @@
...
+
+- `EndPoint` must point to a DoH resolver that understands the [application/dns-json](https://developers.cloudflare.com/api/operations/dns-over-https) format.
+- `Timeout` is the maximum milliseconds the oracle will wait for a DoH response before returning `OracleResponseCode.Timeout`.
+
+> You can run your own DoH gateway and point the oracle to it if you need custom trust anchors or strict egress controls.
+
+## Constructing dns:// URLs
+
+```
+dns://<base-domain>/<path-selector>?selector=<label>&name=<override>&type=<rr-type>&format=<output>
```
erikzhang: We need to be consistent with [RFC4501](https://www.rfc-editor.org/rfc/rfc4501.html).

plugins/OracleService/Protocols/OracleDnsProtocol.cs
erikzhang: If we handle the `dns` URI scheme, we should name it `OracleDnsProtocol`.

docs/oracle-dns-protocol.md
Jim8y: now consistent to RFC4501

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,503 @@
...
+        }
+        return false;
+    }
+
+    private static CertificatePublicKey BuildPublicKey(X509Certificate2 cert)
+    {
+        CertificatePublicKey key = new()
+        {
+            Algorithm = cert.PublicKey.Oid?.FriendlyName ?? cert.PublicKey.Oid?.Value,
+            Encoded = Convert.ToBase64String(cert.GetPublicKey())
```
shargon: I think that it's better to construct this object inside the try/catch clause, it's possible to throw [CryptographicException](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.cryptographicexception?view=net-9.0) during cert.GetPublicKey()

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,503 @@
...
+            return;
+        lock (syncRoot)
+        {
+            if (configured && !force)
+                return;
+            var dnsSettings = OracleSettings.Default?.Dns ?? throw new InvalidOperationException("DNS settings are not loaded.");
+            endpoint = dnsSettings.EndPoint;
+            client.Timeout = dnsSettings.Timeout;
+            client.DefaultRequestHeaders.Accept.Clear();
+            client.DefaultRequestHeaders.Accept.ParseAdd("application/dns-json");
```
shargon: This can be moved to constructor

docs/oracle-dns-protocol.md
```diff
@@ -0,0 +1,129 @@
...
+
+## RFC 4501 URI format
+
+```
+dns:[//authority/]domain[?CLASS=class;TYPE=type][;FORMAT=x509]
+```
+
+- `domain` is the DNS owner name (relative or absolute). Percent-encoding and escaped dots (`%5c.`) follow RFC 4501 rules.
+- `domain` must not include additional path segments; only the owner name belongs here.
+- `authority` is the optional DNS server hint from RFC 4501. The oracle still uses the DoH resolver configured in `OracleService.json`.
```
erikzhang: Why not use the user-specified authority?

docs/oracle-dns-protocol.md
```diff
@@ -0,0 +1,129 @@
...
+- `name` is an oracle extension; if present, it overrides `domain` entirely (useful for percent-encoding complex owner names).
+
+Query parameters can be separated by `;` (RFC style) or `&`.
+
+Examples:
+
+- `dns:1alhai._domainkey.icloud.com?TYPE=TXT` — DKIM TXT record.
+- `dns:simon.example.org?TYPE=CERT;FORMAT=x509` — extract the X.509 payload into `Certificate`.
+- `dns://192.168.1.1/ftp.example.org?TYPE=A` — RFC-compliant authority form (authority is ignored; the configured DoH endpoint is used).
+- `dns:ignored?name=weird%5c.label.example&type=TXT` — uses the `name` override (decoded to `weird\.label.example`).
```
erikzhang: Why can a DNS name contain backslashes (`\`)?

plugins/OracleService/Protocols/OracleDnsProtocol.cs
erikzhang: I think we should use `application/dns-message`.
`application/dns-json` is not standard. The format returned by each DOH server may be different.

docs/oracle-dns-protocol.md
```diff
@@ -0,0 +1,129 @@
...
+1. Install or build the `OracleService` plugin and copy `OracleService.json` next to the plugin binary.
+2. Add the `Dns` section (defaults shown):
+
+```jsonc
+{
+  "PluginConfiguration": {
+    // ...
+    "Dns": {
+      "EndPoint": "https://cloudflare-dns.com/dns-query",
+      "Timeout": 5000
```
Wi1l-B0t, line 19: What Unit ?

Millseconds?

```suggestion
      "TimeoutMilliseconds": 5000
```

plugins/OracleService/OracleSettings.cs
```diff
@@ -36,6 +36,19 @@ public NeoFSSettings(IConfigurationSection section)
...
+class DnsSettings
+{
+    public Uri EndPoint { get; }
+    public TimeSpan Timeout { get; }
+
+    public DnsSettings(IConfigurationSection section)
+    {
+        string endpoint = section.GetValue("EndPoint", "https://cloudflare-dns.com/dns-query");
+        EndPoint = new Uri(endpoint, UriKind.Absolute);
+        Timeout = TimeSpan.FromMilliseconds(section.GetValue("Timeout", 5000));
```
Wi1l-B0t, line 48:
```suggestion
        Timeout = TimeSpan.FromMilliseconds(section.GetValue("TimeoutMilliseconds", 5000));
```

plugins/OracleService/Protocols/OracleDnsProtocol.cs
vncoelho: Indeed if we rely on the JSON-style API (application/dns-json), we would depend on a non-standard

That means if we switch DoH resolver servers (or someone configures a different provider), the JSON structure might change. But maybe it is not our case and we can adopt it

plugins/OracleService/OracleSettings.cs
```diff
@@ -36,6 +36,19 @@ public NeoFSSettings(IConfigurationSection section)
...
 }
 
+class DnsSettings
+{
+    public Uri EndPoint { get; }
+    public TimeSpan Timeout { get; }
+
+    public DnsSettings(IConfigurationSection section)
+    {
+        string endpoint = section.GetValue("EndPoint", "https://cloudflare-dns.com/dns-query");
```
vncoelho, line 46: These days ago, 18th November, cloudfare caused a massive outrage worldwide.

Isn't there a more generic way to do so?

vncoelho: ping @igormcoelho

doubiliu: The current investigation has not found it. DNSSEC is regarded as a paid item by large email service providers and is not universal. In addition, I think the impact of the previous world-class disaster was limited, because our biggest impact was the suspension of services, and the problem of financial security was not too big.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
Jim8y: dns json is a cloudflare private standard. will try to udpate.

docs/oracle-dns-protocol.md
Jim8y: is that necessary?  can be updated if you think its a right thing to do. for me, current design is sufficient.

erikzhang: Relying on a fixed DNS server is unreliable and it may fail. At that time, we must modify the configuration file before we can repair it. We can completely use the server specified by the user in the URI for querying.

Jim8y:
- ✅ application/dns-json → application/dns-message (RFC 8484)
  - ✅ HTTP GET → HTTP POST
  - ✅ JSON parsing → DNS wire format (RFC 1035)
  - ✅ Support user-specified authority (dns://resolver/domain)
  - ✅ Fix CryptographicException handling in BuildPublicKey
  - ✅ Move Accept header to constructor
  - ✅ Integration tests covering Cloudflare, Google, Quad9

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,850 @@
...
+            endpoint = dnsSettings.EndPoint;
+            client.Timeout = dnsSettings.Timeout;
+            configured = true;
+        }
+    }
+
+    internal static string BuildQueryName(Uri uri, NameValueCollection queryParameters = null)
+    {
+        queryParameters ??= ParseQueryString(uri.Query);
+        string overriddenName = NormalizeLabel(GetQueryValue(queryParameters, "name"));
```
erikzhang: There is no `name` in the query string.

cschuchardt88 (dismissed):

plugins/OracleService/OracleService.json
```diff
@@ -14,6 +14,10 @@
       "EndPoint": "http://127.0.0.1:8080",
       "Timeout": 15000
     },
+    "Dns": {
+      "EndPoint": "https://cloudflare-dns.com/dns-query",
+      "TimeoutMilliseconds": 5000
```
erikzhang: L15 and L19, one is `Timeout` and the other is `TimeoutMilliseconds`. It's best to keep them consistent.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -250,14 +221,45 @@ private async Task<DnsMessage> ResolveAsync(string name, ushort type, Uri resolv
...
+
         using HttpResponseMessage response = await client.PostAsync(resolverEndpoint, content, cancellation);
         if (!response.IsSuccessStatusCode)
             throw new InvalidOperationException($"DoH endpoint returned {(int)response.StatusCode} ({response.StatusCode})");
 
-        byte[] responseData = await response.Content.ReadAsByteArrayAsync(cancellation);
+        if (response.Content.Headers.ContentLength.HasValue && response.Content.Headers.ContentLength > OracleResponse.MaxResultSize)
+            throw new ResponseTooLargeException();
+
+        using Stream stream = await response.Content.ReadAsStreamAsync(cancellation);
```
cschuchardt88: `IO Pipelines` would work better here, better buffering and memory management.

https://learn.microsoft.com/en-us/dotnet/standard/io/pipelines

plugins/OracleService/OracleService.cs
```diff
@@ -528,6 +553,35 @@ public static byte[] Filter(string input, string filterArgs)
...
+            return string.IsNullOrEmpty(data) ? "<stackitem:empty>" : $"<stackitem:base64:{data.Length} chars>";
+
+        if (string.IsNullOrEmpty(data))
+            return data;
+
+        const int maxLen = 2048;
+        return data.Length <= maxLen ? data : data[..maxLen] + "...";
+    }
+
+    internal static bool TryDecodeDnsStackItemPayload(string payload, out byte[] result)
```
cschuchardt88:
```suggestion
    internal static bool TryDecodeDnsStackItemPayload(string payload, [NotNullWhen(true)] out byte[] result)
```

plugins/OracleService/OracleService.json
cschuchardt88: Could use backups like `google` https://dns.google/resolve?name=google.com&type=A

plugins/OracleService/OracleService.cs
shargon:
```suggestion
    internal static bool TryDecodeDnsStackItemPayload(string payload, [NotNullWhen(true)] out byte[]? result)
```

plugins/OracleService/Protocols/OracleDnsProtocol.cs
Jim8y: This path no longer constructs an X509Certificate or calls cert.GetPublicKey(). CERT RDATA is returned as raw RFC 4398 data, so public key extraction cannot throw here anymore.

plugins/OracleService/OracleService.json
Jim8y: The configuration now uses Timeout consistently for Https, NeoFS, and Dns, and DnsSettings reads Timeout directly.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
Jim8y: Updated the response reader to use PipeReader while preserving the OracleResponse.MaxResultSize bound.

tests/Neo.Plugins.OracleService.Tests/UT_OracleDnsProtocol.cs
```diff
@@ -0,0 +1,1201 @@
...
+    [TestMethod]
+    public void BuildQueryName_ThrowsWithoutDnsName()
+    {
+        var uri = new Uri("dns://resolver.example/");
+        try
+        {
+            OracleDnsProtocol.BuildQueryName(uri);
+            Assert.Fail("Expected FormatException for missing dnsname.");
+        }
+        catch (FormatException)
```
shargon: Could you use `Assert.ThrowsExactly`? this pattern it's dirty

plugins/OracleService/OracleService.cs
```diff
@@ -542,6 +569,41 @@ public static byte[] Filter(string input, string? filterArgs)
...
+            result = null;
+            return false;
+        }
+
+        try
+        {
+            result = Convert.FromBase64String(payload);
+            return true;
+        }
+        catch (Exception ex) when (ex is ArgumentNullException or FormatException)
```
shargon:
```suggestion
        catch (Exception ex)
```
Avoid unexpected errors that can deal in DoS

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,711 @@
...
+            return (OracleResponseCode.ResponseTooLarge, null);
+        }
+        catch (Exception ex)
+        {
+            return (OracleResponseCode.Error, ex.Message);
+        }
+
+        if (dnsResponse is null)
+            return (OracleResponseCode.Error, "Invalid DNS response.");
+
```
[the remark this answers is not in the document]
Jim8y: Handled in 09d4a4b. The unreachable null response branch was removed.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,711 @@
...
+        foreach (string label in labels)
+        {
+            if (label.Length > 63)
+                throw new FormatException($"DNS label exceeds 63 characters: {label}");
+            byte[] labelBytes = Encoding.ASCII.GetBytes(label);
+            buffer.Add((byte)labelBytes.Length);
+            buffer.AddRange(labelBytes);
+        }
+        buffer.Add(0x00); // Root label
+    }
```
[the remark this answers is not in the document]
Jim8y: Handled in 09d4a4b. DNS name encoding now enforces the 255-octet wire-format limit and has coverage for the maximum valid name and an overlong name.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,711 @@
...
+            return false;
+
+        if (host.Equals("localhost", StringComparison.OrdinalIgnoreCase))
+            return true;
+
+        if (IPAddress.TryParse(host, out IPAddress? address))
+            return address.IsInternal();
+
+        return false;
+    }
```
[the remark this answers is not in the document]
Jim8y: Handled in 09d4a4b. Localhost resolver suffixes are rejected before resolver lookup, with coverage for sub.localhost.

plugins/OracleService/OracleService.cs
```diff
@@ -301,8 +303,20 @@ private async Task ProcessRequestAsync(DataCache snapshot, OracleRequest req)
...
+
+        byte[]? dnsStackBytes = null;
+        if (code == OracleResponseCode.Success && dnsStackOutput)
+        {
+            if (!TryDecodeDnsStackItemPayload(data, out dnsStackBytes))
+            {
+                code = OracleResponseCode.Error;
+                PluginLogger?.Warning("Invalid DNS stack item payload: {OriginalTxid}", req.OriginalTxid);
+            }
+        }
```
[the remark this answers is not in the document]
Jim8y: Handled in 09d4a4b. The final response log is emitted after DNS stack payload validation, so the logged response code matches any Error conversion from decode failure.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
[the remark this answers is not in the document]
Jim8y: Follow-up in 4c4583f. The default DoH HTTP handler now resolves the resolver host during connection setup, rejects the connection if any resolved address is private, and opens the socket directly to the validated address instead of letting HttpClient resolve it again. I also added coverage for mixed public/private resolved address lists.

Jim8y:
Pushed a follow-up for deterministic DNS oracle results and stricter response parsing.

Changes:
- validate DNS response ID and question section
- sort answers by name/type/data
- normalize TTL to 0
- reject truncated TXT records
- reject forward/self compression pointers
- use per-request timeout instead of resetting HttpClient.Timeout

Validation:
- dotnet test tests/Neo.Plugins.OracleService.Tests/Neo.Plugins.OracleService.Tests.csproj --no-restore --filter UT_OracleDnsProtocol --verbosity minimal
- dotnet format tests/Neo.Plugins.OracleService.Tests/Neo.Plugins.OracleService.Tests.csproj --no-restore --verify-no-changes --verbosity minimal

cschuchardt88:
can we add in this PR a way to resolve `Neo Named Service` aka `NNS`?

Related https://github.com/neo-project/proposals/pull/239

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+
+    /// <summary>
+    /// Formats RDATA based on record type for human-readable output.
+    /// </summary>
+    private static string FormatRData(ushort type, byte[] rdata)
+    {
+        if (rdata is null || rdata.Length == 0)
+            return string.Empty;
+
+        return type switch
```
cschuchardt88: **[bug]** `FormatRData` only special-cases A/AAAA/TXT/CERT. CNAME, NS, MX, SRV, and SOA (all in `RecordTypeLookup`) fall through to `Convert.ToBase64String(rdata)` of the raw wire bytes. Those RDATA fields contain domain names that almost always use RFC 1035 compression pointers (`0xC0 0x0C` pointing at the question name is typical). The oracle payload then contains the pointer bytes, not the hostname, so a CNAME answer is useless to contracts. Two resolvers (or the same resolver with/without compression) serialize different bytes for the same record, so oracle nodes will not agree. `SkipDnsName` / `DecodeDnsName` already exist but are not applied to RDATA.

**Suggestion:** Expand name-valued RDATA with `DecodeDnsName` (and parse MX/SRV/SOA fields) into a stable text form, or drop those types until that is implemented. Never copy compressed wire bytes into the consensus payload.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+        if (dnsResponse.ResponseCode != 0)
+            return (OracleResponseCode.Error, $"DNS error (RCODE {dnsResponse.ResponseCode})");
+
+        if (dnsResponse.Answers.Count == 0)
+            return (OracleResponseCode.NotFound, null);
+
+        ResultAnswer[] answers;
+        try
+        {
+            answers = dnsResponse.Answers
```
cschuchardt88: **[bug]** Every resource record in the DNS answer section is serialized, not just the requested type, and owner names keep wire-format case (`a.Name.TrimEnd('.')` at line 189). Sorting is case-insensitive, so order can match while the ByteStrings differ (`WWW.Example.COM` vs `www.example.com`). CNAME chains make this worse: a TYPE=A lookup often returns CNAME+A, while another DoH server returns only A. Docs claim answers are normalized so all oracle nodes serialize the same payload; TTL=0 plus sort is not enough. Disagreeing payloads fail the request with `ConsensusUnreachable` rather than forking, but `dns:` will be unreliable in production.

**Suggestion:** Keep only records whose type equals the query type (or document and implement a fixed CNAME-follow policy). Lowercase names before sort and serialize. Deduplicate `(Name, Type, Data)`. Document that `dns:` URLs without an authority require every oracle operator to use the same `Dns.EndPoint`.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+    /// </summary>
+    private static void EncodeDnsName(List<byte> buffer, string name)
+    {
+        if (string.IsNullOrEmpty(name) || name == ".")
+        {
+            buffer.Add(0x00);
+            return;
+        }
+
+        string[] labels = name.TrimEnd('.').Split('.');
```
cschuchardt88: **[bug]** RFC 4501 (and this PR’s docs) encode a dot inside a label as `\.`, written in the URI as `%5c.`. `NormalizeLabel` unescapes `%5c` to `\`, then `EncodeDnsName` does `name.TrimEnd('.').Split('.')`, so `exa%5c.mple.com` becomes labels `exa\` and `mple` instead of one label `exa.mple`. RFC 1035 `\DDD` octal escapes are also ignored. The documented example form does not round-trip.

**Suggestion:** Split labels on unescaped dots only, and interpret `\.` / `\DDD` per RFC 1035 master-file escaping before encoding the query.

plugins/OracleService/OracleService.cs
```diff
@@ -312,7 +326,20 @@ private async Task ProcessRequestAsync(DataCache snapshot, OracleRequest req)
             {
                 try
                 {
-                    result = Filter(data!, request.Filter);
+                    if (dnsStackOutput)
+                    {
+                        if (!string.IsNullOrEmpty(request.Filter))
```
cschuchardt88: **[bug]** `ProcessRequestAsync` loops all oracle requests that share a URL and the catch sets the shared `code` on filter failure. Any non-empty filter on a `dns:` request throws (`Filter is not supported for dns: requests.`). If one contract uses a filter and another uses the same URL with an empty filter, iteration order decides whether the valid request is also answered as `Error`. HTTPS had a similar catch, but DNS now throws on every non-empty filter, so poisoning is the common case rather than a bad JSONPath.

**Suggestion:** Use a per-iteration `responseCode` (and result). Do not assign `code = OracleResponseCode.Error` for a sibling request. Rejecting a DNS filter should not change other requests for that URL.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+
+        if (addresses.Length == 0)
+            throw new InvalidOperationException("Failed to resolve resolver endpoint.");
+
+        return ValidateEndpointAddresses(addresses);
+    }
+
+    internal static IPAddress[] ValidateEndpointAddresses(IPAddress[] addresses)
+    {
+        if (!OracleSettings.Default.AllowPrivateHost && addresses.Any(p => p.IsInternal()))
```
cschuchardt88: **[bug]** User-specified `dns://authority/` makes every oracle node POST to `https://{authority}/dns-query`. Private-host checks call `IPAddress.IsInternal()`, which only treats 10/8, 127/8, 172.16/12, 192.168/16, loopback, and unspecified as internal. Link-local `169.254.0.0/16` (cloud metadata), CGNAT `100.64.0.0/10`, IPv6 ULA `fc00::/7`, and IPv6 link-local `fe80::/10` pass. `dns://169.254.169.254/example.com?TYPE=A` therefore reaches `https://169.254.169.254/dns-query`. HTTPS has the same helper gap; this PR reuses it in a new user-controlled resolver path and ConnectCallback. HTTPS maps private hosts to `Forbidden`; this path throws and becomes `Error`.

**Suggestion:** Treat 169.254/16, 100.64/10, IPv6 ULA, and IPv6 link-local as internal in the resolver allow-check (or extend `Helper.IsInternal` and use it here). Return `OracleResponseCode.Forbidden` like `OracleHttpsProtocol`. Keep blocking mixed public+private A/AAAA sets.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+
+        string[] labels = name.TrimEnd('.').Split('.');
+        int wireLength = 1; // Root label.
+        foreach (string label in labels)
+        {
+            if (label.Length == 0)
+                throw new FormatException("DNS name contains an empty label.");
+            if (label.Length > 63)
+                throw new FormatException($"DNS label exceeds 63 characters: {label}");
+            byte[] labelBytes = Encoding.ASCII.GetBytes(label);
```
cschuchardt88: **[suggestion]** `Encoding.ASCII.GetBytes(label)` uses the default replacement fallback, so a percent-decoded IDN label (`exämple`) is queried as `ex?mple` with no error. RFC 4501 treats `dnsname` as an IDN-unaware slot; non-ASCII should be rejected (or converted to punycode), not silently corrupted.

**Suggestion:** Use `EncoderExceptionFallback` or reject any label with non-ASCII octets. If IDN is desired, apply IDNA ToASCII first.

plugins/OracleService/OracleSettings.cs
```diff
@@ -36,6 +36,19 @@ public NeoFSSettings(IConfigurationSection section)
...
+        EndPoint = new Uri(endpoint, UriKind.Absolute);
```
cschuchardt88: **[suggestion]** `DnsSettings` accepts any absolute URI as `EndPoint` (http, ftp, custom path). Authority-form URLs are forced to `https` + `/dns-query`, but the configured fallback is not. Plaintext DoH and a per-node Cloudflare default both undermine the “identical results” goal when the URI omits authority. Geo-DNS A records will still differ even with one provider.

**Suggestion:** Require `https` for `Dns.EndPoint`. Document that operators must share the same endpoint, or prefer authority-form `dns://` URLs for anything that must reach consensus.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+        return false;
+    }
+
+    /// <summary>
+    /// Gets the DoH resolver endpoint from the URI authority or falls back to the configured default.
+    /// Per RFC 4501, the authority component specifies the DNS server to query.
+    /// </summary>
+    private Uri GetResolverEndpoint(Uri uri)
+    {
+        // Check if URI has an authority (e.g., dns://resolver.example.com/domain)
```
cschuchardt88: **[suggestion]** Several comments restate the next line (`Check if URI has an authority`, `Fall back to configured endpoint`, header/flag byte comments in `BuildDnsQuery`, `Parse answer section`). `SkipDnsName` is unused dead code left from an earlier parser. Comments should explain non-obvious constraints (why redirects are disabled, why TTL is zeroed), not narrate the wire layout.

**Suggestion:** Delete `SkipDnsName` and the WHAT comments. Keep a short note only where the RFC constraint is not obvious from the code.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
cschuchardt88, line 658: **Bug:** `CNAME`/`NS`/`MX`/`SOA`/`SRV` (all in `RecordTypeLookup`) fall through to `Convert.ToBase64String(rdata)`. Those RDATA fields are domain names and may contain RFC 1035 compression pointers relative to the message, so the bytes are not a hostname and are layout-dependent.

This also hits default `TYPE=A` lookups: `Answers` is not filtered by query type, and CNAME+A is a normal answer section. Two nodes (or two encodings of the same names) will serialize different stack items and fail oracle consensus; contracts also cannot use the pointer bytes.

Expand compressed names in RDATA before formatting (including MX/SOA/SRV prefixes). Consider emitting only records whose type matches the query, or define CNAME-chain behavior explicitly.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+        if (dnsResponse.Answers.Count == 0)
+            return (OracleResponseCode.NotFound, null);
+
+        ResultAnswer[] answers;
+        try
+        {
+            answers = dnsResponse.Answers
+                .Select(a => new ResultAnswer
+                {
+                    Name = a.Name.TrimEnd('.'),
```
cschuchardt88, line 189: **Bug:** Owner names are only `TrimEnd('.')`. DNS is case-insensitive, but the serialized `ByteString` keeps whatever case the resolver echoed (`Example.COM` vs `example.com`). Sorting is `OrdinalIgnoreCase`, so order can match while the payload bytes do not — consensus split.

Normalize with `ToLowerInvariant()` here (and for `queryName` if `Uri` can preserve case).

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+            };
+
+            if (uri.Port > 0 && uri.Port != 443)
+                builder.Port = uri.Port;
+
+            return builder.Uri;
+        }
+
+        // Fall back to configured endpoint
+        return endpoint ?? throw new InvalidOperationException("DNS settings are not loaded.");
```
cschuchardt88, line 417: **Bug:** Fallback to the node-local `Dns.EndPoint` means a `dns:` URL without authority is not a complete oracle request. All CN nodes share the on-chain URL, not `OracleService.json`. Cloudflare vs Google vs Quad9 (all documented, plus “run your own gateway”) will disagree on A/AAAA sets and sometimes on CNAME/TXT presentation.

The contract sample is `dns:1alhai._domainkey.icloud.com?TYPE=TXT` (no authority). Require `dns://authority/...` so the resolver is part of the request, or require every oracle operator to use the same EndPoint and document that anything else fails consensus.

plugins/OracleService/OracleService.cs
cschuchardt88, line 331: **Bug:** Any non-empty `request.Filter` throws here. The surrounding `catch` assigns `code = OracleResponseCode.Error` on the *shared* loop variable. `GetRequestsByUrl` can return multiple requests; if a filtered one is processed first, later requests for the same URL (including empty filter) are stored as Error too.

Use a per-request `responseCode` and do not write it back to `code`. Filter is per-request; the fetched DNS payload is not.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+        if (normalized.Contains('/'))
+            throw new FormatException("dnsname must not contain path segments.");
+        return normalized;
+    }
+
+    private static string? NormalizeLabel(string? value)
+    {
+        if (string.IsNullOrWhiteSpace(value))
+            return null;
+        return Uri.UnescapeDataString(value).Trim().Trim('.');
```
cschuchardt88, line 781: **Bug:** `BuildQueryName` already uses `UriFormat.Unescaped`, then this unescapes again. `dns:example%252ecom` becomes `example.com` instead of a label containing `%2e`.

Docs also claim RFC 4501 escaped dots (`%5c.`). After unescape you get `foo\.bar.example.com`, and `Split('.')` still splits on the embedded dot, so `foo.bar` is not one label.

Unescape once; if RFC 4501 is the spec, treat `\.` as a literal dot inside a label (or drop that claim from the docs).

plugins/OracleService/Protocols/OracleDnsProtocol.cs
cschuchardt88, line 479: **Bug:** `Encoding.ASCII.GetBytes` uses the replacement fallback, so `dns:münchen.example?TYPE=A` is queried as `m?nchen.example` and can still return `Success`.

Reject labels with non-ASCII octets (or IDNA-map to punycode) before encoding. `EncoderExceptionFallback` / a UTF-8 round-trip check would fail closed.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+        while (offset < rdata.Length)
+        {
+            int length = rdata[offset++];
+            if (offset + length > rdata.Length)
+                throw new FormatException("TXT record string is truncated.");
+
+            if (result.Length > 0)
+                result.Append(' ');
+
+            result.Append('"');
```
cschuchardt88, line 685: **Suggestion:** This is dig presentation format (`"chunk1" "chunk2"`). Character-string boundaries are not semantically stable across nameservers, and the contract sample stores `Answer[3]` as the DKIM value *including quotes*.

Concatenate the UTF-8 chunks (what SPF/DKIM expect) and skip the quotes, or document the presentation form and strip in the sample.

plugins/OracleService/Protocols/OracleDnsProtocol.cs
```diff
@@ -0,0 +1,807 @@
...
+    /// </summary>
+    private static DnsMessage ParseDnsResponse(byte[] data, ushort expectedId, string expectedName, ushort expectedType)
+    {
+        if (data is null || data.Length < DnsHeaderSize)
+            throw new FormatException("DNS response too short.");
+
+        DnsMessage message = new()
+        {
+            Id = BinaryPrimitives.ReadUInt16BigEndian(data.AsSpan(0, 2)),
+            Flags = BinaryPrimitives.ReadUInt16BigEndian(data.AsSpan(2, 2))
```
cschuchardt88, line 500: **Suggestion:** Only `Flags & 0x0F` (RCODE) is consulted. A TC=1 response can still be `Success` with a partial answer set — another consensus split. QR=0 (a query, not a response) and a non-zero opcode would also be accepted.

Reject QR=0, opcode ≠ 0, and TC=1 before parsing answers.

cschuchardt88 (commented):
## Review (DNS oracle protocol)

Deterministic-result follow-ups are in place (TTL=0, sorted answers, query ID/question match, per-request timeout, backward-only compression pointers). Several consensus/correctness gaps remain before this is safe to enable on CN oracles.

**Bugs**
- Name-typed RDATA (CNAME/NS/MX/SOA/SRV) is base64 of possibly *compressed* wire bytes; A responses often include a CNAME, so this affects default lookups too.
- Answer owner names keep resolver case; serialized payload can differ across nodes.
- `dns:` without authority uses each node's `Dns.EndPoint` — the on-chain URL is then not a complete request. The docs example is this form.
- A non-empty filter on one same-URL request sets shared `code = Error` and poisons later requests in `GetRequestsByUrl`.
- DNS name is unescaped twice; RFC 4501 `\. ` / `%5c.` is documented but not implemented.
- Non-ASCII labels are silently encoded as `?`.

**Suggestions**
- Concatenate TXT character-strings (no quotes) for stable, contract-friendly data.
- Reject TC=1 / QR=0 DNS messages; require `Content-Type: application/dns-message`.
- Do not require HTTP/2-only; map private resolvers to `Forbidden` like HTTPS.
- Stubbed unit tests still call `Dns.GetHostAddressesAsync` via `EnsureEndpointAllowed`.

PR #920, opened 2025-11-24
Remove transaction if gas is higer than max

shargon:
Related with https://github.com/neo-project/neo/issues/4333

This pull request introduces a minor update to the consensus logic in the `DBFTPlugin`, focusing on transaction validation and error handling. The main change ensures that transactions exceeding the maximum allowed system fee are rejected, and it adds a placeholder for future handling of dropped transactions.

Transaction validation improvements:

* Added a check in `AddTransaction` to reject transactions whose `SystemFee` exceeds `dbftSettings.MaxBlockSystemFee`, preventing high-fee transactions from being added to the consensus context.
* Updated `OnTransaction` to handle cases where `AddTransaction` fails, with a TODO comment for dropping invalid transactions in the future.


vncoelho (commented):
lgtm but needs on master 3x as well

ajara87 (dismissed):

plugins/DBFTPlugin/Consensus/ConsensusService.cs
```diff
@@ -249,13 +249,19 @@ private void OnTransaction(Transaction transaction)
...
+            // TODO: Drop the transaction if it cannot be added to the context
+        }
     }
 
     private bool AddTransaction(Transaction tx, bool verify)
     {
         if (verify)
         {
+            // Must not be more than MaxBlockSystemFee
+            if (tx.SystemFee > dbftSettings.MaxBlockSystemFee) return false;
```
erikzhang, line 263: Maybe we can just remove `MaxBlockSystemFee`.

AnnaShaleva, line 263: We need to request CV in this case and also need a proper log message for it. The primary who proposed this block should be considered as malicious.

shargon: Could deny the service with too much gas

erikzhang: Perhaps limiting through system fees isn't a good approach. Consider that we already have `RegisterCandidate`; in the future, we may have even more expensive methods.

vncoelho: If someone is paying GAS, @shargon, I think that there is no problem.

From my point of view we could have a limit in processing time for the transaction itself, maybe 30 seconds for each transaction and if passed we burn `X` amount. In my vision, more than "30s or something like that" would be a loop.

roman-khimov: > a limit in processing time for the transaction itself,

That's exactly what system fee is, a limit of resources available for execution. And in reality
1. `MaxBlockSystemFee` must be present for network safety. 1000 GAS can stall the network for a long long time.
2. It should be limited to 20-30 GAS with the original ExecFeeFactor of 30, see https://github.com/neo-project/neo/issues/3552.

erikzhang: System fee is not equivalent to CPU time; it also includes storage and other service costs.

roman-khimov: Practically it dominates the price of real transactions and that's where most of abuse potential comes from as well.

vncoelho: > System fee is not equivalent to CPU time; it also includes storage and other service costs.

Indeed there is Storage.
Storage is where we can mostly improve a lot.

We need more types of storages, we need temporary Storage (at least 2 types), permanent storage, consensus layer Storage.
We also need something @igormcoelho will soon describe for portable Storage.

vncoelho: > System fee is not equivalent to CPU time; it also includes storage and other service costs.

Indeed there is Storage.
Storage is where we can mostly improve a lot.

We need more types of storages, we need temporary Storage (at least 2 types), permanent storage, consensus layer Storage.
We also need something @igormcoelho will soon describe for portable Storage.

shargon: Or wrong configured

shargon:
Replaced by https://github.com/neo-project/neo/pull/4362

PR #924, opened 2025-11-26
port message sign to master-n3

ajara87:
Porting https://github.com/neo-project/neo/pull/4286 of @adrian-fjellberg
Proposal https://github.com/neo-project/proposals/pull/213

- Added use of GetSignData [Opcional]
- Added UTs

  - TestOnSignMessageCommand
  - TestOnSignMessageCommandWithoutPassword
  - TestOnSignMessageCommandWrongPassword
  - TestOnSignMessageCommandWithoutAccount
  - TestOnVerifyMessageCommand
  - TestOnSignWithSignatureReplayOnVerifyWithoutSignatureReplay
  - TestOnVerifyMessageCommand

### DIFF
--- a/src/Neo.CLI/CLI/MainService.Wallet.cs
+++ b/src/Neo.CLI/CLI/MainService.Wallet.cs
@@ -583,6 +583,8 @@ partial class MainService
 
         foreach (WalletAccount account in CurrentWallet.GetAccounts().Where(p => p.HasKey))
         {
+            if (account.Contract == null || IsMultiSigContract(account.Contract.Script))
+                continue;
             var key = account.GetKey();
             var signature = Crypto.Sign(signData, key!.PrivateKey, ECCurve.Secp256r1);
 


Jim8y:
> Need https://github.com/neo-project/neo-node/pull/926 for UT 

@ajara87 926 is merged

ajara87:
> > Need #926 for UT
> 
> @ajara87 926 is merged

thank you @Jim8y. I have to upload a proposal and the PR is ready to review.

cschuchardt88 (approved):

shargon:
@cschuchardt88 It's a draft

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +507,77 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+        {
+            ConsoleHelper.Info("Cancelled");
+            return;
+        }
+        if (!CurrentWallet!.VerifyPassword(password))
+        {
+            ConsoleHelper.Error("Incorrect password");
+            return;
+        }
+
```
superboyiii, line 534: I suggest adding this.
```suggestion

        if (message.Length >= 2)
        {
            if ((message[0] == '"' && message[^1] == '"') || (message[0] == '\'' && message[^1] == '\''))
            {
                message = message[1..^1];
            }
        }

```

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +507,86 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+        {
+            ConsoleHelper.Error("Null message");
+            return;
+        }
+
+        var saltBytes = new byte[16];
+        RandomNumberGenerator.Fill(saltBytes);
+        var saltHex = saltBytes.ToHexString().ToLowerInvariant();
+
+        var paramBytes = Encoding.UTF8.GetBytes(saltHex + message);
```
shargon, line 545: I think we should remove this salt. Perhaps we should have an integer (4 bytes) for `ValidUntilBlock`, but I don't see why a salt would be useful here

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +507,86 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+
+        byte[] payload;
+        using (var ms = new MemoryStream())
+        using (var w = new BinaryWriter(ms, Encoding.UTF8, true))
+        {
+            // We add these 4 bytes to prevent the signature from being a valid transaction
+            w.Write((byte)0x01);
```
shargon, line 552: Opcode.RET ?

roman-khimov: This sequence can't be changed.

shargon: Why?

roman-khimov: Same reason as in https://github.com/neo-project/proposals/pull/213#issuecomment-3697518925.

shargon: Something that is wrong must not be used, this version has a replay signature attack, so if other people in the past used a vulnerable "standard" is not a reason for do it official, if they want to use this way, are free to use this version, but the standard must be secure

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -23,7 +25,9 @@
 using Neo.Wallets.NEP6;
 using System.Numerics;
 using System.Security.Cryptography;
+using System.Text;
 using static Neo.SmartContract.Helper;
+using ECCurve = Neo.Cryptography.ECC.ECCurve;
```
cschuchardt88, line 31:
```suggestion
using ECCurve = Neo.Cryptography.ECC.ECCurve;
using Neo.Factories;
```

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +507,86 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+        }
+
+        if (message == null)
+        {
+            ConsoleHelper.Error("Null message");
+            return;
+        }
+
+        var saltBytes = new byte[16];
+        RandomNumberGenerator.Fill(saltBytes);
```
cschuchardt88:
```suggestion
        saltBytes = RandomNumberFactory.NextBytes(saltBytes.length, cryptography: true);
```

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +507,86 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+        var saltHex = saltBytes.ToHexString().ToLowerInvariant();
```
cschuchardt88:
```suggestion
        var saltHex = Convert.ToHexStringLower(saltBytes);
```

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +508,189 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+            if (string.IsNullOrEmpty(salt))
+            {
+                ConsoleHelper.Error("Salt cannot be empty");
+                return;
+            }
+
+            // Reconstruct payload: 010001f0 + VarBytes(Salt + Message) + 0000
+            // Note: salt is used as hex string (lowercase), same as in signing
+            var saltHex = salt.ToLowerInvariant();
+            var paramBytes = Encoding.UTF8.GetBytes(saltHex + message);
```
shargon, line 642: The message can contain the salt. I think we can remove the salt, and recommend to add it in the NEP

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +508,189 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+    /// <param name="message">Original message that was signed</param>
+    /// <param name="signature">Signature in hex format</param>
+    /// <param name="publicKey">Public key in hex format</param>
+    /// <param name="salt">Salt in hex format</param>
+    [ConsoleCommand("verify message", Category = "Wallet Commands")]
+    private void OnVerifyMessageCommand(string message, string signature, string publicKey, string salt)
+    {
+        try
+        {
+            if (message.Length >= 2)
```
shargon: The method `NormalizeMessage` can be used here

cschuchardt88 (dismissed):

Jim8y:
@shargon please add this one to 3.9.

Jim8y (dismissed):

ajara87:
> @ajara87 We need rpc `signmsg` and `verifymsg` in RpcServer as well. But can be made in another PR.

Sure! I'll do It in another PR

src/Neo.CLI/CLI/MainService.Wallet.cs
shargon: Ping @ajara87

src/Neo.CLI/CLI/MainService.Wallet.cs
shargon: @roman-khimov what do you think?

erikzhang: What is `010001f0`?

roman-khimov: > @roman-khimov what do you think?

Same as previous, this code just does what wallets were doing for years. They salted the message, good for them, OK for me.

> What is `010001f0`?

Magic. Chosen _years_ ago to ensure that signed message is never a transaction. Still works for N3 since we don't have transaction version 1 (yet).

Refs:
 * https://github.com/CityOfZion/neon-wallet/blob/bdfcbbf4e167d3d1987ec1d1a0cdc658e4bb8bf5/app/util/N3Helper.js#L196
 * https://github.com/neo-project/proposals/pull/69/files
 * https://docs.switcheo.network/#signing-messages-for-neo

erikzhang: Compared to salt, I prefer adding a timestamp. Additionally, the timestamp, message content, etc., can be encapsulated into structured data.

Jim8y: Maybe we should not bother it so much~~~ as long as it works.

shargon: @roman-khimov the thing is that we added `SignData` call, this add a signature protection that makes incompatible with the previous versions

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +508,183 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+                w.Write((byte)0xF0);
+                w.WriteVarBytes(paramBytes);
+                w.Write((ushort)0);
+                w.Flush();
+                payload = ms.ToArray();
+            }
+
+            // Calculate signData: SHA256(network || Hash256(payload))
+            var hash = new UInt256(Crypto.Hash256(payload));
+            var signData = hash.GetSignData(NeoSystem.Settings.Network);
```
shargon: This already makes incompatible with the wallets version. I don't see that these wallets or providers have previously tried to make their "standard" official (by nep o github issue), so I don't understand why we have to respect a project's own version

src/Neo.CLI/CLI/MainService.Wallet.cs
roman-khimov: But who needs an incompatible version of this thing? As noticed previously, even NeoFS is compatible with this scheme:
https://github.com/nspcc-dev/neofs-sdk-go/blob/23f5f0a8f3de952108d36cdbe589d0a2910ce8be/crypto/ecdsa/wallet_connect.go#L108

because when we played with gateways/wallets it was quickly discovered that they do not allow signing arbitrary data, they can sign messages in this format. So we adapted to it with this additional signature scheme. I don't see why node wallet can't do the same. There is nothing inherently wrong here, mostly it depends on what is signed and how the message signed is used, but that's not a wallet responsibility.

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +508,187 @@ private void OnSignCommand(JObject jsonObjectToSign)
         }
     }
 
+    /// <summary>
+    /// Process "sign message" command
+    /// </summary>
+    /// <param name="message">Message to sign</param>
+    [ConsoleCommand("sign message", Category = "Wallet Commands")]
+    private void OnSignMessageCommand(string message, bool addSignData)
```
shargon: avoidSignatureReplay?

ajara87: done

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +508,197 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+            w.Write((ushort)0);
+            w.Flush();
+            payload = ms.ToArray();
+        }
+
+        ConsoleHelper.Info("Signed Payload: ", $"{payload.ToHexString()}");
+        Console.WriteLine();
+        ConsoleHelper.Info("    Curve: ", "secp256r1");
+        ConsoleHelper.Info("Algorithm: ", "payload = 010001f0 + VarBytes(Salt + Message) + 0000");
+        ConsoleHelper.Info("Algorithm: ", "Sign(SHA256(network || Hash256(payload)))");
```
superboyiii: Different signing way, different info.
```suggestion
        ConsoleHelper.Info("Algorithm: ", avoidSignatureReplay
            ? "Sign(SHA256(network || Hash256(payload)))"
            : "Sign(payload)");
```

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +508,197 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+                w.Write((byte)0x00);
+                w.Write((byte)0x01);
+                w.Write((byte)0xF0);
+                w.WriteVarBytes(paramBytes);
+                w.Write((ushort)0);
+                w.Flush();
+                payload = ms.ToArray();
+            }
+
+            // Calculate signData: SHA256(network || Hash256(payload))
```
superboyiii:
```suggestion
            // signData = payload, or SHA256(network || Hash256(payload)) depends on avoidSignatureReplay
```

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -470,8 +475,8 @@ private void OnListKeyCommand()
     /// <summary>
     /// Process "sign" command
     /// </summary>
-    /// <param name="jsonObjectToSign">Json object to sign</param>
-    [ConsoleCommand("sign", Category = "Wallet Commands")]
+    /// <param name="jsonObjectToSign">The json string that records the transaction information</param>
+    [ConsoleCommand("sign transaction", Category = "Wallet Commands")]
```
superboyiii: We can still keep `sign` in compatible way since `sign message doesn't make any conflicts`
```suggestion
    [ConsoleCommand("sign", Category = "Wallet Commands")]
```

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -503,6 +508,197 @@ private void OnSignCommand(JObject jsonObjectToSign)
...
+        Console.WriteLine();
+
+        var signData = payload;
+        if (avoidSignatureReplay)
+        {
+            var hash = new UInt256(Crypto.Hash256(payload));
+            signData = hash.GetSignData(NeoSystem.Settings.Network);
+        }
+
+        foreach (WalletAccount account in CurrentWallet.GetAccounts().Where(p => p.HasKey))
```
superboyiii, line 584: I'm not sure how can we handle this. But it's apparently that we shouldn't allow these multi-sig address(not enough private keys in the same wallet) to sign. The result will not be correct. But if we want to handle it correctly, we need to give a way to show incomplete result and allow other addresses in that multi-sig address to sign. (Like what we did in mulsig signs transactions). `Or we can ban multi-sig addresses temporarily. They're not in high priority`.

ajara87: done

PR #950, opened 2025-12-20
Fix: unify Grpc dependency

Wi1l-B0t:
There are different Grpc dependency versions.

So add this to unify Grpc dependency.


Jim8y:
For me, this pr achieves what it claims to, and functionally correct. if any further polishment alike can be applied should be applied in following works. No blocking issue identified.

PR #954, opened 2025-12-26
Copy plugin dependencies for isolated loading

Jim8y:
## Summary
- copy plugin dependencies to output for per-plugin AssemblyLoadContext loading

## Testing
- not run (not requested)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -122,7 +122,7 @@ dotnet_naming_symbols.local_functions.applicable_kinds = local_function
 
 dotnet_naming_style.local_function_style.capitalization = pascal_case
 
-file_header_template = Copyright (C) 2015-2025 The Neo Project.\n\n{fileName} file belongs to the neo project and is free\nsoftware distributed under the MIT software license, see the\naccompanying file LICENSE in the main directory of the\nrepository or http://www.opensource.org/licenses/mit-license.php\nfor more details.\n\nRedistribution and use in source and binary forms with or without\nmodifications are permitted.
+file_header_template = Copyright (C) 2015-2026 The Neo Project.\n\n{fileName} file belongs to the neo project and is free\nsoftware distributed under the MIT software license, see the\naccompanying file LICENSE in the main directory of the\nrepository or http://www.opensource.org/licenses/mit-license.php\nfor more details.\n\nRedistribution and use in source and binary forms with or without\nmodifications are permitted.
 
 # Require file header
 dotnet_diagnostic.IDE0073.severity = warning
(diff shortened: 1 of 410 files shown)


cschuchardt88:
This mess with alot of other things. I do not like this way.

shargon:
I understand that this is not required after https://github.com/neo-project/neo/pull/4397, isn't it?

shargon:
> > I understand that this is not required after [neo-project/neo#4397](https://github.com/neo-project/neo/pull/4397), isn't it?
> 
> I think it's different. In this case when CopyLocalLockFileAssemblies is set up to true then NuGet package dependencies are copied to the output directory

But plugins are not mandatory, how can it helps?

erikzhang (dismissed):

cschuchardt88:
This isn't the right the way to do this. https://github.com/neo-project/neo/pull/4181 is the correct the way to do it. Ill be adding this PR to this repo, when i get some time.

erikzhang:
> This isn't the right the way to do this.

Could you explain in more detail?

cschuchardt88:
> > This isn't the right the way to do this.
> 
> Could you explain in more detail?

`CopyLocalLockFileAssemblies` is only for transitive `NuGet` package dependencies. This will include other dependencies like `neo.dll`, `akka.dll` and etc from `neo` repo with all the other libraries from that repo. Which will add extra `dll` that are un-needed.

The proper way to add these dependencies is to add an attribute called `PrivateAssets="all"` to the `PackageReference`. Which will not include them when you add reference to the plugin or to `neo` repo libraries.

erikzhang:
> CopyLocalLockFileAssemblies is only for transitive NuGet package dependencies. This will include other dependencies like neo.dll, akka.dll and etc from neo repo with all the other libraries from that repo. Which will add extra dll that are un-needed.

I got it.

> The proper way to add these dependencies is to add an attribute called PrivateAssets="all" to the PackageReference. Which will not include them when you add reference to the plugin or to neo repo libraries.

However, this would require too many changes to the Neo project. I think there should be a way to reference Neo in a plugin project without copying Neo's dependencies to the local directory.

cschuchardt88:
> However, this would require too many changes to the Neo project. I think there should be a way to reference Neo in a plugin project without copying Neo's dependencies to the local directory.

### `<ProjectReference Include="Neo" PrivateAssets="runtime" />` - This will only add `neo.dll` and not all the other `dll` libraries like 
image


Also we can put this in `Directory.Build.props` file

Like this https://github.com/cschuchardt88/neo/blob/a5bf67b181966e7d2d9205c28d95761ec5ec4fa0/src/Plugins/Directory.Build.props

PR #946, opened 2025-12-13
Nullable

erikzhang:

PR #955, opened 2025-12-29
Oracle Fix

shargon:
Related to https://github.com/neo-project/neo-node/pull/946/files#r2651246329

### DIFF
--- a/plugins/OracleService/OracleService.cs
+++ b/plugins/OracleService/OracleService.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2025 The Neo Project.
+// Copyright (C) 2015-2026 The Neo Project.
 //
 // OracleService.cs file belongs to the neo project and is free
 // software distributed under the MIT software license, see the
--- a/.editorconfig
+++ b/.editorconfig
@@ -122,7 +122,7 @@ dotnet_naming_symbols.local_functions.applicable_kinds = local_function
 
 dotnet_naming_style.local_function_style.capitalization = pascal_case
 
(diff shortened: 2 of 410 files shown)


vncoelho (approved):

shargon:
> Can you explain more about this PR?

If CreateResponse throw an error, because the request doesn't exist, or other, it will deny this call

plugins/OracleService/OracleService.cs
```diff
@@ -395,9 +403,11 @@ private void SyncPendingQueue(DataCache snapshot)
         }
     }
 
-    public static Transaction CreateResponseTx(DataCache snapshot, OracleRequest request, OracleResponse response, ECPoint[] oracleNodes, ProtocolSettings settings, bool useCurrentHeight = false)
+    public static Transaction? CreateResponseTx(DataCache snapshot, OracleRequest request, OracleResponse response, ECPoint[] oracleNodes, ProtocolSettings settings, bool useCurrentHeight = false)
     {
-        var requestTx = NativeContract.Ledger.GetTransactionState(snapshot, request.OriginalTxid)!;
+        var requestTx = NativeContract.Ledger.GetTransactionState(snapshot, request.OriginalTxid);
+        if (requestTx == null) return null;
```
erikzhang, line 409: It cannot be `null`. Since every request is created from a transaction, the transaction must exist.

cschuchardt88: The transaction should return 
```csharp
response.Code = OracleResponseCode.Error;
response.Result = Array.Empty<byte>();
```

shargon:
Not required

PR #959, opened 2026-01-07
[N3] Update Neo package

shargon:


Wi1l-B0t:
UT Failed.

ajara87:
> UT Failed.

UTs are not running due to nullability has change when changing Neo package version. PR https://github.com/shargon/neo-cli/pull/2 to ensure document.Servers is not null and then UTs can run.

ajara87 (dismissed):

PR #960, opened 2026-01-08
StorageDumper plugin optimization

shargon:
To carefully verify the difference between states 3.8 and 3.9, we need to add the gas consumption and transaction states.

Those are the main changes:

- Avoid query `NativeContract.Ledger.CurrentIndex(snapshot)` when we can use the `Block` argument.
- Added `applicationExecutedList` because also it's important to check the differences.

### DIFF
--- a/plugins/StorageDumper/StorageDumper.cs
+++ b/plugins/StorageDumper/StorageDumper.cs
@@ -124,7 +124,7 @@ public class StorageDumper : Plugin
 
             _currentBlock = new JObject()
             {
-                ["block"] = blockIndex,
+                ["block"] = block.Index,
                 ["size"] = stateChangeArray.Count,
                 ["storage"] = stateChangeArray
             };


roman-khimov (commented):
It's certainly not an "optimization", it's an extension and to me it doesn't fit well into what StorageDumper is supposed to provide. The data added is what ApplicationLogs plugin already provides and at the same time it's missing some of the things from it (like notifications). So we can just extend the ApplicationLogs in some way to dump things into JSON (which doesn't require any resync, btw, can just be a CLI command) and we'd have all of the data to compare easily (technically comparison can be performed via RPC interface, but likely it'd be slow for millions of blocks) while storage can still be handled by this plugin.

shargon:
> It's certainly not an "optimization", it's an extension and to me it doesn't fit well into what StorageDumper is supposed to provide. The data added is what ApplicationLogs plugin already provides and at the same time it's missing some of the things from it (like notifications). So we can just extend the ApplicationLogs in some way to dump things into JSON (which doesn't require any resync, btw, can just be a CLI command) and we'd have all of the data to compare easily (technically comparison can be performed via RPC interface, but likely it'd be slow for millions of blocks) while storage can still be handled by this plugin.

Changes applied

plugins/StorageDumper/StorageDumper.cs
```diff
@@ -129,10 +124,11 @@ private void OnPersistStorage(uint network, DataCache snapshot)
...
             var bsItem = new JObject()
             {
-                ["block"] = blockIndex,
-                ["size"] = stateChangeArray.Count,
-                ["storage"] = stateChangeArray
+                ["block"] = block.Index,
+                ["size"] = stateChangeArray.Count
             };
+
+            if (stateChangeArray.Count > 0) bsItem["storage"] = stateChangeArray;
```
Wi1l-B0t: This line changes the current behaviour.
The `"storage"` is always present for current behaviour, even if it is empty.

shargon: Less space

Wi1l-B0t: > Less space

It's trivial

Wi1l-B0t: It's not user-friendly if the field is only present when it is not empty,

roman-khimov: Not critical, but I'd not change the behavior here, it's a dumper, it can be a bit verbose.

shargon: > It's not user-friendly if the field is only present when it is not empty

It's a usual behaviour

PR #967, opened 2026-01-14
[N3] Neo-cli integration tests with expect ensuring that all 12 plugins are installed

vncoelho:
First draft for a expect that will ensure all Plugins are installed and running on the neo-cli.

It may need adjustments on configs and Plugins DLLS load conflicts.

### DIFF
--- a/plugins/Directory.Build.props
+++ b/plugins/Directory.Build.props
@@ -10,7 +10,7 @@
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>https://github.com/neo-project/neo-node.git</RepositoryUrl>
     <PackageTags>NEO;Blockchain</PackageTags>
-    <Version>3.9.2</Version>
+    <VersionPrefix>3.9.3</VersionPrefix>
     <TargetFramework>net10.0</TargetFramework>
     <RootNamespace>$(PackageId)</RootNamespace>
     <ImplicitUsings>enable</ImplicitUsings>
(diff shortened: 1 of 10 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Neo Node",
```
vncoelho, line 2: This is duplicated from another PR, I will remove later when the other PR is merged.

.github/workflows/test-locally.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
vncoelho: This file can be removed before merging

.github/workflows/main.yml
```diff
@@ -44,3 +44,58 @@ jobs:
...
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v5
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v5
+      with:
+        dotnet-version: ${{ env.DOTNET_VERSION }}
+    - name: Check format
```
shargon: Format is not required here, it's already done

vncoelho: that is true

vncoelho:
@shargon, I removed format and also removed the check for linux

shargon (approved):

vncoelho:
I think that I will not port it to N4 for now because that branch may still be more unstable.

Issue #973, opened 2026-01-15
Open a second wallet in CLI will not UnregisterSigner for the first wallet


superboyiii:
0. `create a.json`
1. `create b.json`
2. `open wallet a.json`
3. You will see `Error: Signer a already exists`

PR #974, opened 2026-01-15
Fix create wallet error

shargon:
Close https://github.com/neo-project/neo-node/issues/973


superboyiii:
It's not in a scenario of `create wallet`, but `open wallet`.
Please check https://github.com/neo-project/neo-node/pull/976

shargon:
> It's not in a scenario of `create wallet`, but `open wallet`. Please check #976

I can't replicate it without create wallet in the same session, is not unegistered during creation, so this patch is required

superboyiii:
Test pass

PR #975, opened 2026-01-15
Fix create wallet

shargon:
Port of https://github.com/neo-project/neo-node/pull/974

### DIFF
--- a/src/Neo.CLI/CLI/MainService.cs
+++ b/src/Neo.CLI/CLI/MainService.cs
@@ -272,9 +272,7 @@ public partial class MainService : ConsoleServiceBase, IWalletProvider
         void DisplayError(string primaryMessage, string? secondaryMessage = null)
         {
             ConsoleHelper.Error(primaryMessage + Environment.NewLine +
-                                (secondaryMessage != null ? secondaryMessage + Environment.NewLine : "") +
-                                "Press any key to exit.");
-            Console.ReadKey();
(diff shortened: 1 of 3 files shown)


src/Neo.CLI/CLI/MainService.cs
```diff
@@ -154,6 +154,7 @@ public void CreateWallet(string path, string password, bool createDefaultAccount
         }
         wallet.Save();
 
+        if (CurrentWallet is not null) SignerManager.UnregisterSigner(CurrentWallet.Name);
```
erikzhang, line 157: Does SignerManager's role overlap with IWalletProvider's?

Wi1l-B0t (approved):

src/Neo.CLI/CLI/MainService.cs
ajara87: IWalletProvider covers access to the currently active wallet. SignerManager is being used as an ISigner registry to centralize signing (and potentially support other signers beyond wallets). That said I agree an improvement would be to have SignerManager automatically stay in sync by subscribing to IWalletProvider.WalletChanged (e.g. as a follow-up) so we don’t need to manually Register/Unregister in every wallet changing command

ajara87:
Need https://github.com/neo-project/neo-node/pull/989

shargon:
> UT failed.

master build failed, is not related to this PR

PR #984, opened 2026-01-23
Remove invalid transactions from PrepareRequest payload

erikzhang:
If the policy configurations among consensus nodes are inconsistent (e.g., `MaxBlockSize` or `MaxBlockSystemFee`), it may lead to some invalid transactions being persistently retained in some consensus nodes, preventing block creation. A mechanism is needed to ensure that when a transaction is identified as invalid by more than `F` consensus nodes, all consensus nodes will refrain from including that transaction in the `PrepareRequest` payload.

### DIFF
--- a/plugins/StorageDumper/StorageDumper.cs
+++ b/plugins/StorageDumper/StorageDumper.cs
@@ -74,7 +74,7 @@ public class StorageDumper : Plugin
             prefix = BitConverter.GetBytes(contract.Id);
         }
         var states = _system.StoreView.Find(prefix);
-        JArray array = new JArray(states.Where(p => !StorageSettings.Default!.Exclude.Contains(p.Key.Id)).Select(p => new JObject
+        var array = new JArray(states.Where(p => !StorageSettings.Default!.Exclude.Contains(p.Key.Id)).Select(p => new JObject
         {
             ["key"] = Convert.ToBase64String(p.Key.ToArray()),
             ["value"] = Convert.ToBase64String(p.Value.ToArray())
@@ -88,14 +88,9 @@ public class StorageDumper : Plugin
 
     void Blockchain_Committing_Handler(NeoSystem system, Block block, DataCache snapshot, IReadOnlyList<Blockchain.ApplicationExecuted> applicationExecutedList)
     {
-        InitFileWriter(system.Settings.Network, snapshot);
-        OnPersistStorage(system.Settings.Network, snapshot);
-    }
+        InitFileWriter(system.Settings.Network, block);
 
-    private void OnPersistStorage(uint network, DataCache snapshot)
-    {
-        var blockIndex = NativeContract.Ledger.CurrentIndex(snapshot);
-        if (blockIndex >= StorageSettings.Default!.HeightToBegin)
+        if (block.Index >= StorageSettings.Default!.HeightToBegin)
         {
             var stateChangeArray = new JArray();
 
@@ -127,13 +122,12 @@ public class StorageDumper : Plugin
                 stateChangeArray.Add(state);
             }
 
-            var bsItem = new JObject()
+            _currentBlock = new JObject()
             {
-                ["block"] = blockIndex,
+                ["block"] = block.Index,
                 ["size"] = stateChangeArray.Count,
                 ["storage"] = stateChangeArray
             };
-            _currentBlock = bsItem;
         }
     }
 
@@ -152,14 +146,13 @@ public class StorageDumper : Plugin
         }
     }
 
-    private void InitFileWriter(uint network, IReadOnlyStore snapshot)
+    private void InitFileWriter(uint network, Block block)
     {
-        uint blockIndex = NativeContract.Ledger.CurrentIndex(snapshot);
         if (_writer == null
-            || blockIndex % StorageSettings.Default!.BlockCacheSize == 0)
+            || block.Index % StorageSettings.Default!.BlockCacheSize == 0)
         {
-            string path = GetOrCreateDirectory(network, blockIndex);
-            var filepart = (blockIndex / StorageSettings.Default!.BlockCacheSize) * StorageSettings.Default.BlockCacheSize;
+            string path = GetOrCreateDirectory(network, block.Index);
+            var filepart = (block.Index / StorageSettings.Default!.BlockCacheSize) * StorageSettings.Default.BlockCacheSize;
             path = $"{path}/dump-block-{filepart}.dump";
             if (_writer != null)
             {
@@ -180,7 +173,7 @@ public class StorageDumper : Plugin
         return dirPathWithBlock;
     }
 
-    private string GetDirectoryPath(uint network, uint blockIndex)
+    private static string GetDirectoryPath(uint network, uint blockIndex)
     {
         uint folder = (blockIndex / StorageSettings.Default!.StoragePerFolder) * StorageSettings.Default.StoragePerFolder;
         return $"./StorageDumper_{network}/BlockStorage_{folder}";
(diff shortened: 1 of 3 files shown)


plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
```diff
@@ -192,6 +193,16 @@ private void OnChangeViewReceived(ExtensiblePayload payload, ChangeView message)
 
         Log($"{nameof(OnChangeViewReceived)}: height={message.BlockIndex} view={message.ViewNumber} index={message.ValidatorIndex} nv={message.NewViewNumber} reason={message.Reason}");
         context.ChangeViewPayloads[message.ValidatorIndex] = payload;
+        switch (message.Reason)
+        {
+            case ChangeViewReason.TxRejectedByPolicy:
+            case ChangeViewReason.TxInvalid:
+                if (context.InvalidTransactions.TryGetValue(message.RejectedHash, out var hashset))
+                    hashset.Add(message.ValidatorIndex);
```
shargon: `ValidatorIndex` may change between epochs, maybe it's better to add the pubKey?

plugins/DBFTPlugin/Messages/ChangeView.cs
```diff
@@ -33,9 +34,17 @@ public class ChangeView : ConsensusMessage
     /// </summary>
     public ChangeViewReason Reason;
 
+    public UInt256 RejectedHash;
```
shargon: Maybe it can be an array, if we found multiple txs in same block it can deny the service, we should propagate all once

plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
```diff
@@ -83,6 +84,9 @@ internal void EnsureMaxBlockLimitation(Transaction[] txs)
         // Iterate transaction until reach the size or maximum system fee
         foreach (Transaction tx in txs)
         {
+            if (InvalidTransactions.TryGetValue(tx.Hash, out var hashset))
+                if (hashset.Count > F) continue;
```
shargon, line 88: It should be removed from mempool?

plugins/DBFTPlugin/Messages/ChangeView.cs
```diff
@@ -44,12 +53,26 @@ public override void Deserialize(ref MemoryReader reader)
         base.Deserialize(ref reader);
         Timestamp = reader.ReadUInt64();
         Reason = (ChangeViewReason)reader.ReadByte();
+        switch (Reason)
+        {
+            case ChangeViewReason.TxRejectedByPolicy:
+            case ChangeViewReason.TxInvalid:
+                RejectedHash = reader.ReadSerializable<UInt256>();
```
roman-khimov: And btw, this is an incompatible change, inconsistent node versions will break the consensus.

plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
erikzhang: Done.

plugins/DBFTPlugin/Messages/ChangeView.cs
erikzhang: Done.

plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
erikzhang: Removing transactions from the mempool might improve performance, but considering that these transactions are considered valid by this node, they should not be removed from the mempool.  Instead, they should simply be excluded from block proposals to maintain the stability of the consensus algorithm.

plugins/DBFTPlugin/Messages/ChangeView.cs
erikzhang: Therefore, it must be included in the next version and cannot be upgraded separately.

plugins/DBFTPlugin/Messages/ChangeView.cs
```diff
@@ -44,12 +53,26 @@ public override void Deserialize(ref MemoryReader reader)
         base.Deserialize(ref reader);
         Timestamp = reader.ReadUInt64();
         Reason = (ChangeViewReason)reader.ReadByte();
+        switch (Reason)
+        {
+            case ChangeViewReason.TxRejectedByPolicy:
+            case ChangeViewReason.TxInvalid:
+                RejectedHashes = reader.ReadSerializableArray<UInt256>(ushort.MaxValue);
```
shargon, line 60: If it's the end of the stream we should deserialize as empty, or we will be incompatible with other nodes (hardfork)

plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
vncoelho: agreed

plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
```diff
@@ -83,6 +84,9 @@ internal void EnsureMaxBlockLimitation(Transaction[] txs)
         // Iterate transaction until reach the size or maximum system fee
         foreach (Transaction tx in txs)
         {
+            if (InvalidTransactions.TryGetValue(tx.Hash, out var hashset))
```
vncoelho, line 87: Add some comment

// Verify if there is, at least, one honest node that consider this TX as invalid. If so, skip it.

plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
```diff
@@ -21,12 +21,13 @@ namespace Neo.Plugins.DBFTPlugin.Consensus;
...
 {
-    public ExtensiblePayload MakeChangeView(ChangeViewReason reason)
+    public ExtensiblePayload MakeChangeView(ChangeViewReason reason, params UInt256[] rejectedHashes)
     {
         return ChangeViewPayloads[MyIndex] = MakeSignedPayload(new ChangeView
         {
+            Timestamp = TimeProvider.Current.UtcNow.ToTimestampMS(),
             Reason = reason,
-            Timestamp = TimeProvider.Current.UtcNow.ToTimestampMS()
+            RejectedHashes = rejectedHashes
```
ajara87, line 30:
```suggestion
            RejectedHashes = rejectedHashes ?? Array.Empty<UInt256>()
```
Should we initialize rejectedHashes if someone call MakeChangeView with null?

erikzhang: It can be `null` if reason is not `TxRejectedByPolicy` or `TxInvalid`.

plugins/DBFTPlugin/Messages/ChangeView.cs
erikzhang: Currently, there is no way to determine if the `MemoryReader` has reached the end of the data. Modifying the `MemoryReader` would require releasing a new version of Neo. Therefore, we will not make any changes for now and will release the update with the next version. This only involves modifications to the consensus protocol and will not result in a hard fork.

plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
erikzhang: One is not enough; at least more than `F` are needed.

vncoelho: That means one honest

PR #986, opened 2026-01-27
feat: add console output synchronization locker

Jim8y:
## Summary
Fixes neo-project/neo#2995

## Changes
- Add `public static readonly object Locker` to ConsoleHelper for external synchronization
- Wrap `Info()` and `Log()` methods with lock to prevent interleaved output from multiple threads/plugins

## Usage
Plugins can now use:
```csharp
lock(ConsoleHelper.Locker)
{
    // multiple Console.Write calls that won't be interleaved
}
```

The built-in Info/Warning/Error methods are also now thread-safe.

### DIFF
--- a/src/Neo.ConsoleService/ConsoleHelper.cs
+++ b/src/Neo.ConsoleService/ConsoleHelper.cs
@@ -23,6 +23,13 @@ public static class ConsoleHelper
     private static readonly ConsoleColorSet WarningColor = new(ConsoleColor.Yellow);
     private static readonly ConsoleColorSet ErrorColor = new(ConsoleColor.Red);
 
+    /// <summary>
+    /// A shared lock object for synchronizing console output.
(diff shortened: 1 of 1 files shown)


src/Neo.ConsoleService/ConsoleHelper.cs
```diff
@@ -23,6 +23,13 @@ public static class ConsoleHelper
     private static readonly ConsoleColorSet WarningColor = new(ConsoleColor.Yellow);
     private static readonly ConsoleColorSet ErrorColor = new(ConsoleColor.Red);
 
+    /// <summary>
+    /// A shared lock object for synchronizing console output.
+    /// Use this to prevent interleaved output from multiple threads/plugins.
+    /// Example: lock(ConsoleHelper.Locker) { /* multiple Console.Write calls */ }
+    /// </summary>
+    public static readonly object Locker = new();
```
erikzhang, line 31: Why public?

PR #987, opened 2026-01-28
Enable customized storage path for Plugins.

superboyiii:
People want somewhere to store all plugin storage files for many years. However, we only support LevelDBStore and RocksDBStore. This feature could enable customized storage path for Plugins. It's 100% compatible with current solution.
It has two modes:
Example 1:
```json
    "Storage": {
      "Engine": "LevelDBStore",
      "Path": "Data_LevelDB_{0}"
    }
```
or 
```json
    "Storage": {
      "Engine": "LevelDBStore",
      "Path": "AAA_{0}"
    }
```
This one is how currently CLI works, I keep it the same. It will create a folder named as `Data_LevelDB_334F454E` or `AAA_334F454E` in StoragePath. But it can only handle path for 1 plugin. Usually it's LevelDBStore or RocksDBStore because other plugins can't work without them :)

Example 2:
```
    "Storage": {
      "Engine": "LevelDBStore",
      "Path": "E:\\PR-test\\node-PR####\\data"
    }
```
or
```
    "Storage": {
      "Engine": "LevelDBStore",
      "Path": "../data"
    }
```
If Path is not empty(default root path) or including `{0}`, it means new base path is enabled, then all Plugin storage path will be under this `Path`.
```
E:\PR-test\node-PR####\data>tree /a
Folder PATH listing for volume
Volume serial number is 8C69-7726
E:.
+---ApplicationLogs_334F454E
+---Data_LevelDB_334F454E
+---StorageDumper_334F454E
|   \---BlockStorage_0
\---TokenBalanceData_334F454E
```


src/Neo.ConsoleService/Neo.ConsoleService.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <ItemGroup>
+    <PackageReference Include="Neo" Version="3.9.1" />
```
shargon: I prefer to avoid core as reference here, this library must be independent, can you avoid this?

superboyiii: I will try.

superboyiii:
@shargon Done

src/Neo.ConsoleService/UnifiedStoragePath.cs
```diff
@@ -0,0 +1,75 @@
...
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+namespace Neo;
+
+/// <summary>
+/// Helpers to read unified storage base path from config.json (Neo.CLI)
+/// and combine plugin-specific storage paths under that base directory.
+/// </summary>
+public static class UnifiedStoragePath
```
shargon: I think this can be moved to Plugin class, to me seems that is not related to ConsoleService

superboyiii: > I think this can be moved to Plugin class, to me seems that is not related to ConsoleService

But that means these plugins including storage path have to implement this method for each. Or maybe I could implement a toolkit project under plugin.

shargon: No, Plugin class base, in Neo

superboyiii: > No, Plugin class base, in Neo

Done

PR #990, opened 2026-02-01
Change View fix

shargon:
I think we miss this change view after an invalid transaction

### DIFF
--- a/plugins/ApplicationLogs/ApplicationLogs.csproj
+++ b/plugins/ApplicationLogs/ApplicationLogs.csproj
@@ -13,6 +13,12 @@
 	</ItemGroup>
 
 	<ItemGroup>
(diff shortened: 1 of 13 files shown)


AnnaShaleva:
Turns out NeoGo implementation also sends CV in this case (but with `TxInvalid` reason instead of `TxRejectedByPolicy`):
https://github.com/nspcc-dev/dbft/blob/5a44b4c19d470c3682324b49f7281f4a121811fe/dbft.go#L400
https://github.com/nspcc-dev/dbft/blob/5a44b4c19d470c3682324b49f7281f4a121811fe/dbft.go#L406
https://github.com/nspcc-dev/neo-go/blob/9c7004d3311f306984ef9277ec205c2db0760b4d/pkg/consensus/consensus.go#L578

PR #992, opened 2026-02-03
Add Dockerfile and Quick Start

superboyiii:
CLI docker for common usages.

### DIFF
--- a/src/Neo.CLI/Docker/Dockerfile
+++ b/src/Neo.CLI/Docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet/aspnet:10.0 as Build
+FROM mcr.microsoft.com/dotnet/aspnet:10.0 AS Build
 RUN apt-get update && apt-get install -y \
     net-tools \
     telnet \


src/Neo.CLI/Docker/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:24.04 as Build
+RUN apt-get update
+RUN apt-get -y install net-tools telnet bash-completion wget curl lrzsz zip
+RUN apt-get update && apt-get install -y software-properties-common
+RUN add-apt-repository ppa:dotnet/backports
+RUN apt-get update && apt-get install -y dotnet-sdk-10.0
+RUN apt-get install -y apt-transport-https && apt-get update && apt update && apt install build-essential unzip -y
+RUN apt-get install -y sqlite3 libsqlite3-dev libunwind8-dev
+RUN apt-get install -y screen
```
AnnaShaleva: It's a best practice to install system dependencies via single RUN command, ref. https://docs.docker.com/build/building/best-practices/#run, ref. https://github.com/nspcc-dev/neo-bench/blob/dc7f00caa09f6e78dbf7fef9b31d7dc509e6503d/.docker/build/Dockerfile.sharp#L13.

Also, you may take a look at https://github.com/nspcc-dev/neo-bench/blob/master/.docker/build/Dockerfile.sharp and https://github.com/nspcc-dev/neo-bench/blob/master/.docker/build/Dockerfile.sharp.sources.

src/Neo.CLI/Docker/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:24.04 as Build
```
AnnaShaleva: Why not to use `FROM mcr.microsoft.com/dotnet/sdk:10.0 AS Build`? It's less heavy than `ubuntu` and then you don't need to install `dotnet-sdk-10.0` manually.

AnnaShaleva: And then as final image you may use dotnet runtime, because SDK is not really needed to run the neo-cli:
```
FROM mcr.microsoft.com/dotnet/aspnet:10.0 AS final
```

src/Neo.CLI/Docker/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+ARG CLI_VERSION=v3.9.2
+ARG PLUGIN_VERSION=
+RUN if [ -z "$PLUGIN_VERSION" ]; then \
+        ./prepare-node.sh $CLI_VERSION; \
+    else \
+        ./prepare-node.sh $CLI_VERSION $PLUGIN_VERSION; \
+    fi
+
+RUN apt-get update && apt-get install -y vim
```
AnnaShaleva: Looks like this dependency can be avoided.

src/Neo.CLI/Docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    wget \
+    curl \
+    lrzsz \
+    zip \
+    unzip \
+    sqlite3 \
+    libsqlite3-dev \
+    libunwind8-dev \
+    screen \
+    && rm -rf /var/lib/apt/lists/*
```
vncoelho, line 15: very good,

I was using
```bash
	    libsnappy-dev \
	    libleveldb-dev \
	    librocksdb-dev \
```

But maybe not necessary

superboyiii: In this docker, it uses default LevelDB, so it should be OK.

superboyiii:
Tested. It works well.

src/Neo.CLI/Docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM mcr.microsoft.com/dotnet/aspnet:10.0 as Build
```
AnnaShaleva:
```suggestion
FROM mcr.microsoft.com/dotnet/aspnet:10.0 AS Build
```

Nitpicking, to avoid the following build warning:
```
'as' and 'FROM' keywords' casing do not match
```

PR #995, opened 2026-02-04
Fix: Incorrect handling of public key

shargon:
Close #994


Wi1l-B0t:
UT Failed.

shargon:
> UT Failed.

N4 build is broken, will be solved in other pr

Issue #999, opened 2026-02-06
[N4] dBFT with Double Speakers


vncoelho:
We have designed, conducted extensive tests, published papers and mathematical models (MILP) for dBFT 1.0, dBFT 2.0 and double speakers.
The research can not be lost because double speakers is a great improvement that can improve stability and open opportunities for interesting scenarios for running CNs with fallback primaries (collaborative nodes can randomly participate and propose blocks every block without any issue).
It is as important as porting Anti-MEV addition to dBFT 2.0 from NeoX to N4.

For such reason, we will be porting the PR https://github.com/neo-project/neo/pull/3254 to N4.

We hope we can soon merge the PR, it has beeb battle tested for years by our team 
, and it should be the standard for our tests during N4 development.

PR #1001, opened 2026-02-10
[N4] dBFT with Double Speakers

vncoelho:
Closes https://github.com/neo-project/neo-node/issues/999


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
+{
```
vncoelho, line 1: will be moved to another PR later

plugins/DBFTPlugin/Consensus/ConsensusContext.cs
```diff
@@ -159,12 +179,11 @@ public void Dispose()
...
 
-    public Block? EnsureHeader()
+    public Block? EnsureHeader(uint pID)
     {
         if (TransactionHashes == null) return null;
-        Block.Header.MerkleRoot ??= MerkleTree.ComputeRoot(TransactionHashes);
-        return Block;
-    }
+        Block[pID].Header.MerkleRoot ??= MerkleTree.ComputeRoot(TransactionHashes[pID]);
+        return Block[pID];
```
cschuchardt88, line 186: **Bug:** `EnsureHeader` lost its closing `}` (so `Load()` is nested inside it and this does not compile).

Also `TransactionHashes` is now `UInt256[][]` and is never null. The source check was `if (TransactionHashes[pID] == null) return null`. As written, a commit that arrives before the prepare request will NRE in `MerkleTree.ComputeRoot` instead of being stashed.

plugins/DBFTPlugin/Consensus/ConsensusContext.cs
```diff
@@ -119,20 +139,20 @@ public ConsensusContext(NeoSystem neoSystem, DbftSettings settings, ISigner sign
...
-            sc.AddSignature(contract, Validators[i], GetMessage<Commit>(CommitPayloads[i]!).Signature.ToArray());
+            if (GetMessage(CommitPayloads[pID][i])?.ViewNumber != ViewNumber) continue;
+            sc.AddSignature(contract, Validators[i], GetMessage<Commit>(CommitPayloads[pID][i]!).Signature.ToArray());
             j++;
         }
-        Block.Header.Witness = sc.GetWitnesses()[0];
-        Block.Transactions = TransactionHashes!.Select(p => Transactions![p]).ToArray();
-        return Block;
+        Block[pID].Header.Witness = sc.GetWitnesses()[0];
+        Block[pID].Transactions = TransactionHashes!.Select(p => Transactions![p]).ToArray();
```
cschuchardt88, line 154: **Bug:** Still `TransactionHashes!.Select(p => Transactions![p])`, but both are now arrays-of-collections. This does not compile and would not select the txs for `pID`.

Source: `TransactionHashes[pID].Select(p => Transactions[pID][p]).ToArray()`.

plugins/DBFTPlugin/Consensus/ConsensusContext.cs
```diff
@@ -283,50 +329,71 @@ public void Save()
...
+            Transactions[pID] = transactions.Length == 0 && !RequestSentOrReceived ? null : transactions.ToDictionary(p => p.Hash);
+            VerificationContext[pID] = new TransactionVerificationContext();
+            if (Transactions[pID] != null)
+            {
+                foreach (Transaction tx in Transactions[pID].Values)
+                    VerificationContext[pID].AddTransaction(tx);
+            }
         }
+        ChangeViewPayloads[pID] = reader.ReadNullableArray<ExtensiblePayload>(neoSystem.Settings.ValidatorsCount);
+        LastChangeViewPayloads[pID] = reader.ReadNullableArray<ExtensiblePayload>(neoSystem.Settings.ValidatorsCount);
```
cschuchardt88, line 367: **Bug:** `pID` is out of scope here (the `for` ended on the previous line), and `ChangeViewPayloads` / `LastChangeViewPayloads` are still 1-D arrays. Source assigned them directly:

```csharp
ChangeViewPayloads = reader.ReadNullableArray<ExtensiblePayload>(...);
LastChangeViewPayloads = reader.ReadNullableArray<ExtensiblePayload>(...);
```

plugins/DBFTPlugin/Consensus/ConsensusContext.cs
```diff
@@ -88,10 +105,13 @@ public bool ValidatorsChanged
     }
 
     #region Consensus States
-    public bool RequestSentOrReceived => PreparationPayloads[Block.PrimaryIndex] != null;
-    public bool ResponseSent => !WatchOnly && PreparationPayloads[MyIndex] != null;
-    public bool CommitSent => !WatchOnly && CommitPayloads[MyIndex] != null;
-    public bool BlockSent => Block.Transactions != null;
+    public bool RequestSentOrReceived => PreparationPayloads[0][Block[0].PrimaryIndex] != null || (ViewNumber == 0 && PreparationPayloads[1][Block[1].PrimaryIndex] != null);
```
cschuchardt88, line 108: **Bug (speaker admission):** This combined flag is true once *either* speaker's PrepareRequest is stored. `OnPrepareRequestReceived` still returns immediately when it is set, so the second speaker's proposal is dropped (including from recovery). Dual-proposal PreCommit/response logic can never observe both blocks.

Admission needs to be per-`pId` (reject only a duplicate request for the same speaker / same `pId`).

plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
```diff
@@ -29,22 +29,22 @@ public ExtensiblePayload MakeChangeView(ChangeViewReason reason)
...
-        if (CommitPayloads[MyIndex] is ExtensiblePayload payload)
+        if (CommitPayloads[pId][MyIndex] is ExtensiblePayload payload)
             return payload;
 
-        var block = EnsureHeader()!;
-        CommitPayloads[MyIndex] = MakeSignedPayload(new Commit
+        var block = EnsureHeader(pId)!;
+        CommitPayloads[pId][MyIndex] = MakeSignedPayload(new Commit
         {
             Signature = _signer.SignBlock(block, _myPublicKey!, dbftSettings.Network)
```
cschuchardt88, line 40: **Bug (commit rule):** `Commit.PId` is never set, so it serializes as 0. Fallback commits stored in `CommitPayloads[1]` are applied by peers into `CommitPayloads[0][validator]`.

Source: `PId = pId` on this object initializer. `GetCommitPayloadsFromRecoveryMessage` also omits `PId`.

plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
```diff
@@ -126,52 +140,64 @@ public ExtensiblePayload MakeRecoveryRequest()
...
     }
 
-    public ExtensiblePayload MakePrepareResponse()
+    public ExtensiblePayload MakePrepareResponse(uint pId)
     {
-        return PreparationPayloads[MyIndex] = MakeSignedPayload(new PrepareResponse
+        return PreparationPayloads[pId][MyIndex] = MakeSignedPayload(new PrepareResponse
         {
-            PreparationHash = PreparationPayloads[Block.PrimaryIndex]!.Hash
+            PreparationHash = PreparationPayloads[Block[pId].PrimaryIndex]!.Hash,
```
cschuchardt88, line 188: **Bug:** `PreparationPayloads[Block[pId].PrimaryIndex]` uses a validator index as the first dimension of a length-2 array (compile error / wrong payload).

Source: `PreparationPayloads[pId][Block[pId].PrimaryIndex].Hash`.

plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
```diff
@@ -91,27 +91,41 @@ internal void EnsureMaxBlockLimitation(Transaction[] txs)
...
 
-        TransactionHashes = hashes.ToArray();
+        TransactionHashes[pId] = hashes.ToArray();
     }
 
-    public ExtensiblePayload MakePrepareRequest()
+    internal IEnumerable<Transaction> PickTransactions()
     {
         var maxTransactionsPerBlock = neoSystem.Settings.MaxTransactionsPerBlock;
+        var verifiedTxes = neoSystem.MemPool.GetSortedVerifiedTransactions((int)maxTransactionsPerBlock));
```
cschuchardt88, line 104: **Bug:** Extra `)` — this does not compile. Source called `GetSortedVerifiedTransactions()` with no extra paren (max-tx clipping lived in `EnsureMaxBlockLimitation`).

plugins/DBFTPlugin/Consensus/ConsensusService.Check.cs
```diff
@@ -83,17 +98,35 @@ private void CheckExpectedView(byte viewNumber)
...
-        {
-            ExtensiblePayload payload = context.MakeCommit();
-            Log($"Sending {nameof(Commit)}");
-            context.Save();
-            localNode.Tell(new LocalNode.SendDirectly(payload));
-            // Set timer, so we will resend the commit in case of a networking issue
-            ChangeTimer(context.TimePerBlock);
-            CheckCommits();
-        }
+        if (context.PreparationPayloads.Count(p => p != null) >= context.M && context.TransactionHashes[pId]!.All(p => context.Transactions[pId]!.ContainsKey(p)))
```
cschuchardt88, line 103: **Bug (commit rule):** Leftover dBFT2 gate. `PreparationPayloads` is now `ExtensiblePayload[][]`, so `Count(p => p != null)` is the number of inner arrays (at most 2). For N≥4, M≥3 this is never true and PreCommit/Commit never runs.

Source `CheckPreparations` only required `TransactionHashes[pId]` to be filled, then counted `PreparationPayloads[pId]`.

plugins/DBFTPlugin/Consensus/ConsensusService.Check.cs
```diff
@@ -47,20 +47,35 @@ private bool CheckPrepareResponse()
...
+            CheckPreparations(pId);
         }
         return true;
     }
 
-    private void CheckCommits()
+    private void CheckPreCommits(uint pId, bool forced = false)
     {
-        if (context.CommitPayloads.Count(p => context.GetMessage(p)?.ViewNumber == context.ViewNumber) >= context.M && context.TransactionHashes!.All(p => context.Transactions!.ContainsKey(p)))
+        if (forced || context.PreCommitPayloads[pId].Count(p => p != null) >= context.M && context.TransactionHashes[pId].All(p => context.Transactions[pId].ContainsKey(p)))
```
cschuchardt88, line 58: **Bug (commit rule):** `&&` binds tighter than `||`, so this is `forced || (count >= M && allTxs)`. `CheckPreparations` calls `CheckPreCommits(0, true)` on view>0 and on M priority prepares, which then `MakeCommit`s even if txs are missing/null.

Always require the tx set, e.g. `allTxs && (forced || count >= M)`.

plugins/DBFTPlugin/Consensus/ConsensusService.Check.cs
```diff
@@ -83,17 +98,35 @@ private void CheckExpectedView(byte viewNumber)
...
+                    Log($"Sending {nameof(PreCommit)} to pId={pId}");
+                    context.Save();
+                    localNode.Tell(new LocalNode.SendDirectly(payload));
+                    // Set timer, so we will resend the commit in case of a networking issue
+                    ChangeTimer(context.TimePerBlock);
+                    CheckPreCommits(pId);
+                }
+                if (context.ViewNumber == 0 && pId == 0 && preparationsCount >= context.M)
+                {
+                    CheckPreCommits(0, true);
```
cschuchardt88, line 127: **Suggestion (commit rule):** M prepares on the priority speaker skip the PreCommit quorum and force a commit. Nodes that first-saw fallback and already committed via M PreCommits can finalize a different block at the same height. Confirm this fast path against the double-speaker spec; if both speakers are live it is a fork risk.

plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
```diff
@@ -153,29 +158,48 @@ private void OnPrepareRequestReceived(ExtensiblePayload payload, PrepareRequest
...
         }
     }
 
     private void OnPrepareResponseReceived(ExtensiblePayload payload, PrepareResponse message)
     {
         if (message.ViewNumber != context.ViewNumber) return;
-        if (context.PreparationPayloads[message.ValidatorIndex] != null || context.NotAcceptingPayloadsDueToViewChanging) return;
-        if (context.PreparationPayloads[context.Block.PrimaryIndex] != null && !message.PreparationHash.Equals(context.PreparationPayloads[context.Block.PrimaryIndex]!.Hash))
+        if (context.PreparationPayloads[message.PId][message.ValidatorIndex] != null || context.NotAcceptingPayloadsDueToViewChanging) return;
+        if (context.PreparationPayloads[message.PId][context.Block.PrimaryIndex] != null && !message.PreparationHash.Equals(context.PreparationPayloads[message.PId][context.Block[message.PId].PrimaryIndex]!.Hash))
```
cschuchardt88, line 172: **Bug:** `context.Block` is `Block[]`; `PrimaryIndex` does not compile. Source gated this on `RequestSentOrReceived` and compared `PreparationPayloads[message.PId][context.Block[message.PId].PrimaryIndex]`.

Also reject `message.PId > 1` before indexing — a hostile `PId` currently kills the consensus actor.

plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
```diff
@@ -249,6 +274,11 @@ private void OnRecoveryMessageReceived(RecoveryMessage message)
                 foreach (ExtensiblePayload changeViewPayload in changeViewPayloads)
                     if (ReverifyAndProcessPayload(changeViewPayload)) validChangeViews++;
             }
+            else if (context.IsAPrimary)
+            {
+                uint pId = Convert.ToUInt32(!context.IsPriorityPrimary);
+                SendPrepareRequest(pId);
```
cschuchardt88, line 280: **Bug:** Source nested `else if (context.IsAPrimary) SendPrepareRequest(pId)` under `view == current && !RequestSentOrReceived`. This `else if` is attached to `view > current`, so a primary resends a prepare when processing an *older* recovery message.

`SendPrepareRequest(uint pId)` also does not exist on this branch (`ConsensusService.cs` still has a no-arg `SendPrepareRequest()`).

plugins/DBFTPlugin/Messages/RecoveryMessage/RecoveryMessage.cs
```diff
@@ -73,6 +77,7 @@ internal ExtensiblePayload[] GetChangeViewPayloads(ConsensusContext context)
             BlockIndex = BlockIndex,
             ValidatorIndex = p.ValidatorIndex,
             ViewNumber = p.OriginalViewNumber,
+            PId = PId,
```
cschuchardt88, line 80: **Bug:** `ChangeView` has no `PId` property (that type was not updated). This does not compile. Drop the assignment or add the field to `ChangeView` with matching serialize/deserialize/`Size`.

plugins/DBFTPlugin/Messages/RecoveryMessage/RecoveryMessage.cs
```diff
@@ -91,27 +96,42 @@ internal ExtensiblePayload[] GetCommitPayloadsFromRecoveryMessage(ConsensusConte
...
         return context.CreatePayload(PrepareRequestMessage, compact.InvocationScript);
     }
 
     internal ExtensiblePayload[] GetPrepareResponsePayloads(ConsensusContext context)
     {
-        UInt256? preparationHash = PreparationHash ?? context.PreparationPayloads[context.Block.PrimaryIndex]?.Hash;
+        UInt256? preparationHash = PreparationHash ?? context.PreparationPayloads[context.Block[PId].PrimaryIndex]?.Hash;
         if (preparationHash is null) return Array.Empty<ExtensiblePayload>();
-        return PreparationMessages.Values.Where(p => p.ValidatorIndex != context.Block.PrimaryIndex).Select(p => context.CreatePayload(new PrepareResponse
+        return PreparationMessages.Values.Where(p => p.ValidatorIndex != context.Block[0].PrimaryIndex).Select(p => context.CreatePayload(new PrepareResponse
```
cschuchardt88, line 108: **Bug:** Two indexing mistakes vs source:
1. `PreparationPayloads[context.Block[PId].PrimaryIndex]` must be `PreparationPayloads[PId][context.Block[PId].PrimaryIndex]`.
2. The filter uses `Block[0].PrimaryIndex` instead of `Block[PId].PrimaryIndex`, so fallback recovery can drop/include the wrong primary compact.

plugins/DBFTPlugin/Messages/RecoveryMessage/RecoveryMessage.cs
```diff
@@ -21,12 +21,14 @@ namespace Neo.Plugins.DBFTPlugin.Messages;
...
     public required Dictionary<byte, ChangeViewPayloadCompact> ChangeViewMessages;
     public PrepareRequest? PrepareRequestMessage;
     /// The PreparationHash in case the PrepareRequest hasn't been received yet.
     /// This can be null if the PrepareRequest information is present, since it can be derived in that case.
     public UInt256? PreparationHash;
     public required Dictionary<byte, PreparationPayloadCompact> PreparationMessages;
+    public required Dictionary<byte, PreCommitPayloadCompact> PreCommitMessages;
     public required Dictionary<byte, CommitPayloadCompact> CommitMessages;
 
     public override int Size => base.Size
```
cschuchardt88, line 34: **Bug:** `Size` still ignores `PId` (`sizeof(uint)`) and `PreCommitMessages`. Serialize/Deserialize write both. `Verify` also does not bound-check `PreCommitMessages` validator indices (it still only checks ChangeView/Preparation/Commit).

cschuchardt88 (commented):
Incomplete port of dBFT 3.0 double speakers (neo-modules#680 / neo#3254) onto N4 `DBFTPlugin`. Context/messages were split by `pId` and `PreCommit` was added, but **`ConsensusService.cs` was not migrated** (`IsPrimary`, `GetPrimaryIndex`, `MakePrepareRequest()`, `CheckPreparations()`, single `Transactions`/`TransactionHashes`). `PrimaryTimerMultiplier` (1 vs 4/3) is defined and never applied.

Several methods also diverge from the source PR and do not compile (`EnsureHeader` missing `}`, `CreateBlock` still uses unindexed `TransactionHashes`/`Transactions`, `ChangeViewPayloads[pID]` after the deserialize loop, extra `)` in `PickTransactions`, `ChangeView.PId`, `Block.PrimaryIndex` on a `Block[]`). Consensus-sensitive issues below: speaker admission, commit tagging, and commit rules.

**Not in the diff (body-only):**
- `plugins/DBFTPlugin/Consensus/ConsensusService.cs` — timer, `SendPrepareRequest(uint pId)`, `OnTimer`, `OnStart` recovery fill, `OnTransaction`/`AddTransaction` must be per-`pId` as in the source service.
- Recovery `GetCommitPayloadsFromRecoveryMessage` does not set `Commit.PId` (defaults to 0), so recovered fallback commits land in the priority slot.
- `message.PId` is used as a raw array index with no `pId <= 1` check (consensus-actor DoS).

Please treat this as not mergeable until it builds and the dual-speaker prepare/commit path matches the source behavior (or an updated spec).

vncoelho:
This needs rebase

PR #1016, opened 2026-03-18
testing some rpcserver calls

vncoelho:
The idea here is to add calls for RPCServer and other plugins in the future, making this integration tests for neo-node to test different plugins functionalities.

For now it can be enabled always. In the future we can trigger it on specific moments if it starts to become a problem.


vncoelho:
I am working on another branch

PR #1023, opened 2026-04-01
Fix: adaptation for Wallet setName

Wi1l-B0t:
Adaptation for Wallet setName

https://github.com/neo-project/neo/pull/4509

### DIFF
--- a/plugins/SQLiteWallet/SQLiteWallet.cs
+++ b/plugins/SQLiteWallet/SQLiteWallet.cs
@@ -34,7 +34,11 @@ internal class SQLiteWallet : Wallet
     private readonly ScryptParameters _scrypt;
     private readonly Dictionary<UInt160, SQLiteWalletAccount> _accounts;
 
-    public override string Name => GetFileNameWithoutExtension(Path);
+    public override string Name
+    {
+        get => GetWalletName();
+        set => ChangeWalletName(value);
+    }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


plugins/SQLiteWallet/SQLiteWallet.cs
```diff
@@ -191,6 +195,28 @@ private WalletDataContext BuildDatabase()
...
+            using var ctx = new WalletDataContext(Path);
+            var name = LoadStoredData(ctx, "WalletName");
+            if (name == null || name.Length == 0)
+                return GetFileNameWithoutExtension(Path);
+            return Encoding.UTF8.GetString(name);
+        }
+    }
+
+    private void ChangeWalletName(string name)
+    {
```
ajara87, line 211:
```suggestion
    {
        if (name == null) throw new ArgumentNullException(nameof(name));
```

ajara87: Check that name is not null?

shargon: We use nullable enable in csproj, should be not required

shargon (approved):

plugins/SQLiteWallet/SQLiteWallet.cs
Wi1l-B0t: nullable enabled.

ajara87 (approved):

PR #1019, opened 2026-03-29
[N4] port message sign

ajara87:
Port https://github.com/neo-project/neo-node/pull/924

PR #1027, opened 2026-04-05
[N3] - Check message is null

ajara87:
Check if message is null in sign and verify.

Related to https://github.com/neo-project/neo-node/pull/1019


shargon (approved):
We should check when it's null if it's not nullable, but it's ok

PR #1030, opened 2026-04-17
Reduce MaxTransactionsPerBlock from 512 to 200

shargon:
As agreed in the last meeting, this value will be reduced

### DIFF
--- a/src/Neo.CLI/config.json
+++ b/src/Neo.CLI/config.json
@@ -33,7 +33,7 @@
     "Network": 860833102,
     "AddressVersion": 53,
     "MillisecondsPerBlock": 15000,
-    "MaxTransactionsPerBlock": 512,
+    "MaxTransactionsPerBlock": 200,
     "MemoryPoolMaxTransactions": 50000,
     "MaxTraceableBlocks": 2102400,
     "Hardforks": {},


shargon:
@erikzhang Not sure if N4 should be affected.

AnnaShaleva:
I suppose we'd better keep N4 value unchanged since it's a separate chain. Also `200` is designated to be applied during 3s-blocks transition period only, it will be reverted once we're sure N3 is OK with 3s blocks.

ajara87 (approved):

Wi1l-B0t:
> I suppose we'd better keep N4 value unchanged since it's a separate chain. Also `200` is designated to be applied during 3s-blocks transition period only, it will be reverted once we're sure N3 is OK with 3s blocks.

Why?

superboyiii:
> > I suppose we'd better keep N4 value unchanged since it's a separate chain. Also `200` is designated to be applied during 3s-blocks transition period only, it will be reverted once we're sure N3 is OK with 3s blocks.
> 
> Why?

No 3s consensus time then no 200 MaxTransactionsPerBlock. N4 is not launched yet, all settings are in default value. We don't need to change default parameters.

PR #1029, opened 2026-04-17
Neo.CLI: set MaxTransactionsPerBlock to 200 for N3 mainnet

AnnaShaleva:
Ref. https://neonewstoday.com/governance/neo-council-approves-3-second-block-time-defers-fee-reduction-to-gorgon-hard-fork/. The default value will not be changed.

PR #1031, opened 2026-04-17
Change MaxTransactionsPerBlock from 512 to 200

shargon:
As agreed in the last meeting, this value will be reduced


shargon:
I didn't saw https://github.com/neo-project/neo-node/pull/1029

PR #1032, opened 2026-04-18
Fix: adaptation for Crypto.Sign

Wi1l-B0t:
Fix:  adaptation for Crypto.Sign.

The definition of `Crypto.Sign` has been changed.

### DIFF
--- a/src/Neo.GUI/GUI/SigningDialog.cs
+++ b/src/Neo.GUI/GUI/SigningDialog.cs
@@ -80,11 +80,10 @@ internal partial class SigningDialog : Form
         }
 
         var account = (WalletEntry)cmbAddress.SelectedItem;
-        var keys = account.Account.GetKey();
-
+        var keypair = account.Account.GetKey();
         try
         {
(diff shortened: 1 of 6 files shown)


src/Neo.GUI/GUI/SigningDialog.cs
```diff
@@ -80,11 +80,10 @@ private void button1_Click(object sender, EventArgs e)
         }
 
         var account = (WalletEntry)cmbAddress.SelectedItem;
-        var keys = account.Account.GetKey();
-
+        var keypair = account.Account.GetKey();
```
ajara87, line 83: add check if keypair is null?

```suggestion
        var keypair = account.Account.GetKey();
        if (keypair == null)
            return;
```

shargon: We should show an error, but Null error is not good.

shargon (approved):
LGTM

ajara87 (approved):

PR #1039, opened 2026-04-29
Add Rpc sign and relay method

superboyiii:
It's 100% compatible with the format of CLI.

### DIFF
--- a/plugins/RpcServer/RpcServer.Node.cs
+++ b/plugins/RpcServer/RpcServer.Node.cs
@@ -15,6 +15,7 @@ using Neo.Json;
 using Neo.Ledger;
 using Neo.Network.P2P;
 using Neo.Network.P2P.Payloads;
+using Neo.SmartContract;
 using static Neo.Ledger.Blockchain;
 
 namespace Neo.Plugins.RpcServer;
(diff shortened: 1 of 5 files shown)


plugins/RpcServer/RpcServer.Node.cs
```diff
@@ -213,6 +214,64 @@ protected internal virtual JToken SendRawTransaction(string base64Tx)
...
+        {
+            throw new RpcException(RpcErrorFactory.InvalidParams($"Invalid signature context: {ex.Message}"));
+        }
+
+        context.Completed.True_Or(RpcErrorFactory.InvalidParams("The signature is incomplete."));
+        if (context.Verifiable is not Transaction tx)
+            throw new RpcException(RpcErrorFactory.InvalidParams("Only support to relay transaction."));
+
+        tx.Witnesses = context.GetWitnesses();
+        var reason = system.Blockchain.Ask<RelayResult>(tx).Result;
```
shargon, line 271: Could you check first if context IsCompleted?

superboyiii: We need parse first, so we can make sure 100% complete. If we check before parse, then we need check complete again after parse.

shargon: No, after parse, but maybe the signature is not full to relay

superboyiii: > No, after parse, but maybe the signature is not full to relay

But it's already checked in
```C#
context.Completed.True_Or(RpcErrorFactory.InvalidParams("The signature is incomplete."));
```
f0cf2de9-6bde-4768-8bc8-73bdb81d5bea

Wi1l-B0t (approved):

plugins/RpcServer/RpcServer.Node.cs
shargon: My bad

shargon (approved):

PR #1043, opened 2026-05-08
Fix OracleService start-after-stop state

Jim8y:
## Summary
- Recreate the OracleService cancellation source when an operator starts the service again after `stop oracle`.
- Keep an in-progress stop from being reset by another `start oracle` command before the previous processing task finishes.
- Dispose stale oracle protocol instances before creating fresh protocol handlers.
- Store finished oracle request timestamps using the current UTC time.
- Add regression coverage for the start-after-stop lifecycle and finished request cache timestamps.

## Testing
- `dotnet test tests/Neo.Plugins.OracleService.Tests/Neo.Plugins.OracleService.Tests.csproj -c Release --no-restore`
- `dotnet format --no-restore --verify-no-changes --include plugins/OracleService/OracleService.cs tests/Neo.Plugins.OracleService.Tests/UT_OracleService.cs`
- GitHub Actions on `9a2fd930f50541bbb5dd4674b4c60db346365ae1`: `Test (ubuntu-latest)`, `Test (windows-latest)`, `Test-with-plugins (ubuntu-latest)`, `codecov/patch`, and `codecov/project` passed.

### DIFF
--- a/plugins/OracleService/OracleService.cs
+++ b/plugins/OracleService/OracleService.cs
@@ -46,17 +46,17 @@ public sealed class OracleService : Plugin
 
     private Wallet wallet;
     private readonly ConcurrentDictionary<ulong, OracleTask> pendingQueue = new();
-    private readonly ConcurrentDictionary<ulong, DateTime> finishedCache = new();
+    internal readonly ConcurrentDictionary<ulong, DateTime> finishedCache = new();
     private Timer timer;
     internal CancellationTokenSource cancelSource = new();
-    private OracleStatus status = OracleStatus.Unstarted;
-    private Task processingTask;
-    private readonly object lifecycleLock = new();
+    internal OracleStatus status = OracleStatus.Unstarted;
+    internal Task processingTask;
+    private readonly Lock lifecycleLock = new();
     private IWalletProvider walletProvider;
     private int counter;
-    private NeoSystem _system;
+    internal NeoSystem _system;
 
-    private readonly Dictionary<string, IOracleProtocol> protocols = new();
+    internal readonly Dictionary<string, IOracleProtocol> protocols = new();
 
     public override string Description => "Built-in oracle plugin";
 
@@ -178,7 +178,7 @@ public sealed class OracleService : Plugin
     }
 
     [ConsoleCommand("stop oracle", Category = "Oracle", Description = "Stop oracle service")]
-    private void OnStop()
+    internal void OnStop()
     {
         lock (lifecycleLock)
         {
@@ -585,7 +585,7 @@ public sealed class OracleService : Plugin
         }
     }
 
-    private void MarkRequestFinished(ulong requestId)
+    internal void MarkRequestFinished(ulong requestId)
     {
         finishedCache.TryAdd(requestId, TimeProvider.Current.UtcNow);
     }
@@ -652,7 +652,7 @@ public sealed class OracleService : Plugin
         public readonly DateTime Timestamp = TimeProvider.Current.UtcNow;
     }
 
-    enum OracleStatus
+    internal enum OracleStatus
     {
         Unstarted,
         Running,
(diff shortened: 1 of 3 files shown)


ajara87 (dismissed):

plugins/OracleService/OracleService.cs
```diff
@@ -150,7 +151,13 @@ public Task Start(Wallet wallet)
         status = OracleStatus.Running;
         timer = new Timer(OnTimer, null, RefreshIntervalMilliSeconds, Timeout.Infinite);
         ConsoleHelper.Info($"Oracle started");
-        return ProcessRequestsAsync();
+        return ProcessRequestsAsync(cancelSource.Token);
+    }
+
+    private void ResetCancellationSource()
```
shargon, line 164: I think that it's better to avoid multiple start that allow to reset, and is not disposed the previous one

shargon (changes requested):

Jim8y:
Addressed the lifecycle feedback by disposing stale protocol instances before rebuilding them, and added coverage for the start-after-stop path.

ajara87 (dismissed):

shargon (commented):
When is restarted the service? we don't have a restart in node

plugins/OracleService/OracleService.cs
Jim8y: Handled by keeping the current processing task alive while the service is stopping. A repeated `start oracle` now returns the current task instead of replacing the cancellation source or protocol instances, and the protocol handlers are disposed when the run finishes before a later start creates fresh ones.

Jim8y:
There is no node-level restart path here. The lifecycle covered by this PR is the OracleService console flow: an operator can run `stop oracle` and then `start oracle` again in the same node process.

I updated the PR title and description to use `start after stop` instead of `restart`, so it does not imply a node restart. The guard is only for preventing a second `start oracle` while the previous stop is still draining from replacing the active cancellation source or protocol instances.

Jim8y:
@shargon Could you re-check this when you have a chance? The wording has been updated from restart to start-after-stop; the covered path is the `stop oracle` / `start oracle` console flow in the same node process, not a node restart.

tests/Neo.Plugins.OracleService.Tests/UT_OracleService.cs
```diff
@@ -98,4 +102,141 @@ public void TestCreateOracleResponseTx()
...
+            await secondRun.WaitAsync(TimeSpan.FromSeconds(5));
+            oracle.Dispose();
+        }
+    }
+
+    [TestMethod]
+    public void TestStartWhileStoppingDoesNotResetCurrentRun()
+    {
+        var oracle = new OracleService();
+        var statusType = typeof(OracleService).GetNestedType("OracleStatus", BindingFlags.NonPublic);
```
shargon: I think we should avoid reflection tests, that's why internal and internal visible was made

plugins/OracleService/OracleService.cs
```diff
@@ -48,8 +48,10 @@ public sealed class OracleService : Plugin
     private readonly ConcurrentDictionary<ulong, OracleTask> pendingQueue = new();
     private readonly ConcurrentDictionary<ulong, DateTime> finishedCache = new();
     private Timer timer;
-    internal readonly CancellationTokenSource cancelSource = new();
+    internal CancellationTokenSource cancelSource = new();
     private OracleStatus status = OracleStatus.Unstarted;
+    private Task processingTask;
+    private readonly object lifecycleLock = new();
```
shargon:
```suggestion
    private readonly Lock lifecycleLock = new();
```

shargon (commented):
@superboyiii could you test it?

superboyiii:
> @superboyiii could you test it?

Sure

superboyiii:
@Jim8y UT failed.

superboyiii:
The bug is true. From my test result, it's fixed. Good job, Jimmy.

superboyiii:
Origin:
image

Now:
image

shargon (approved):

superboyiii (approved):

superboyiii (approved):

PR #1045, opened 2026-05-08
Ignore duplicate future state roots

Jim8y:
## Summary
- ignore duplicate future state roots instead of throwing from the pending cache
- add a StateService regression test for repeated future state root handling

## Tests
- dotnet test tests/Neo.Plugins.StateService.Tests/Neo.Plugins.StateService.Tests.csproj -c Release --no-restore --logger "console;verbosity=normal"
- dotnet format --no-restore --verify-no-changes --include plugins/StateService/Storage/StateStore.cs tests/Neo.Plugins.StateService.Tests/UT_StatePlugin.cs
- git diff --check

### DIFF
--- a/plugins/StateService/Storage/StateStore.cs
+++ b/plugins/StateService/Storage/StateStore.cs
@@ -109,8 +109,7 @@ class StateStore : UntypedActor
         if (LocalRootIndex is null) throw new InvalidOperationException(nameof(StateStore) + " could not get local root index");
         if (LocalRootIndex < stateRoot.Index && stateRoot.Index < LocalRootIndex + MaxCacheCount)
         {
-            _cache.Add(stateRoot.Index, stateRoot);
-            return true;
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

Wi1l-B0t:
What is the purpose of this modification?

Jim8y:
The purpose is to make future state root handling idempotent. A future root can be received more than once while it is still waiting in the pending cache, for example through repeated network delivery or a retry path before the local root index catches up.

Using `Dictionary.Add` makes that duplicate delivery throw and can interrupt StateService even though the root is already known. `TryAdd` keeps the first cached value and returns `false` for the duplicate, so the caller can ignore it without changing persisted state. The added test covers this repeated future-root case.

shargon (commented):
The return of this method is always discarded, isn't it? It should be better if the ut use Akka instead of reflection, but as I see the bool result is not checked, maybe should log it when false?

PR #1047, opened 2026-05-09
Add pending tx and auto relay mechanism in RPC and CLI

superboyiii:
**`Motivation`**: Transactions can fail relay with NotYetValid when ValidUntilBlock is too far ahead of the current chain head. This PR adds the DeferredRelay plugin so nodes can queue those transactions locally (when enabled) and retry relay later, instead of dropping them immediately.

**`DeferredRelay.json`**: Under PluginConfiguration, set MaxTransactions and CheckFrequency both > 0 to enable the feature; CheckFrequency sets how often (in persisted blocks) the node rescans the local pending store. Requires the RpcServer plugin.

**`CLI`**: New command `list pending` — prints the JSON snapshot of locally queued NotYetValid transactions and related height / plugin settings (same data shape the RPC exposes).

**`RPC`**: New method `getpendingvaliduntilrelay` (no params) — returns that snapshot as JSON; when disabled or the store is not active, enabled is false. sendrawtransaction still returns -510 with data: "NotYetValid"; when DeferredRelay is enabled, qualifying txs are queued automatically via relay events.


Wi1l-B0t:
More unit tests?

vncoelho (commented):
I think this should be an extra plugin as an optional node service.

shargon:
Maybe it's easier allow the user to config a higer value than `MaxValidUntilBlockIncrement`

```
if (ValidUntilBlock <= height || ValidUntilBlock > height + snapshot.GetMaxValidUntilBlockIncrement(settings))
    return VerifyResult.Expired;
```

We already have `ProtocolSettings` in `tx.VerifyStateDependent`, so if the node wants to retain higher values, it can simply set a higher one

superboyiii:
> Maybe it's easier allow the user to config a higer value than `MaxValidUntilBlockIncrement`
> 
> ```
> if (ValidUntilBlock <= height || ValidUntilBlock > height + snapshot.GetMaxValidUntilBlockIncrement(settings))
>     return VerifyResult.Expired;
> ```
> 
> We already have `ProtocolSettings` in `tx.VerifyStateDependent`, so if the node wants to retain higher values, it can simply set a higher one

According to what I discussed with @roman-khimov previously, for safety reasons, we even tend to narrow the `MaxValidUntilBlockIncrement` period. Especially for the multi-sig scene, currently we only have 1/5 day as `MaxValidUntilBlockIncrement`. We urgently need this storing  pending tx and auto relay mechanism. In furure, we recommend constructing transactions and collecting signatures in early time and set `validuntilblock` at enough time after `Currnet height` + `MaxValidUntilBlockIncrement` .

superboyiii:
> I think this should be an extra plugin as an optional node service.

It's not complex, the trigger is `relay` command in CLI and RPC, so it's not worth creating a new plugin just for this.

superboyiii:
> More unit tests?

Done

vncoelho:
> > Maybe it's easier allow the user to config a higer value than `MaxValidUntilBlockIncrement`
> > 
> > ```
> > if (ValidUntilBlock <= height || ValidUntilBlock > height + snapshot.GetMaxValidUntilBlockIncrement(settings))
> >     return VerifyResult.Expired;
> > ```
> > 
> > We already have `ProtocolSettings` in `tx.VerifyStateDependent`, so if the node wants to retain higher values, it can simply set a higher one
> 
> According to what I discussed with @roman-khimov previously, for safety reasons, we even tend to narrow the `MaxValidUntilBlockIncrement` period. Especially for the multi-sig scene, currently we only have 1/5 day as `MaxValidUntilBlockIncrement`. We urgently need this storing  pending tx and auto relay mechanism. In furure, we recommend constructing transactions and collecting signatures in early time and set `validuntilblock` at enough time after `Currnet height` + `MaxValidUntilBlockIncrement` . 

I do not think this is the correct way to go, adding an "extra mempool".

superboyiii:
> 
> I do not think this is the correct way to go, adding an "extra mempool".

It's a very little local IStore which will not be synced to peers. It's only for temporary use and it will be removed when the height is passed. In terms of resource consumption, it is imperceptible and almost negligible.

vncoelho:
> > 
> > I do not think this is the correct way to go, adding an "extra mempool".
> 
> It's a very little local mempool which will not be synced to peers. It's only for temporary use and it will be removed when the height is passed. In terms of resource consumption, it is imperceptible and almost negligible.
> 

Since it is not synced, it is indeed a local mempool.

However, in my opinion,  it should be a plugin.

superboyiii:
> However, in my opinion, it should be a plugin.

It's not active on default. Not a problem. A new plugin means new command and new RPC, it's not necessary.

shargon:
> > Maybe it's easier allow the user to config a higer value than `MaxValidUntilBlockIncrement`
> > ```
> > if (ValidUntilBlock <= height || ValidUntilBlock > height + snapshot.GetMaxValidUntilBlockIncrement(settings))
> >     return VerifyResult.Expired;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We already have `ProtocolSettings` in `tx.VerifyStateDependent`, so if the node wants to retain higher values, it can simply set a higher one
> 
> According to what I discussed with @roman-khimov previously, for safety reasons, we even tend to narrow the `MaxValidUntilBlockIncrement` period. Especially for the multi-sig scene, currently we only have 1/5 day as `MaxValidUntilBlockIncrement`. We urgently need this storing pending tx and auto relay mechanism. In furure, we recommend constructing transactions and collecting signatures in early time and set `validuntilblock` at enough time after `Currnet height` + `MaxValidUntilBlockIncrement` .

It's a bypass of `MaxValidUntilBlockIncrement`, ao whats the difference with this new mempool or having the tx in the original mempool?

superboyiii:
> > 
> > According to what I discussed with @roman-khimov previously, for safety reasons, we even tend to narrow the `MaxValidUntilBlockIncrement` period. Especially for the multi-sig scene, currently we only have 1/5 day as `MaxValidUntilBlockIncrement`. We urgently need this storing pending tx and auto relay mechanism. In furure, we recommend constructing transactions and collecting signatures in early time and set `validuntilblock` at enough time after `Currnet height` + `MaxValidUntilBlockIncrement` .
> 
> It's a bypass of `MaxValidUntilBlockIncrement`, ao whats the difference with this new mempool or having the tx in the original mempool?

This new one is an isolated IStore. It's not in cache, so it's not mempool in fact. Such kind of pending tx will not go into original mempool before because it's verified as `Expired`. But this time, I make a chance for these txs go into this isolated IStore. They're pending until correct time, then they will be relayed automatically and be removed from this isolated IStore.

roman-khimov:
I tend to agree with @vncoelho on this, an additional plugin makes sense. Overriding MVUBI locally to put into the main mempool is not a solution, there are subscriptions to mempool, there are interfaces to obtain pooled transactions, there are rebroadcasts tied to mempool, CNs get their transactions from the same main mempool as well. We can't have invalid (as per protocol) transactions there, something will break for sure this way.

So given that it's a separate store the only question is where to put it. Currently it's tied to CLI, but it'd not dependent on CLI (transactions can be pushed via RPC and CLI code shouldn't care much about where transactions are stored), so a separate plugin would be structurally more sound.

superboyiii:
I will wait for conclusion on https://github.com/neo-project/neo/pull/4545, so I can decide how to make changes. Since @vncoelho and @roman-khimov prefer a Plugin way, it's OK for me.

src/Neo.CLI/CLI/PendingValidUntilRelay.cs
```diff
@@ -0,0 +1,247 @@
...
+                if (relayResult.Result is VerifyResult.Succeed or VerifyResult.AlreadyInPool or VerifyResult.AlreadyExists)
+                    toRemove.Add(key);
+            }
+        }
+
+        foreach (byte[] key in toRemove)
+            host.Store.Delete(key);
+    }
+
+    private static int CountPendingRelayEntries(IStore store)
```
ajara87: This can be very costly. Wouldn’t it be better to have a separate counter?

superboyiii: > This can be very costly. Wouldn’t it be better to have a separate counter?

It's not necessary. Although this one is costly, `PendingRelayMaxTransactions` aims to a local IStore, so it stays small.  TryOffer is not called at high rates. At least for now, I haven't found any user scenarios where this new feature will be used frequently on local CLI. It's almost for multi-sig scenario.

Jim8y:
Since neo-project/neo#4545 has been merged, this PR needs to be updated before the design can be re-reviewed. The far-future `ValidUntilBlock` case no longer maps to `VerifyResult.Expired`; after rebasing onto current `master-n3`, it should use the new `VerifyResult.ValidUntilBlockTooFarInFuture` result explicitly. Otherwise those transactions will not enter the pending store through the CLI/RPC offer path.

Wi1l-B0t:
Conflicts

superboyiii:
> Since [neo-project/neo#4545](https://github.com/neo-project/neo/pull/4545) has been merged, this PR needs to be updated before the design can be re-reviewed. The far-future `ValidUntilBlock` case no longer maps to `VerifyResult.Expired`; after rebasing onto current `master-n3`, it should use the new `VerifyResult.ValidUntilBlockTooFarInFuture` result explicitly. Otherwise those transactions will not enter the pending store through the CLI/RPC offer path.

I will make it as a plugin. Need a little time to modify.

superboyiii:
@shargon @ajara87 Please review again.

plugins/DeferredRelay/DeferredRelay.json
```diff
@@ -0,0 +1,12 @@
+{
+  "PluginConfiguration": {
+    "Path": "DeferredRelay_{0}",
+    "Network": 860833102,
```
shargon, line 4: It can relay to a different network, so this value should be taken from the `ProtocolSettings`

plugins/DeferredRelay/DeferredRelaySettings.cs
```diff
@@ -0,0 +1,52 @@
...
+    public UnhandledExceptionPolicy ExceptionPolicy { get; }
+
+    public static DeferredRelaySettings Default { get; private set; } = default!;
+
+    public bool Enabled => MaxTransactions > 0 && CheckFrequency > 0;
+
+    private DeferredRelaySettings(IConfigurationSection section)
+        : this(
+            section.GetValue("Path", "DeferredRelay_{0}")!,
+            section.GetValue("Network", 5195086u),
```
shargon: Wrong default value

src/Neo.CLI/CLI/MainService.Network.cs
```diff
@@ -159,8 +160,12 @@ private void OnRelayCommand(JObject jsonObjectToRelay)
                 return;
             }
             tx.Witnesses = context.GetWitnesses();
-            NeoSystem.Blockchain.Tell(tx);
-            Console.WriteLine($"Data relay success, the hash is shown as follows: {Environment.NewLine}{tx.Hash}");
+            var relayResult = NeoSystem.Blockchain.Ask<Blockchain.RelayResult>(tx, TimeSpan.FromSeconds(30))
+                .ConfigureAwait(false).GetAwaiter().GetResult();
+            if (relayResult.Result == VerifyResult.Succeed)
+                Console.WriteLine($"Data relay success, the hash is shown as follows: {Environment.NewLine}{tx.Hash}");
```
shargon:
```suggestion
                Console.WriteLine($"Data relay success, the hash is shown as follows:{Environment.NewLine}{tx.Hash}");
```

.github/workflows/main.yml
```diff
@@ -82,7 +82,17 @@ jobs:
     - name: Install dependencies
       run: sudo apt-get install -y libleveldb-dev expect
     - name: Run tests with expect
-      run: expect ./.github/workflows/test-neo-cli-plugins.expect
+      run: |
+        # Count loadable plugins only (exclude MPTTrie dependency and RpcClient library).
```
shargon, line 86: Why the exclude?

superboyiii: > Why the exclude?

Their classes are `Neo.Cryptography.MPT`  and `Neo.Network.RPC.RpcClient`.
Only classes that inherit from Plugin are considered real plugins.

plugins/DeferredRelay/DeferredRelaySettings.cs
superboyiii: I found most plugins are using this value as default. I will fix all of them.

plugins/DeferredRelay/DeferredRelayEngine.cs
```diff
@@ -0,0 +1,172 @@
...
+            if (height >= tx.ValidUntilBlock)
+            {
+                toRemove.Add(key);
+                continue;
+            }
+
+            if (height < tx.ValidUntilBlock && tx.ValidUntilBlock <= height + maxInc)
+            {
+                var relayResult = system.Blockchain.Ask<Blockchain.RelayResult>(tx, TimeSpan.FromSeconds(30))
+                    .ConfigureAwait(false).GetAwaiter().GetResult();
```
ajara87: I think it would be better to use an asynchronous method to avoid blocking due to this statement

plugins/DeferredRelay/DeferredRelay.json
superboyiii: > It can relay to a different network, so this value should be taken from the `ProtocolSettings`

No. It's no way to relay to another chain. If DeferredRelay.json Network ≠ the node’s ProtocolSettings.Network, OnSystemLoaded returns immediately: no store, no actor, no queue, no retry. The plugin does nothing on that node. Cross-chain relay is not possible here; The worst case is a silent misconfiguration (you think DeferredRelay is on, but it isn’t).

plugins/DeferredRelay/DeferredRelayEngine.cs
superboyiii: > I think it would be better to use an asynchronous method to avoid blocking due to this statement

You're right. Now moved queue processing to `ProcessQueuedAsync` and schedule it from the actor without blocking `OnReceive`; `relay` uses await Blockchain.Ask instead of GetAwaiter().GetResult().

ajara87: great!

superboyiii:
@shargon @ajara87 Ready to merge?

ajara87 (dismissed):

ajara87:
> @shargon @ajara87 Ready to merge?

lgtm

plugins/DeferredRelay/DeferredRelayEngine.cs
```diff
@@ -0,0 +1,196 @@
...
+        return true;
+    }
+
+    public static bool ShouldProcessPersist(Block block, DeferredRelaySettings settings)
+    {
+        if (!settings.Enabled)
+            return false;
+        if (block.Index == 0)
+            return false;
+        return block.Index % settings.CheckFrequency == 0;
```
Wi1l-B0t, line 136: The default value of `CheckFrequency` is 0.

So line 120 may throw exception.

superboyiii: > The default value of `CheckFrequency` is 0.
> 
> So line 120 may throw exception.

If `CheckFrequency` is 0, then it will return in `line 116` (`if (!settings.Enabled) return false;`), because `public bool Enabled => MaxTransactions > 0 && CheckFrequency > 0;`

superboyiii: I improved it, made it more readable although it's unreachable.

plugins/DeferredRelay/DeferredRelayPlugin.cs
```diff
@@ -0,0 +1,95 @@
...
+        RpcServerPlugin.RegisterMethods(this, DeferredRelaySettings.Default.Network);
+
+        if (!DeferredRelaySettings.Default.Enabled)
+            return;
+
+        var networkId = system.Settings.Network.ToString("X8");
+        var pluginPath = string.Format(DeferredRelaySettings.Default.Path, networkId);
+        var path = PluginHelper.ApplyUnifiedStoragePath(pluginPath);
+        var fullPath = GetFullPath(path);
+        System.IO.Directory.CreateDirectory(System.IO.Path.GetFullPath(fullPath));
```
ajara87: Wouldn’t it be redundant to call `GetFullPath` twice?

superboyiii: > Wouldn’t it be redundant to call `GetFullPath` twice?

You're right. Fixed.

plugins/DeferredRelay/DeferredRelay.json
shargon: So what's the sense of have this variable in the config, if it must be the same as the current `ProjectSettings`?

plugins/DeferredRelay/DeferredRelayPlugin.cs
```diff
@@ -0,0 +1,95 @@
...
+    public override string ConfigFile => Combine(RootPath, "DeferredRelay.json");
+    protected override UnhandledExceptionPolicy ExceptionPolicy => DeferredRelaySettings.Default.ExceptionPolicy;
+
+    protected override void Configure() =>
+        DeferredRelaySettings.Load(GetConfiguration());
+
+    protected override void OnSystemLoaded(NeoSystem system)
+    {
+        if (system.Settings.Network != DeferredRelaySettings.Default.Network)
+            return;
```
shargon, line 46: Remove from config, and use only the current one

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -469,7 +470,10 @@ public async void Start(CommandLineOptions options)
 
     public void Stop()
     {
-        Interlocked.Exchange(ref _neoSystem, null)?.Dispose();
+        var sys = Interlocked.Exchange(ref _neoSystem, null);
+        if (sys is null)
+            return;
+        sys.Dispose();
```
shargon: What's the difference?

plugins/DeferredRelay/DeferredRelayPlugin.cs
superboyiii: I just follow the style from other plugins. For example:
https://github.com/neo-project/neo-node/blob/273d7296e2bdcba7950d499b97a33dc8b77b2d93/plugins/DBFTPlugin/DBFTPlugin.cs#L62
https://github.com/neo-project/neo-node/blob/273d7296e2bdcba7950d499b97a33dc8b77b2d93/plugins/StateService/StatePlugin.cs#L64
We all know actually, in current code, there're almost no requirement for multi network plugins in current neo system, but all plugins are still on that way. I just keep it consistent with the others. It's not bad, in case a user install a testnet plugin on mainnet node, then it will not really relay that. It's more safe than only default neosystem.

src/Neo.CLI/CLI/MainService.cs
superboyiii: No difference. Oops. Reverted.

plugins/RpcServer/RpcServer.Wallet.cs
```diff
@@ -943,7 +944,7 @@ private JObject SignAndRelay(DataCache snapshot, Transaction tx)
         if (context.Completed)
         {
             tx.Witnesses = context.GetWitnesses();
-            system.Blockchain.Tell(tx);
+            _ = system.Blockchain.Ask<Blockchain.RelayResult>(tx, TimeSpan.FromSeconds(30)).Result;
```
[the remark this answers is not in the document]
superboyiii: Yes, that's a bug. Fixed.

plugins/DeferredRelay/DeferredRelayPlugin.cs
ajara87: if all plugins follow the same way, I think we should apply to this one

ajara87 (dismissed):

plugins/DeferredRelay/DeferredRelayEngine.cs
```diff
@@ -0,0 +1,198 @@
...
+            root["count"] = 0;
+            return root;
+        }
+
+        root["enabled"] = true;
+        root["pendingcheckfrequency"] = settings.CheckFrequency;
+        root["pendingrelaymaxtransactions"] = settings.MaxTransactions;
+
+        var arr = (JArray)root["pending"]!;
+        foreach ((byte[] key, byte[] value) in store.Find())
```
Wi1l-B0t, line 55: Scan all key-value entries ?

plugins/DeferredRelay/DeferredRelayEngine.cs
```diff
@@ -0,0 +1,198 @@
...
+            return false;
+        }
+    }
+
+    private static int CountEntries(IStore store)
+    {
+        int n = 0;
+        foreach ((byte[] key, _) in store.Find())
+        {
+            if (key.Length == UInt256.Length) n++;
```
Wi1l-B0t, line 241: Why is the key length of the entry 32 bytes? Is this an entry of a transaction?

superboyiii: > Why is the key length of the entry 32 bytes? Is this an entry of a transaction?

Yes, the key of `NotYetValid ` transaction is `txid` itself.

plugins/DeferredRelay/DeferredRelayEngine.cs
superboyiii: Yes, `GetPendingState` only backs the `getpendingvaliduntilrelay` RPC and the `list pending` CLI command. The entry count is already capped by `settings.MaxTransactions` at write time in `TryOffer`. Although seems it's heavy on cost, it's not a frequent use case (people will not frequently send future tx on his own node, he knows what he wants and it's not a public service).

plugins/DeferredRelay/DeferredRelayEngine.cs
superboyiii: Add some notes.

superboyiii:
Add `getrawpendingtx`, so we can get details of specific `NotYetValid` tx like what `getrawtransaction` does.

f25c2cdface79ef6e3cadd82391f7880
1c855dfc9927fab42d9b7e91107c691a

plugins/DeferredRelay/DeferredRelayPlugin.cs
shargon: Ok, I will clean it in a different PR

plugins/DeferredRelay/DeferredRelayPlugin.cs
```diff
@@ -0,0 +1,154 @@
...
+            tx = raw.AsSerializable<Transaction>();
+        }
+        catch
+        {
+            throw new RpcException(RpcError.UnknownTransaction.WithData($"Corrupt entry for {hash}."));
+        }
+
+        var json = tx.ToJson(_neoSystem.Settings);
+        var snapshot = _neoSystem.StoreView;
+        if (NativeContract.Ledger.ContainsBlock(snapshot, _neoSystem.GenesisBlock.Hash))
```
shargon: Check genesis is not required

plugins/DeferredRelay/DeferredRelayEngine.cs
```diff
@@ -0,0 +1,220 @@
...
+
+internal static class DeferredRelayEngine
+{
+    /// <summary>
+    /// Builds a snapshot of locally queued NotYetValid transactions and plugin settings (RPC/CLI listing).
+    /// </summary>
+    public static JObject GetPendingState(NeoSystem system, IStore? store, DeferredRelaySettings settings)
+    {
+        var snapshot = system.StoreView;
+        bool ledgerReady = NativeContract.Ledger.ContainsBlock(snapshot, system.GenesisBlock.Hash);
```
shargon: Genesis is always persisted

plugins/DeferredRelay/DeferredRelayEngine.cs
```diff
@@ -0,0 +1,220 @@
...
+
+        if (CountEntries(store) >= settings.MaxTransactions)
+            return false;
+
+        using var ms = new MemoryStream();
+        using (var writer = new BinaryWriter(ms))
+        {
+            ((ISerializable)tx).Serialize(writer);
+        }
+        store.Put(key, ms.ToArray());
```
shargon: tx.ToArray() is easier

plugins/DeferredRelay/DeferredRelayEngine.cs
```diff
@@ -0,0 +1,220 @@
...
+        //   value = serialized Transaction (ISerializable blob)
+        // The length filter is a schema guard: any key whose length is not UInt256.Length
+        // is not a transaction (e.g. future metadata keys with a different prefix/length,
+        // or externally-injected entries) and must be ignored, not counted or deserialized.
+        int n = 0;
+        foreach ((byte[] key, _) in store.Find())
+        {
+            if (key.Length == UInt256.Length) n++;
+        }
+        return n;
```
shargon, line 243: This result should be catched, and if there is no new tx, avoid storage access

superboyiii:
@shargon Thanks to your suggestion. All improved.

ajara87 (approved):

vncoelho (commented):
LGTM as well, I will test along this next week

Wi1l-B0t (approved):

shargon (approved):

vncoelho:
> LGTM as well, I will test along this next week

I was still on tests.

plugins/DeferredRelay/DeferredRelayActor.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    protected override void PostStop() =>
+        Context.System.EventStream.Unsubscribe(Self);
+
+    protected override void OnReceive(object message)
+    {
+        switch (message)
+        {
+            case Blockchain.RelayResult { Inventory: Transaction tx, Result: VerifyResult.NotYetValid } rr:
+                DeferredRelayEngine.TryOffer(_neo, _store, _settings, tx, rr.Result, _counter);
```
shargon, line 53: @superboyiii  We should check that the address have gas for it.
We need a whitelist. We should allow only certains address, people should have this queue and relay his own transactions.

Otherwise how they deal with spam?

PR #1048, opened 2026-05-12
NEP-23 compatible

shargon:
This PR preserve nep-23 until approve and merge https://github.com/neo-project/proposals/pull/231

Require https://github.com/neo-project/proposals/pull/231

### DIFF
--- a/plugins/RpcServer/RpcServer.Node.cs
+++ b/plugins/RpcServer/RpcServer.Node.cs
@@ -99,6 +99,7 @@ partial class RpcServer
             case VerifyResult.OverSize:
                 throw new RpcException(RpcError.InvalidSize.WithData(reason.ToString()));
             case VerifyResult.Expired:
+            case VerifyResult.NotYetValid:
                 throw new RpcException(RpcError.ExpiredTransaction.WithData(reason.ToString()));
             case VerifyResult.InsufficientFunds:
                 throw new RpcException(RpcError.InsufficientFunds.WithData(reason.ToString()));
--- a/plugins/RpcServer/RpcServer.Node.cs
+++ b/plugins/RpcServer/RpcServer.Node.cs
@@ -99,7 +99,7 @@ partial class RpcServer
             case VerifyResult.OverSize:
                 throw new RpcException(RpcError.InvalidSize.WithData(reason.ToString()));
             case VerifyResult.Expired:
-            case VerifyResult.NotYetValid:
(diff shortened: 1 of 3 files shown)


superboyiii:
@shargon Don't forget to update ./plugins/Directory.Build.props

plugins/RpcServer/RpcServer.Node.cs
```diff
@@ -99,6 +99,7 @@ private static JObject GetRelayResult(VerifyResult reason, UInt256 hash)
             case VerifyResult.OverSize:
                 throw new RpcException(RpcError.InvalidSize.WithData(reason.ToString()));
             case VerifyResult.Expired:
+            case VerifyResult.NotYetValid: // TODO: return different value when https://github.com/neo-project/proposals/pull/231 merged
                 throw new RpcException(RpcError.ExpiredTransaction.WithData(reason.ToString()));
```
superboyiii, line 103: I think you mean this
```suggestion
            case VerifyResult.Expired:
                   throw new RpcException(RpcError.ExpiredTransaction.WithData(reason.ToString()));
            case VerifyResult.NotYetValid: // TODO: return different value when https://github.com/neo-project/proposals/pull/231 merged
                   throw new RpcException(RpcError.NotYetValidTransaction.WithData(reason.ToString()));
```

ajara87: agree

shargon: No, we need to be compatible with the current nep, until the next one is merged

superboyiii: > No, we need to be compatible with the current nep, until the next one is merged

Current NEP throws reason code in Expired. But now you deleted it.

shargon: Before: Expired and NotYetValid throw Expired error code.
Now: Expired and NotYetValid throw Expired error code.

Nothing changed. It will change when NEP will be merged, now we need to throw the same error

superboyiii: > Before: Expired and NotYetValid throw Expired error code. Now: Expired and NotYetValid throw Expired error code.
> 
> Nothing changed. It will change when NEP will be merged, now we need to throw the same error

I see. My fault.

Jim8y (approved):
Looks good to me. The Neo package version is updated consistently for CLI and plugins, and the temporary NotYetValid mapping preserves the current RPC error behavior until the NEP-23 update defines the dedicated response.

ajara87 (approved):

PR #1049, opened 2026-05-20
Preventing redirection from HTTPS to non-HTTPS in Oracle

erikzhang:


plugins/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -61,6 +61,8 @@ public void Dispose()
                 if (message.Headers.Location is not null)
                 {
                     uri = message.Headers.Location;
+                    if (uri.Scheme != "https")
```
ajara87:
```suggestion
                    if (uri.Scheme != Uri.UriSchemeHttps)
```

shargon: Ignore case

ajara87: Uri.Scheme is already normalized so case-insensitive comparison shouldn't be needed here. Using `Uri.UriSchemeHttps` would make it clearer

plugins/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -61,6 +61,8 @@ public void Dispose()
                 if (message.Headers.Location is not null)
                 {
                     uri = message.Headers.Location;
+                    if (uri.Scheme != Uri.UriSchemeHttps)
```
shargon, line 64: @erikzhang But we allow http from the beginning, isn't it? We ahould check if it's a downgrade

erikzhang: https://github.com/neo-project/neo-node/blob/87bcd22a11be9341afb6f38507240fe0e6cb6ead/plugins/OracleService/OracleService.cs#L154

We allow https only.

ajara87 (approved):

shargon (approved):

erikzhang:
Port to N3?

ajara87:
> Port to N3?
Done

PR #1050, opened 2026-05-21
[N3] Preventing redirection from HTTPS to non-HTTPS in Oracle

ajara87:
Port https://github.com/neo-project/neo-node/pull/1049


erikzhang:
Need to check the states.

ajara87:
> Need to check the states.

Do you mean checking whether the StatusCode is in the 3xx range?

shargon (approved):

shargon:
> Need to check the states.

This is for new queries, so states are not affected

roman-khimov (approved):
Had a thought on making it Gorgon-dependent since technically it's a behavior change. But then even if we have mixed plugins on the network they should fall back to `ConsensusUnreachable` and then request is finished and resulting transaction can't be changed once it's in, so I don't see any way to exploit this.

PR #1052, opened 2026-05-28
Remove network from plugins

### LINKED ISSUE #1051: Remove Network from plugins configurations and use `ProtocolSettings.Default`
This issue tracks the action point raised in the discussion [discussion_r3309930202](https://github.com/neo-project/neo-node/pull/1047#discussion_r3309930202) within PR [#1047](https://github.com/neo-project/neo-node/pull/1047). 

**Recommendation:**
Remove the Network setting from all plugins configurations and use `ProtocolSettings.Default`, as there is no multi-network support in any plugin.

**PR Context:**
PR #1047 introduces the DeferredRelay plugin, allowing nodes to locally queue NotYetValid transactions and retry relay later.

If further details are needed, please consult the original PR conversation.

shargon:
Close https://github.com/neo-project/neo-node/issues/1051


shargon (commented):
@superboyiii could you test it RpcServer, it was the hardest one

vncoelho (commented):
Lgtm, i will test asap

vncoelho:
@shargon , could you, if possible, update your own forked repository from "neo-cli" to "neo-node", this will help my script here...
But it is a minor thing but causes some problems because some folders also change the name.

shargon:
> @shargon , could you, if possible, update your own forked repository from "neo-cli" to "neo-node", this will help my script here...
> But it is a minor thing but causes some problems because some folders also change the name.

Done

plugins/RpcServer/RpcServerPlugin.cs
```diff
@@ -67,7 +81,8 @@ protected override void OnSystemLoaded(NeoSystem system)
         }
 
         rpcRpcServer.StartRpcServer();
-        servers.TryAdd(s.Network, rpcRpcServer);
+        servers.TryAdd(system.Settings.Network, rpcRpcServer);
+        serverIndex++;
```
AnnaShaleva, line 85: Before this patch, it was allowed to setup multiple RPC servers (one server per network). I think that right now, when multi-network support is removed from plugins, it should be possible to setup only a single RPC server (for a single network).

We should convert `RpcServerPlugin.servers` dictionary to `private static readonly RpcServer server` and adjust the corresponding logic accordingly.

shargon: If in the future we have multiple network, we can create multiple instances of the plugin initialized with different neosystem. I want to be compatible with old configuration files...

vncoelho (dismissed):

superboyiii:
@shargon UT failed.

superboyiii:
Since [DeferredRelay](https://github.com/neo-project/neo-node/tree/master-n3/plugins/DeferredRelay) is merged, need remove that network id as well.

superboyiii:
Tested OK. Plugins work well. Just need modify [DeferredRelay](https://github.com/neo-project/neo-node/tree/master-n3/plugins/DeferredRelay).

AnnaShaleva (commented):
@shargon please, check the https://github.com/shargon/neo-node/pull/3. Other than that LGTM.

To be honest, I thought about network removal a lot of times. It makes it easier to use the same build of NeoCLI for mainnet/testnet, you don't need to reconfigure every plugin manually on every network switch. I like this change.

AnnaShaleva (approved):

ajara87 (approved):

vncoelho:
Let me test again, 2 days

erikzhang:
In fact, the initial addition of "network" was intended to enable Node to support running multiple networks simultaneously within a single process in the future.

shargon:
> In fact, the initial addition of "network" was intended to enable Node to support running multiple networks simultaneously within a single process in the future.

The best way to archive that is have multiple NeoSystem, and load multiple instances of this plugin with this NeoSystem.

vncoelho (approved):
Neo node tests with expect on tests has been helping a lot.

Everything has been more consistent now for building and running.

I will extend the tests for the release itself.

vncoelho:
Port for n4

PR #1054, opened 2026-05-28
Ensure that the transaction sent was with valid size

shargon:
This pull request adds improved validation and error handling for transaction signing and relaying in the `MainService.Wallet.cs` file. The key changes are grouped below:

**Transaction Validation and Error Handling Improvements:**

* Added a check to ensure transactions do not exceed `Transaction.MaxTransactionSize`, displaying an error and aborting if they do.
* Enhanced feedback on transaction relay status: after attempting to relay a transaction, the code now checks the result and prints a success or error message depending on whether the relay succeeded.

**Code Maintenance:**

* Added a `using static Neo.Ledger.Blockchain;` directive to simplify qualified references in the file.

### DIFF
--- a/src/Neo.CLI/CLI/MainService.Wallet.cs
+++ b/src/Neo.CLI/CLI/MainService.Wallet.cs
@@ -975,7 +975,7 @@ partial class MainService
         if (context.Completed)
         {
             tx.Witnesses = context.GetWitnesses();
-            var reason = NeoSystem.Blockchain.Ask<RelayResult>(tx).Result;
+            var reason = NeoSystem.Blockchain.Ask<RelayResult>(tx, TimeSpan.FromSeconds(30)).Result;
 
             if (reason.Result == Ledger.VerifyResult.Succeed)
             {
--- a/src/Neo.CLI/CLI/MainService.Wallet.cs
+++ b/src/Neo.CLI/CLI/MainService.Wallet.cs
@@ -975,7 +975,6 @@ partial class MainService
         if (context.Completed)
         {
             tx.Witnesses = context.GetWitnesses();
-            NeoSystem.Blockchain.Tell(tx);
             var reason = NeoSystem.Blockchain.Ask<RelayResult>(tx).Result;
 
             if (reason.Result == Ledger.VerifyResult.Succeed)
--- a/src/Neo.CLI/CLI/MainService.Wallet.cs
+++ b/src/Neo.CLI/CLI/MainService.Wallet.cs
@@ -27,6 +27,7 @@ using Neo.Wallets.NEP6;
 using System.Numerics;
 using System.Security.Cryptography;
 using System.Text;
+using static Neo.Ledger.Blockchain;
 using static Neo.SmartContract.Helper;
 using ECCurve = Neo.Cryptography.ECC.ECCurve;
 using ECPoint = Neo.Cryptography.ECC.ECPoint;
@@ -965,12 +966,25 @@ partial class MainService
             ConsoleHelper.Error("Failed creating contract params: " + GetExceptionMessage(e));
             throw;
         }
+        if (tx.Size > Transaction.MaxTransactionSize)
+        {
+            ConsoleHelper.Error($"Transaction size exceeds the maximum allowed size of {Transaction.MaxTransactionSize} bytes.");
+            return;
+        }
         CurrentWallet!.Sign(context);
         if (context.Completed)
         {
             tx.Witnesses = context.GetWitnesses();
             NeoSystem.Blockchain.Tell(tx);
-            ConsoleHelper.Info("Signed and relayed transaction with hash:\n", $"{tx.Hash}");
+            var reason = NeoSystem.Blockchain.Ask<RelayResult>(tx).Result;
+
+            if (reason.Result == Ledger.VerifyResult.Succeed)
+            {
+                ConsoleHelper.Info("Signed and relayed transaction with hash:\n", $"{tx.Hash}");
+                return;
+            }
+
+            ConsoleHelper.Error($"Failed to relay transaction with hash {tx.Hash}. Status: {reason.Result}");
         }
         else
         {
--- a/plugins/RpcServer/RpcServer.Node.cs
+++ b/plugins/RpcServer/RpcServer.Node.cs
@@ -211,7 +211,6 @@ partial class RpcServer
         var tx = Result.Ok_Or(
             () => Convert.FromBase64String(base64Tx).AsSerializable<Transaction>(),
             RpcError.InvalidParams.WithData($"Invalid Transaction Format: {base64Tx}"));
-
         var reason = system.Blockchain.Ask<RelayResult>(tx).Result;
         return GetRelayResult(reason.Result, tx.Hash);
     }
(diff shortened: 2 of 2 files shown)


shargon:
@superboyiii please check this pr

ajara87:
Could you add UTs for codecov?

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -965,12 +966,25 @@ private void SignAndSendTx(DataCache snapshot, Transaction tx)
...
+        }
         CurrentWallet!.Sign(context);
         if (context.Completed)
         {
             tx.Witnesses = context.GetWitnesses();
             NeoSystem.Blockchain.Tell(tx);
-            ConsoleHelper.Info("Signed and relayed transaction with hash:\n", $"{tx.Hash}");
+            var reason = NeoSystem.Blockchain.Ask<RelayResult>(tx).Result;
+
+            if (reason.Result == Ledger.VerifyResult.Succeed)
```
AnnaShaleva, line 980: This changes the behaviour of the command. Are you sure that we'd like to wait for the verification result here? Is it expected for the end-user? If so, then we should probably adjust other cases in Neo.CLI where `NeoSystem.Blockchain.Tell` was used (`OnRelayCommand` etc.).

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -965,12 +966,25 @@ private void SignAndSendTx(DataCache snapshot, Transaction tx)
...
+            ConsoleHelper.Error($"Transaction size exceeds the maximum allowed size of {Transaction.MaxTransactionSize} bytes.");
+            return;
+        }
         CurrentWallet!.Sign(context);
         if (context.Completed)
         {
             tx.Witnesses = context.GetWitnesses();
             NeoSystem.Blockchain.Tell(tx);
-            ConsoleHelper.Info("Signed and relayed transaction with hash:\n", $"{tx.Hash}");
+            var reason = NeoSystem.Blockchain.Ask<RelayResult>(tx).Result;
```
AnnaShaleva: Another question is: if we use `Ask` here, shouldn't we remove `NeoSystem.Blockchain.Tell` two lines above? I thought that `Ask` invokes `Tell` by itself:
https://github.com/akkadotnet/akka.net/blob/6938c721a143261561d0ce41bc895862eab6085d/src/core/Akka/Actor/Futures.cs#L167
https://github.com/akkadotnet/akka.net/blob/6938c721a143261561d0ce41bc895862eab6085d/src/core/Akka/Actor/ActorRef.cs#L202

AnnaShaleva (commented):
The size check itself is OK to me.

src/Neo.CLI/CLI/MainService.Wallet.cs
shargon: How he will know if was relayed?

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -965,12 +966,25 @@ private void SignAndSendTx(DataCache snapshot, Transaction tx)
...
+        if (tx.Size > Transaction.MaxTransactionSize)
+        {
+            ConsoleHelper.Error($"Transaction size exceeds the maximum allowed size of {Transaction.MaxTransactionSize} bytes.");
+            return;
+        }
         CurrentWallet!.Sign(context);
         if (context.Completed)
         {
             tx.Witnesses = context.GetWitnesses();
             NeoSystem.Blockchain.Tell(tx);
```
shargon:
```suggestion
```

src/Neo.CLI/CLI/MainService.Wallet.cs
shargon: Fixed!

src/Neo.CLI/CLI/MainService.Wallet.cs
AnnaShaleva: OK, I'm not against of it, but still, it's a bahaviour change.

Let's then update other commands that use the same pattern. Behaviour should be unified across CLI.

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -965,12 +966,24 @@ private void SignAndSendTx(DataCache snapshot, Transaction tx)
...
+            ConsoleHelper.Error($"Transaction size exceeds the maximum allowed size of {Transaction.MaxTransactionSize} bytes.");
+            return;
+        }
         CurrentWallet!.Sign(context);
         if (context.Completed)
         {
             tx.Witnesses = context.GetWitnesses();
-            NeoSystem.Blockchain.Tell(tx);
-            ConsoleHelper.Info("Signed and relayed transaction with hash:\n", $"{tx.Hash}");
+            var reason = NeoSystem.Blockchain.Ask<RelayResult>(tx).Result;
```
ajara87:
```suggestion
            var reason = NeoSystem.Blockchain.Ask<RelayResult>(tx, TimeSpan.FromSeconds(30)).Result;
```

ajara87 (approved):

src/Neo.CLI/CLI/MainService.Wallet.cs
vncoelho: i agree with @AnnaShaleva , historically it has not been like this

shargon: > it's a bahaviour change.

Like all the code that we change 😅

This doesn't wait to be persisted in a block, it wait for verification, state independant one, so it will be fast and you will know if there was any error, currently, you can't

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -965,12 +966,24 @@ private void SignAndSendTx(DataCache snapshot, Transaction tx)
...
+            ConsoleHelper.Error($"Transaction size exceeds the maximum allowed size of {Transaction.MaxTransactionSize} bytes.");
+            return;
+        }
         CurrentWallet!.Sign(context);
         if (context.Completed)
         {
             tx.Witnesses = context.GetWitnesses();
-            NeoSystem.Blockchain.Tell(tx);
-            ConsoleHelper.Info("Signed and relayed transaction with hash:\n", $"{tx.Hash}");
+            var reason = NeoSystem.Blockchain.Ask<RelayResult>(tx, TimeSpan.FromSeconds(30)).Result;
```
Wi1l-B0t, line 978: Why add this line?

ajara87: `Tell` is fire-and-forget in `Akka.NET` it gives no feedback. `Ask<RelayResult>` waits for the Blockchain actor's response so we can report back to the user whether the relay actually succeeded or failed, instead of always printing success unconditionally

shargon: It was shown relayed, and maybe is not

AnnaShaleva (approved):

shargon:
Already merged in https://github.com/neo-project/neo-node/pull/1047/changes#diff-9b3a6b4aa212aa1abf6257290529ba305a63d93551d49ae55d7e62f102f96807R975

PR #1055, opened 2026-05-28
Ensure max size in Witness

shargon:
Require https://github.com/neo-project/neo/pull/4555

This pull request introduces validation to enforce maximum allowed lengths for invocation and verification scripts in witnesses, improving security and robustness in the RPC server. It also adds comprehensive unit tests to ensure these constraints are properly enforced.

**Witness Script Length Validation:**

* Added constants `MaxVerificationScriptLength` and `MaxInvocationScriptLength` (both set to 1024) in `ParameterConverter.cs` to define allowed script sizes.
* Updated the `ToWitnesses` method to throw an `RpcException` if either the invocation or verification script exceeds its respective maximum length, providing a clear error message with the offending index.

**Testing:**

* Added a new test file `UT_ParameterConverter_WitnessSize.cs` with extensive unit tests covering valid and invalid script lengths, including edge cases and multiple witnesses, to verify the new validation logic.


plugins/RpcServer/ParameterConverter.cs
```diff
@@ -305,11 +307,18 @@ private static Witness[] ToWitnesses(this JArray json)
...
             {
                 InvocationScript = Convert.FromBase64String(invocation?.AsString() ?? string.Empty),
                 VerificationScript = Convert.FromBase64String(verification?.AsString() ?? string.Empty)
-            });
+            };
+
+            if (witness.InvocationScript.Length > MaxInvocationScriptLength)
+                throw new RpcException(RpcError.InvalidParams.WithData($"Invocation script length exceeds the maximum allowed length at {i}."));
+            if (witness.VerificationScript.Length > MaxVerificationScriptLength)
+                throw new RpcException(RpcError.InvalidParams.WithData($"Verification script length exceeds the maximum allowed length at {i}."));
```
AnnaShaleva, line 322: Can't use these constants, it's too restrictive. We may deal with custom witnesses here (we use them in NeoFS-derived transactions already). Related to https://github.com/neo-project/neo/pull/3385#discussion_r1665873398 and https://github.com/neo-project/neo/pull/3209.

AnnaShaleva: Ref. https://github.com/neo-project/neo/issues/3539.

shargon: We should change the constants, but currently these are the values

AnnaShaleva: Which values do you suggest? 1024 is too small.

plugins/RpcServer/ParameterConverter.cs
```diff
@@ -24,6 +24,8 @@ namespace Neo.Plugins.RpcServer;
 public static class ParameterConverter
 {
     private static readonly Dictionary<Type, Func<JToken, object>> s_conversions;
+    private const int MaxVerificationScriptLength = 1024;
+    private const int MaxInvocationScriptLength = 1024;
```
ajara87:
```suggestion
    internal const int MaxVerificationScriptLength = 1024;
    internal const int MaxInvocationScriptLength = 1024;
```
I suggest changing it to internal and using it in the unit tests so they use the real value

plugins/RpcServer/ParameterConverter.cs
shargon: I think we can use MaxScriptSize with the sum of all of the witnesses, and remove this one

Jim8y: I agree with this direction. The RPC converter should not apply the serialized Witness 1024-byte per-field limit here; these parameters are also used for invoke/verification flows where custom witnesses need to be accepted. An aggregate bound over the supplied witness scripts would keep the resource limit without rejecting valid custom witness cases.

shargon: This custom witnesses must be valid and serializables, so we need the same limits

shargon:
Require https://github.com/neo-project/neo/pull/4555 to pass UT

ajara87 (dismissed):

plugins/RpcServer/ParameterConverter.cs
AnnaShaleva: > 1024 is too small.

I wasn't aware that 1024 was [a default](https://github.com/neo-project/neo/pull/4555#discussion_r3323160760) for a long time. Then the previous version with `MaxInvocationScriptLength` and `MaxVerificationScriptLength` set to 1024 is OK to me.

Wi1l-B0t (dismissed):

AnnaShaleva (commented):
@shargon let's fetch public constants from https://github.com/neo-project/neo/pull/4554. After that LGTM.

tests/Neo.Plugins.RpcServer.Tests/UT_ParameterConverter_WitnessSize.cs
```diff
@@ -0,0 +1,334 @@
...
+            }
+        };
+
+        // Act & Assert
+        Assert.ThrowsExactly<RpcException>(
+            () => json.ToSignersAndWitnesses(_addressVersion));
+    }
+
+    [TestMethod]
+    public void TestWitnessSize_OneByteOverLimit_VerificationScript()
```
AnnaShaleva, line 314: Nitpicking: these tests may be squashed into a single one with a list of input data, no need to duplicate Arrange/Assert sections. E.g.: https://github.com/nspcc-dev/neo-go/blob/c41054fb7c40e6312e227e49f7cc3f148f113495/cli/cmdargs/parser_test.go#L21

It will be easier to review and track them, less code duplication.

AnnaShaleva (approved):

PR #1057, opened 2026-06-01
Fix known docker issue

superboyiii:
Fix two Docker build/runtime issues for neo-cli: 
1. Shell scripts checked out with Windows `CRLF` line endings broke `prepare-node.sh` inside the Linux container (causing apt/wget failures and shell syntax errors).
2. `start.sh` was launching neo-cli with stdout redirected to a log file, which disabled its interactive TTY mode so `screen -r neo` couldn't accept any commands.

PR #1062, opened 2026-06-02
[N4] Port Docker use to N4

ajara87:
# Port PR related to 
https://github.com/neo-project/neo-node/pull/992
https://github.com/neo-project/neo-node/pull/1057

# Summary
- CLI docker for common usages.
- Fix two Docker build/runtime issues for neo-cli:

  - Shell scripts checked out with Windows CRLF line endings broke prepare-node.sh inside the Linux container (causing apt/wget failures and shell syntax errors).
  - start.sh was launching neo-cli with stdout redirected to a log file, which disabled its interactive TTY mode so screen -r neo couldn't accept any commands.

### DIFF
--- a/src/Neo.CLI/Docker/Dockerfile
+++ b/src/Neo.CLI/Docker/Dockerfile
@@ -1,0 +1,33 @@
+FROM mcr.microsoft.com/dotnet/aspnet:10.0 AS Build
+RUN apt-get update && apt-get install -y \
+    net-tools \
+    telnet \
+    bash-completion \
+    wget \
+    curl \
+    lrzsz \
+    zip \
+    unzip \
+    sqlite3 \
+    libsqlite3-dev \
+    libunwind8-dev \
+    screen \
+    dos2unix \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /neo
+COPY prepare-node.sh .
+RUN dos2unix prepare-node.sh && chmod +x prepare-node.sh
+
+ARG CLI_VERSION=v3.9.2
+ARG PLUGIN_VERSION=
+RUN if [ -z "$PLUGIN_VERSION" ]; then \
+        ./prepare-node.sh $CLI_VERSION; \
+    else \
+        ./prepare-node.sh $CLI_VERSION $PLUGIN_VERSION; \
+    fi
+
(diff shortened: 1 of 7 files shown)


src/Neo.CLI/Docker/README.md
```diff
@@ -0,0 +1,52 @@
+# Neo CLI Docker Usage Examples
+
+## Example 1: Install CLI v3.9.2 (with matching plugins)
```
shargon: These version should be changed

ajara87: Since N4 has not been released yet, what version should we use?

shargon: 4.0.0

src/Neo.CLI/Docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    libunwind8-dev \
+    screen \
+    dos2unix \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /neo
+COPY prepare-node.sh .
+RUN dos2unix prepare-node.sh && chmod +x prepare-node.sh
+
+ARG CLI_VERSION=v3.9.2
```
AnnaShaleva: Should also update version.

src/Neo.CLI/Docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM mcr.microsoft.com/dotnet/aspnet:10.0 AS Build
```
AnnaShaleva: Remove `AS Build`, it's a single base image that is used as a runtime. Usually `AS Build`and `AS Runtime` are used to distinguish images for build/run.

src/Neo.CLI/Docker/README.md
ajara87: done

src/Neo.CLI/Docker/Dockerfile
ajara87: done

src/Neo.CLI/Docker/Dockerfile
ajara87: done

AnnaShaleva (approved):
Good for now. Once v4.0.0 is released, we'll be able to test it and fix bugs (if any).

shargon (approved):

shargon:
Agree with @AnnaShaleva

PR #1070, opened 2026-06-09
Improve DeferredRelay

superboyiii:
Fix known issue in `DeferredRelay` and improve some verification workflow.
1. `NotYetValid` early return — Queued txs previously skipped `Policy/funds/witness` checks; we now run `state-dependent` verification before `store.Put`, skipping only the time gates.
2. Pre-offer filtering — `VerifyForOffer` mirrors core `VerifyStateDependent` (Policy, funds, network fee, witness) so far-future/no-GAS txs are rejected at enqueue time.
3. `Per-sender` cap — `MaxTransactionsPerSender` limits how many deferred txs one sender can queue.
4. Sender fee aggregation — DeferredQueueContext sums queued fees per sender so multiple queued txs cannot exceed balance individually.
5. Relay failure eviction — Definitive relay failures (PolicyFail, Invalid*, OverSize, InsufficientFunds) remove entries instead of retrying until expiry.
6. Min network fee — `MinNetworkFee` (In `GAS`, same format as RpcServer `MaxFee`) rejects low-fee spam before enqueue.

### DIFF
--- a/plugins/DeferredRelay/DeferredQueueContext.cs
+++ b/plugins/DeferredRelay/DeferredQueueContext.cs
@@ -1,0 +1,102 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// DeferredQueueContext.cs file belongs to the neo project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using Neo.Extensions;
+using Neo.Ledger;
+using Neo.Network.P2P.Payloads;
+using Neo.Persistence;
+
+namespace Neo.Plugins.DeferredRelay;
+
+/// <summary>
+/// Tracks per-sender queue depth and builds a <see cref="TransactionVerificationContext"/> that
+/// includes already-queued transactions (fee/oracle aggregation for deferred offers).
+/// </summary>
+internal sealed class DeferredQueueContext
+{
+    private readonly object _lock = new();
+    private readonly Dictionary<UInt160, int> _senderCounts = new();
+    private readonly List<Transaction> _queued = new();
+
+    public static DeferredQueueContext Bootstrap(IStore store)
+    {
+        var context = new DeferredQueueContext();
+        foreach ((byte[] key, byte[] value) in store.Find())
+        {
+            if (key.Length != UInt256.Length) continue;
+            try
+            {
+                context.OnQueued(value.AsSerializable<Transaction>());
+            }
+            catch
+            {
+                // Corrupt entries are ignored here; ProcessQueuedAsync removes them later.
+            }
+        }
+
+        return context;
+    }
+
+    public bool CanAcceptSender(UInt160 sender, uint maxPerSender)
+    {
+        if (maxPerSender == 0) return true;
+        lock (_lock)
+        {
+            _senderCounts.TryGetValue(sender, out int count);
+            return count < maxPerSender;
+        }
+    }
+
+    public TransactionVerificationContext CreateVerificationContext()
+    {
+        var context = new TransactionVerificationContext();
+        lock (_lock)
+        {
+            foreach (Transaction tx in _queued)
+                context.AddTransaction(tx);
+        }
+
+        return context;
+    }
+
+    public void OnQueued(Transaction tx)
+    {
+        lock (_lock)
+        {
+            _queued.Add(tx);
+            _senderCounts.TryGetValue(tx.Sender, out int count);
+            _senderCounts[tx.Sender] = count + 1;
+        }
+    }
+
+    public void OnRemoved(Transaction tx)
+    {
+        lock (_lock)
+        {
+            for (int i = _queued.Count - 1; i >= 0; i--)
+            {
+                if (_queued[i].Hash == tx.Hash)
+                {
+                    _queued.RemoveAt(i);
+                    break;
+                }
+            }
+
+            if (_senderCounts.TryGetValue(tx.Sender, out int count))
+            {
+                if (count <= 1)
+                    _senderCounts.Remove(tx.Sender);
+                else
+                    _senderCounts[tx.Sender] = count - 1;
+            }
+        }
+    }
+}
--- a/plugins/DeferredRelay/DeferredQueueContext.cs
+++ b/plugins/DeferredRelay/DeferredQueueContext.cs
@@ -24,7 +24,7 @@ internal sealed class DeferredQueueContext
 {
     private readonly Lock _lock = new();
     private readonly Dictionary<UInt160, int> _senderCounts = new();
-    private readonly List<Transaction> _queued = new();
+    private readonly Dictionary<UInt256, Transaction> _queued = new();
 
     public static DeferredQueueContext Bootstrap(IStore store)
     {
@@ -60,7 +60,7 @@ internal sealed class DeferredQueueContext
         var context = new TransactionVerificationContext();
         lock (_lock)
         {
-            foreach (Transaction tx in _queued)
+            foreach (Transaction tx in _queued.Values)
                 context.AddTransaction(tx);
         }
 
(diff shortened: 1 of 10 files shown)


plugins/DeferredRelay/DeferredQueueContext.cs
```diff
@@ -0,0 +1,102 @@
...
+namespace Neo.Plugins.DeferredRelay;
+
+/// <summary>
+/// Tracks per-sender queue depth and builds a <see cref="TransactionVerificationContext"/> that
+/// includes already-queued transactions (fee/oracle aggregation for deferred offers).
+/// </summary>
+internal sealed class DeferredQueueContext
+{
+    private readonly object _lock = new();
+    private readonly Dictionary<UInt160, int> _senderCounts = new();
```
shargon, line 26: This are never cleared, so it's easy to fill. I think it should be easier if we realize a verifyStateDependant before add it

superboyiii: I already added `verifyStateDependant`, I found the issue here is about `Bootstrap`. It reloads all persisted entries without re-checking expiry or current state, so stale txs can occupy per-sender slots and fee aggregation until the next ProcessQueuedAsync. I will fix soon.

superboyiii: > This are never cleared, so it's easy to fill. I think it should be easier if we realize a verifyStateDependant before add it

Now it's solved.

plugins/DeferredRelay/DeferredRelayVerification.cs
```diff
@@ -0,0 +1,191 @@
...
+        VerifyResult.InvalidScript => true,
+        VerifyResult.InvalidAttribute => true,
+        VerifyResult.InvalidSignature => true,
+        VerifyResult.Invalid => true,
+        VerifyResult.OverSize => true,
+        VerifyResult.InsufficientFunds => true,
+        _ => false,
+    };
+
+    internal static VerifyResult VerifyForOffer(
```
shargon, line 44: Why replicate all the logic and not use `Verify`?

superboyiii: > Why replicate all the logic and not use `Verify`?

That's why we need `DeferredRelay`, Verify/VerifyStateDependent will reject `NotYetValid` in neo.

plugins/DeferredRelay/DeferredRelaySettings.cs
```diff
@@ -43,22 +46,53 @@ private DeferredRelaySettings(IConfigurationSection section)
...
+    internal static DeferredRelaySettings Create(uint maxTransactions, uint checkFrequency, uint maxTransactionsPerSender = 0u, long minNetworkFee = 0L) =>
+        new("DeferredRelay_{0}", maxTransactions, maxTransactionsPerSender, minNetworkFee, checkFrequency, UnhandledExceptionPolicy.Ignore);
+
+    /// <summary>
+    /// Parses a GAS-denominated decimal from plugin configuration into datoshi (RpcServer <c>GetGasAmount</c>).
+    /// </summary>
+    private static long GetGasAmount(IConfigurationSection section, string key, long defaultValue)
+    {
+        if (!section.GetSection(key).Exists()) return defaultValue;
+        return (long)new BigDecimal(section.GetValue<decimal>(key), NativeContract.GAS.Decimals).Value;
```
ajara87, line 96: should not be checked to avoid overflow in case of a value configured too big?

plugins/DeferredRelay/DeferredQueueContext.cs
```diff
@@ -0,0 +1,102 @@
...
+
+namespace Neo.Plugins.DeferredRelay;
+
+/// <summary>
+/// Tracks per-sender queue depth and builds a <see cref="TransactionVerificationContext"/> that
+/// includes already-queued transactions (fee/oracle aggregation for deferred offers).
+/// </summary>
+internal sealed class DeferredQueueContext
+{
+    private readonly object _lock = new();
```
Wi1l-B0t: use `System.Threading.Lock` instead ?

plugins/DeferredRelay/DeferredRelaySettings.cs
superboyiii: > should not be checked to avoid overflow in case of a value configured too big?

long.MaxValue = 9,223,372,036,854,775,807,  GAS decimal=8, it's almost impossible that someone input 92,233,720,368.55 GAS as `MinNetworkFee`. And .NET will thorw `OverflowException` if long is overflow. So it's safe.

plugins/DeferredRelay/DeferredQueueContext.cs
superboyiii: Done.

plugins/DeferredRelay/DeferredRelayVerification.cs
shargon: Is not possible to call verify state dependandt only? With reflection maybe

plugins/DeferredRelay/DeferredQueueContext.cs
```diff
@@ -0,0 +1,102 @@
...
+internal sealed class DeferredQueueContext
+{
+    private readonly Lock _lock = new();
+    private readonly Dictionary<UInt160, int> _senderCounts = new();
+    private readonly List<Transaction> _queued = new();
+
+    public static DeferredQueueContext Bootstrap(IStore store)
+    {
+        var context = new DeferredQueueContext();
+        foreach ((byte[] key, byte[] value) in store.Find())
```
shargon, line 32: We don't have any prefix or this store is exclusive for the plugin?

superboyiii: No key prefix: queued txs are stored as key = tx.Hash (32-byte UInt256) in a plugin-dedicated store, only exclusive for the this plugin. Like:
```
├─ApplicationLogs_3480724E
├─Data_LevelDB_3480724E
└─DeferredRelay_3480724E
```

plugins/DeferredRelay/DeferredRelayVerification.cs
superboyiii: > Is not possible to call verify state dependandt only? With reflection maybe

VerifyStateDependent is public and callable directly, but calling it returns NotYetValid early, before the checks we need — so it doesn’t work for our use case. Reflection doesn’t fix that.Another way is creating a proper API in neo core to skip the validity-window gates. This needs only a little code, but apparently, changing Neo only for this doesn't make sense.

shargon (dismissed):

shargon:
@superboyiii  It would be advisable to add a README.md file for this plugin, explaining that it should be used by internal nodes.

plugins/DeferredRelay/DeferredQueueContext.cs
```diff
@@ -0,0 +1,102 @@
...
+
+/// <summary>
+/// Tracks per-sender queue depth and builds a <see cref="TransactionVerificationContext"/> that
+/// includes already-queued transactions (fee/oracle aggregation for deferred offers).
+/// </summary>
+internal sealed class DeferredQueueContext
+{
+    private readonly Lock _lock = new();
+    private readonly Dictionary<UInt160, int> _senderCounts = new();
+    private readonly List<Transaction> _queued = new();
```
ajara87: should not be better use a dictionary instead of list? complex could decrease from O(n) to O(1) in some parts. For example, OnRemoved

vncoelho (dismissed):

superboyiii:
> @superboyiii It would be advisable to add a README.md file for this plugin, explaining that it should be used by internal nodes.

That's good. I will add it soon.

plugins/DeferredRelay/DeferredQueueContext.cs
superboyiii: > should not be better use a dictionary instead of list? complex could decrease from O(n) to O(1) in some parts. For example, OnRemoved

Good catch on OnRemoved

superboyiii:
@shargon @ajara87 All suggestion are implemented.

ajara87 (approved):

shargon (approved):

PR #1072, opened 2026-06-11
v3.10.0

superboyiii:


ajara87 (approved):

src/Neo.CLI/Docker/README.md
```diff
@@ -1,41 +1,41 @@
...
-docker run -d -p 10332:10332 --name neo-node v3.9.2-plugins-3.9.0
+docker run -d -p 10332:10332 --name neo-node v3.10.0-plugins-3.9.2
 
 # Enter the container and test
 docker exec -it neo-node bash
 curl -X POST http://localhost:10332 -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"getblockcount","params":[],"id":1}'
 ```
 
-## Example 3: Install CLI v3.10.0 (with matching plugins)
+## Example 3: Install CLI v3.9.2 with Plugin v3.9.0
```
shargon, line 31: 3.10.0?

superboyiii: > 3.10.0?

It's an example, doesn't need change in every version. Just want to illustrate the usage of `--build-arg`. If installing v3.10.0, then example 1 is enough.

shargon (approved):

PR #1078, opened 2026-06-19
Add NodeDiagnostics monitoring plugin

Jim8y:
## Summary
- Add the NodeDiagnostics plugin for node diagnostics telemetry, crash reporting, node liveness heartbeats, consensus stall alerts, and exception notifications.
- Capture unhandled exceptions and unobserved task exceptions, with immediate bounded flushing for fatal events.
- Report `ConsensusStall` error events when the local block height does not advance for `ConsensusStallThresholdSeconds` seconds; the default threshold is 30 seconds.
- Publish process-level heartbeat checks through Better Stack and Healthchecks-compatible URLs.
- Publish node-level liveness payloads through webhook-compatible status sinks, including block height, header height, and seconds since the local block height last advanced.
- Support configurable sinks for custom webhooks, experimental Sentry-shaped payloads, experimental Google Cloud Error Reporting payloads, Better Stack heartbeats, and Healthchecks heartbeats.
- Allow operators to configure endpoints, tokens, token headers, custom headers, heartbeat intervals, consensus stall threshold, severity thresholds, retries, request timeouts, queue limits, payload truncation limits, and operator tags.
- Enforce HTTPS for token-bearing endpoints and keep tokens out of logs.
- Add an optional startup diagnostic event so operators can validate endpoint, token, header, and payload settings during rollout.
- Add operator-facing documentation with supported providers, sample configuration, security guidance for tokens, rollout validation, provider notes, and node liveness behavior.
- Improve reliability with per-request timeouts, bounded queue drop warnings, retry controls, a `User-Agent` header, bounded worker shutdown, one alert per stalled height, and focused tests for timeout/configuration/payload behavior.
- Rename plugin files, directories, namespaces, configuration file, and test project from NodeOps to NodeDiagnostics.

## Validation
- `dotnet format --no-restore --verify-no-changes --verbosity diagnostic`
- `dotnet test tests/Neo.Plugins.NodeDiagnostics.Tests/Neo.Plugins.NodeDiagnostics.Tests.csproj` (20 passed)
- `dotnet build neo-node.sln`

Note: solution build reports existing NU1903 warnings for SQLitePCLRaw.lib.e_sqlite3 in SQLiteWallet projects.


plugins/NodeDiagnostics/NodeDiagnosticsDispatcher.cs
```diff
@@ -0,0 +1,474 @@
...
+
+    public NodeDiagnosticsDispatcher(NodeDiagnosticsSettings settings, Func<NeoSystem?> systemProvider)
+        : this(settings, systemProvider, new HttpClient(), disposeClient: true)
+    {
+    }
+
+    internal NodeDiagnosticsDispatcher(NodeDiagnosticsSettings settings, HttpClient httpClient, bool disposeClient = false)
+        : this(settings, () => null, httpClient, disposeClient)
+    {
+    }
```
Wi1l-B0t:
```suggestion
        : this(settings, systemProvider, new HttpClient(), disposeClient: true) { }

    internal NodeDiagnosticsDispatcher(NodeDiagnosticsSettings settings, HttpClient httpClient, bool disposeClient = false)
        : this(settings, () => null, httpClient, disposeClient) { }
```

plugins/NodeDiagnostics/NodeDiagnosticsDispatcher.cs
```diff
@@ -0,0 +1,474 @@
...
+    private readonly Channel<NodeDiagnosticsEvent> _channel;
+    private readonly CancellationTokenSource _shutdown = new();
+    private static readonly JsonSerializerOptions SerializerOptions = new(JsonSerializerDefaults.Web);
+    private static readonly TimeSpan ConsensusStallCheckInterval = TimeSpan.FromSeconds(1);
+    private static readonly string UserAgent = $"Neo.NodeDiagnostics/{typeof(NodeDiagnosticsPlugin).Assembly.GetName().Version?.ToString() ?? "unknown"}";
+    private Task? _eventWorker;
+    private Task? _heartbeatWorker;
+    private Task? _consensusStallWorker;
+    private long _droppedEvents;
+    private readonly object _dropWarningLock = new();
```
Wi1l-B0t: use `System.Threading.Lock` instead of `object`?

plugins/NodeDiagnostics/NodeDiagnosticsDispatcher.cs
```diff
@@ -0,0 +1,474 @@
...
+        try
+        {
+            while (!_shutdown.IsCancellationRequested)
+            {
+                await SendConsensusStallIfNeededAsync(DateTimeOffset.UtcNow, _shutdown.Token).ConfigureAwait(false);
+                await Task.Delay(ConsensusStallCheckInterval, _shutdown.Token).ConfigureAwait(false);
+            }
+        }
+        catch (OperationCanceledException)
+        {
```
Wi1l-B0t, line 423: Just ignore?

plugins/NodeDiagnostics/NodeDiagnosticsSettings.cs
```diff
@@ -0,0 +1,199 @@
...
+        ServiceName = section.GetValue("ServiceName", "neo-node")!;
+        NodeName = section.GetValue("NodeName", string.Empty)!;
+        Tags = LoadTags(section.GetSection("Tags"));
+        CaptureUnhandledExceptions = section.GetValue("CaptureUnhandledExceptions", true);
+        CaptureUnobservedTaskExceptions = section.GetValue("CaptureUnobservedTaskExceptions", true);
+        SendStartupDiagnosticEvent = section.GetValue("SendStartupDiagnosticEvent", false);
+        IncludeStackTrace = section.GetValue("IncludeStackTrace", true);
+        HeartbeatIntervalSeconds = section.GetValue("HeartbeatIntervalSeconds", 60);
+        ConsensusStallThresholdSeconds = section.GetValue("ConsensusStallThresholdSeconds", 30);
+        MaxQueueSize = section.GetValue("MaxQueueSize", 1024);
```
cschuchardt88, line 59: **bug:** Default `ConsensusStallThresholdSeconds` of 30 is shorter than one DBFT view-change window. With a 15s `TimePerBlock`, view 1's timer is `TimePerBlock << (view+1)` = 60s, so a healthy node that takes a single view change will emit `ConsensusStall`.

Please raise the default (JSON sample too) to cover at least one view change plus margin, and consider rejecting too-low values in `Validate()` rather than only checking `> 0`.

plugins/NodeDiagnostics/NodeDiagnosticsDispatcher.cs
```diff
@@ -0,0 +1,473 @@
...
+        if (nodeState.BlockHeight is null || nodeState.SecondsSinceLastBlockAdvance is null)
+            return false;
+        if (nodeState.SecondsSinceLastBlockAdvance.Value < _settings.ConsensusStallThresholdSeconds)
+            return false;
+
+        lock (_livenessLock)
+        {
+            if (_lastReportedConsensusStallHeight == nodeState.BlockHeight)
+                return false;
+            _lastReportedConsensusStallHeight = nodeState.BlockHeight;
```
cschuchardt88, line 360: **bug:** This records the stalled height as already reported *before* `SendEventsAsync` runs. If the sink returns 5xx or times out, `ProcessConsensusStallAsync` will not retry that height, so operators can miss the only stall alert for a hung node.

Set `_lastReportedConsensusStallHeight` only after a successful send, or keep it unset until the upload succeeds so the next 1s tick retries.

plugins/NodeDiagnostics/NodeDiagnosticsPlugin.cs
```diff
@@ -0,0 +1,134 @@
...
+            _dispatcher.TryEnqueue(report);
+    }
+
+    private void TaskScheduler_UnobservedTaskException(object? sender, UnobservedTaskExceptionEventArgs e)
+    {
+        if (_dispatcher is null) return;
+
+        var report = NodeDiagnosticsEvent.FromException("UnobservedTaskException", e.Exception, false, _settings, _system);
+        _dispatcher.TryEnqueue(report);
+        e.SetObserved();
```
cschuchardt88, line 129: **bug:** `SetObserved()` runs even when `TryEnqueue` returns false. A full diagnostics queue then swallows the unobserved task exception process-wide with no sink report (and this handler is subscribed by default).

Only call `SetObserved()` after a successful enqueue, or make observing opt-in so a dropped event can still surface through the normal unobserved-task path.

plugins/NodeDiagnostics/NodeDiagnosticsDispatcher.cs
```diff
@@ -0,0 +1,473 @@
...
+        }
+    }
+
+    private async Task ProcessHeartbeatAsync()
+    {
+        try
+        {
+            while (!_shutdown.IsCancellationRequested)
+            {
+                await SendHeartbeatAsync(_shutdown.Token).ConfigureAwait(false);
```
cschuchardt88, line 398: **suggestion:** The wait is `await SendHeartbeatAsync` then `await Delay(HeartbeatInterval)`, so the real ping period includes HTTP retries. Default `MaxRetries=3` / `RequestTimeoutMilliseconds=5000` can stretch a 60s heartbeat past a Better Stack/Healthchecks check window whenever the sink is slow.

Use a `PeriodicTimer` (or delay based on remaining interval) so pings stay on wall-clock schedule.

plugins/NodeDiagnostics/NodeDiagnosticsSinkSettings.cs
```diff
@@ -0,0 +1,165 @@
...
+    {
+        if (string.IsNullOrWhiteSpace(Name))
+            throw new ArgumentException("Sink name cannot be empty.", nameof(Name));
+        if (string.IsNullOrWhiteSpace(Method))
+            throw new ArgumentException("Method cannot be empty.", nameof(Method));
+        if (Enabled && Method != "GET" && Method != "POST" && Method != "PUT")
+            throw new ArgumentException("Method must be GET, POST, or PUT.", nameof(Method));
+        if (!string.IsNullOrEmpty(Token) && string.IsNullOrWhiteSpace(TokenHeader))
+            throw new ArgumentException("TokenHeader cannot be empty when Token is configured.", nameof(TokenHeader));
+        if (Enabled && Endpoint!.Scheme != Uri.UriSchemeHttps && SendsCredentials())
```
cschuchardt88, line 150: **suggestion:** HTTPS is enforced only for `Token` / known credential *headers*. Heartbeat providers (and GCP's documented API-key-in-URL mode) put the secret in the endpoint path/query, so `http://hc-ping.com/<uuid>` is accepted.

Treat an enabled status/heartbeat endpoint as a credential and require HTTPS whenever `Endpoint` is set.

plugins/NodeDiagnostics/NodeDiagnosticsEvent.cs
```diff
@@ -0,0 +1,250 @@
...
+            OSDescription: RuntimeInformation.OSDescription,
+            ProcessArchitecture: RuntimeInformation.ProcessArchitecture.ToString(),
+            ProcessId: System.Environment.ProcessId,
+            ServiceName: settings.ServiceName,
+            Environment: settings.Environment,
+            NodeName: string.IsNullOrWhiteSpace(settings.NodeName) ? null : settings.NodeName,
+            Network: system?.Settings.Network,
+            NodeVersion: Assembly.GetEntryAssembly()?.GetName().Version?.ToString(),
+            PluginVersion: typeof(NodeDiagnosticsPlugin).Assembly.GetName().Version?.ToString(),
+            BlockHeight: null,
```
cschuchardt88, line 78: **suggestion:** Crash/exception payloads leave liveness fields null even though the dispatcher already tracks block height, header height, and seconds since last persist. Filling these (or enriching in `TryEnqueue`/`SendNow`) would let error sinks correlate failures with stall vs. sync issues.

plugins/NodeDiagnostics/NodeDiagnosticsDispatcher.cs
```diff
@@ -0,0 +1,473 @@
...
+    {
+        _settings = settings;
+        _systemProvider = systemProvider;
+        _httpClient = httpClient;
+        _disposeClient = disposeClient;
+        _channel = Channel.CreateBounded<NodeDiagnosticsEvent>(new BoundedChannelOptions(settings.MaxQueueSize)
+        {
+            SingleReader = true,
+            SingleWriter = false,
+            FullMode = BoundedChannelFullMode.Wait
```
cschuchardt88, line 57: **nit:** `TryEnqueue` drops via `TryWrite`, and the README describes drop-on-full, but `FullMode` is `Wait`. `TryWrite` still returns false today; `WriteAsync` later would block an exception handler. Prefer `BoundedChannelFullMode.DropWrite` to match the documented policy.

cschuchardt88 (commented):
## Summary

Adds a NodeDiagnostics plugin that forwards unhandled/unobserved exceptions, optional startup probes, process heartbeats, and local block-stall alerts to configurable HTTP sinks (custom webhook, experimental Sentry/GCP Error Reporting, Better Stack/Healthchecks). Queueing, retries, HTTPS-for-token, and tests for routing/payload/timeouts look solid. The default stall threshold will false-alarm through a normal DBFT view change, stall alerts can be dropped after a failed send, and `SetObserved()` can swallow unobserved task exceptions that never made the queue.

## Issues

- **bug** plugins/NodeDiagnostics/NodeDiagnosticsSettings.cs:59 — Default `ConsensusStallThresholdSeconds` is 30, which is shorter than a single DBFT view-change window (`TimePerBlock << (view+1)` is 60s at view 1 with 15s blocks). Healthy nodes will emit `ConsensusStall` during ordinary view changes. Raise the default (and consider a floor in `Validate()`) to cover at least one view change plus margin.
  Status: open

- **bug** plugins/NodeDiagnostics/NodeDiagnosticsDispatcher.cs:360 — `_lastReportedConsensusStallHeight` is set before `SendEventsAsync` succeeds. A 5xx/timeout means that height is never retried, so the one-alert-per-height guarantee becomes zero alerts if the sink is briefly down. Record the height only after a successful send (or retry on the next 1s tick until it succeeds).
  Status: open

- **bug** plugins/NodeDiagnostics/NodeDiagnosticsPlugin.cs:129 — `e.SetObserved()` runs even when `TryEnqueue` returns false. A full queue (or disabled event sinks after subscribe) marks the exception observed and drops it, changing process-wide unobserved-task behavior without a report. Only call `SetObserved()` after a successful enqueue, or make observing opt-in.
  Status: open

- **suggestion** plugins/NodeDiagnostics/NodeDiagnosticsDispatcher.cs:398 — Heartbeat period is `SendHeartbeatAsync` duration plus `HeartbeatInterval`. With default retries (3×5s + delays) a Better Stack/Healthchecks check configured at 60s can miss pings whenever a sink is slow. Drive pings from a fixed schedule (`PeriodicTimer` / delay-first) so the interval does not include HTTP time.
  Status: open

- **suggestion** plugins/NodeDiagnostics/NodeDiagnosticsSinkSettings.cs:150 — HTTPS is required only for `Token` / known credential headers. Better Stack and Healthchecks secrets live in the URL path (GCP also documents API keys in the endpoint). `http://hc-ping.com/<uuid>` currently validates. Treat heartbeat/status URLs as credentials and require HTTPS whenever `Enabled`.
  Status: open

- **suggestion** plugins/NodeDiagnostics/NodeDiagnosticsEvent.cs:78 — Exception/crash events always send `BlockHeight`/`HeaderHeight`/`SecondsSinceLastBlockAdvance` as null. The dispatcher already tracks liveness; attach it so error sinks can correlate crashes with stall/sync state.
  Status: open

- **nit** plugins/NodeDiagnostics/NodeDiagnosticsDispatcher.cs:57 — Queue is documented as drop-on-full and `TryEnqueue` uses `TryWrite`, but `FullMode` is `Wait`. Use `BoundedChannelFullMode.DropWrite` so a later `WriteAsync` cannot block an exception handler.
  Status: open

PR #1079, opened 2026-06-20
Fix snapshot point read cache options

Jim8y:
## Summary
- Keep fill-cache disabled for snapshot scans and explicitly enabled for point reads.
- Verify cache configuration and every Contains/TryGet route, including iterator state machines.
- Exercise real snapshot reads for existing, missing, and empty values, forward/backward scans, and snapshot isolation.

## Validation
- Storage tests: 9/9 pass in Debug and Release.
- Mutation checks detect both incorrect point-cache settings and a single TryGet overload using scan options.
- Formatting and diff checks pass. No CI or dependency changes.

### DIFF
--- a/tests/Neo.Plugins.Storage.Tests/SnapshotReadOptionsTest.cs
+++ b/tests/Neo.Plugins.Storage.Tests/SnapshotReadOptionsTest.cs
@@ -1,0 +1,102 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// SnapshotReadOptionsTest.cs file belongs to the neo project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using System.Reflection;
+
+namespace Neo.Plugins.Storage.Tests;
+
+[TestClass]
+public class SnapshotReadOptionsTest
+{
+    [TestMethod]
+    public void LevelDbSnapshotUsesDedicatedReadOptionsForScanAndPointReads()
+    {
+        var snapshotType = GetSnapshotType(typeof(LevelDBStore));
+        var scanReadOptions = GetField(snapshotType, "_scanReadOptions");
+        var pointReadOptions = GetField(snapshotType, "_pointReadOptions");
+
+        AssertMethodUsesOnly(snapshotType, "Find", scanReadOptions, pointReadOptions);
+        AssertMethodUsesOnly(snapshotType, "GetEnumerator", scanReadOptions, pointReadOptions);
+        AssertMethodUsesOnly(snapshotType, "Contains", pointReadOptions, scanReadOptions);
+        AssertMethodUsesOnly(snapshotType, "TryGet", pointReadOptions, scanReadOptions);
+    }
+
+    [TestMethod]
+    public void RocksDbSnapshotUsesDedicatedReadOptionsForScanAndPointReads()
+    {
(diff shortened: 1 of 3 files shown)


erikzhang:
What improvements have been made?

Jim8y:
The improvement is in the snapshot point-read path.

Before this change, each storage snapshot used a single read option with `FillCache=false`. That option is appropriate for scan/iterator reads, but it was also used by `Contains` and `TryGet`. Point reads could still hit existing LevelDB/RocksDB block-cache entries, but cache misses did not populate the cache.

This can matter for repeated snapshot point reads during transaction verification. The benefit depends on the working set and existing block-cache and OS-cache contents. The microbenchmark below does not establish that this setting caused a particular consensus slowdown or outage.

This PR keeps `FillCache=false` for scan paths only:

- `Find` / iterator reads still use no-fill-cache read options to avoid cache pollution during scans.
- `Contains` / `TryGet` now use separate snapshot read options with the default fill-cache behavior.
- The same separation is applied to both LevelDBStore and RocksDBStore.

I also ran a local microbenchmark for repeated snapshot point reads. This is not a full chain benchmark, but it matches the hot point-read pattern this PR changes.

Environment:

- macOS arm64
- .NET SDK 10.0.108
- Release build
- 100,000 entries, 512 hot keys
- 10 warmup repetitions, then 7 measured rounds
- Each round: 512,000 point operations (`TryGet` + `Contains`)
- Result below uses median ops/sec

| Store | Before | After | Change |
|---|---:|---:|---:|
| LevelDBStore | 1,345,771 ops/s | 1,864,834 ops/s | +38.6% |
| RocksDBStore | 343,553 ops/s | 1,353,306 ops/s | +293.9% |

The main goal is not to speed up scans. The goal is to avoid making normal snapshot point reads behave like bulk scans from the cache perspective.

shargon:
But then Contains is not cached?

tests/Neo.Plugins.Storage.Tests/SnapshotReadOptionsTest.cs
```diff
@@ -0,0 +1,102 @@
...
+
+using System.Reflection;
+
+namespace Neo.Plugins.Storage.Tests;
+
+[TestClass]
+public class SnapshotReadOptionsTest
+{
+    [TestMethod]
+    public void LevelDbSnapshotUsesDedicatedReadOptionsForScanAndPointReads()
```
cschuchardt88, line 29: **[suggestion]** `LevelDbSnapshotUsesDedicatedReadOptionsForScanAndPointReads` and `RocksDbSnapshotUsesDedicatedReadOptionsForScanAndPointReads` never construct a snapshot and never inspect fill-cache. They only scan method IL for the `_scan*` vs `_point*` field tokens. A constructor that set `FillCache=false` on the point-read options (or forgot `SetFillCache(false)` on the scan options) would still pass, which is the behavior this PR is meant to lock in. `UsesField` also matches a metadata token anywhere in the body (no `ldfld`/`ldflda` check) and `AssertMethodUsesOnly` treats “any overload/state-machine method references the expected field” as success, so one `TryGet` overload could silently miss the point-read options if the other hit.

**Suggestion:** Also scan the constructors (or otherwise assert setup) so scan options are the only ones that disable fill-cache, and require every `TryGet` overload to load the point-read field. Prefer opcode-aware `ldfld` matching over a raw 4-byte token search.

shargon:
> But then Contains is not cached?

ping @Jim8y

Jim8y:
@shargon `Contains` uses point-read options in both stores. `FillCache=false` prevents filling the cache, not hitting existing entries. Tests now verify the settings and every overload; Debug and Release each pass 9/9.

shargon (approved):

ajara87 (approved):

PR #1058, opened 2026-06-01
Check `MaxFee` during RPC.CancelTransaction

shargon:
This pull request refactors the `CancelTransaction` method in `RpcServer.Wallet.cs` to consistently use a snapshot of the system state rather than the live `StoreView`. This change ensures more reliable and isolated transaction handling by operating on a consistent view of the blockchain state during cancellation operations.

**Refactoring for consistent snapshot usage:**

* Replaced usage of `system.StoreView` with a locally scoped `snapshot` (created via `system.GetSnapshotCache()`) throughout the `CancelTransaction` method to ensure all transaction checks and operations are performed on a consistent state. [[1]](diffhunk://#diff-117ada1714c3ef54ab7a33751d40147a8bcb45c8f47ad9a664eb15a277c7c353L783-R785) [[2]](diffhunk://#diff-117ada1714c3ef54ab7a33751d40147a8bcb45c8f47ad9a664eb15a277c7c353L798-R815)
* Updated calls to `MakeTransaction`, `AssetDescriptor`, `SignAndRelay`, and introduced `EnsureNetworkFee` to use the new `snapshot` variable, further enforcing the use of a consistent state during transaction cancellation.

PR #1080, opened 2026-06-23
fix RpcServer SendToAddress/SendFrom/SendMany

### LINKED ISSUE #1069: `sendtoaddress`/`sendfrom`/`sendmany` build unfundable txs → `-511 InsufficientFunds` at relay
## Summary

  The wallet RPC send methods can build, sign, and relay a transaction whose fee-payer (`Signers[0]`) cannot cover the network fee, producing `InsufficientFunds` (-511) at relay — even though `Wallet.MakeTransaction` already validated that the selected sender could afford the fee. The network fee  is **raised after `MakeTransaction`** (i.e. after the signer was selected and the fee validated), with no re-check that the selected sender can still pay. Verified on neo-node v3.9.2 / neo core v3.9.1.

  ## Root cause

  [`Wallet.MakeTransaction`](https://github.com/neo-project/neo/blob/v3.9.1/src/Neo/Wallets/Wallet.cs#L641) selects the fee-payer (`Signers[0]`) from [GAS-holding accounts (L600)](https://github.com/neo-project/neo/blob/v3.9.1/src/Neo/Wallets/Wallet.cs#L600) and only returns a tx the chosen sender can afford:
  ```cs
  tx.NetworkFee = tx.CalculateNetworkFee(...);            // Wallet.cs:669
  if (value >= tx.SystemFee + tx.NetworkFee) return tx;   // Wallet.cs:670  ← signer selected + fee validated here
  ```

  Then, **after** that selection, [`EnsureNetworkFee`](https://github.com/neo-project/neo-node/blob/v3.9.2/plugins/RpcServer/RpcServer.Wallet.cs#L547-L553) raises the network fee, with no re-check of the chosen sender:
  ```cs
  long calFee = tx.Size * NativeContract.Policy.GetFeePerByte(snapshot) + 100000;
  if (tx.NetworkFee < calFee) tx.NetworkFee = calFee;     // raises NetworkFee above the validated value
  ```

  At relay, `Transaction.VerifyStateDependent` requires `GAS.BalanceOf(Signers[0]) >= SystemFee + NetworkFee(raised)`. When the raised fee exceeds what was validated at [`Wallet.cs:670`](https://github.com/neo-project/neo/blob/v3.9.1/src/Neo/Wallets/Wallet.cs#L670), the selected sender no longer covers it and the tx is rejected with `-511`. Affects `sendtoaddress`, `sendfrom`, and `sendmany` (all call `EnsureNetworkFee`).

superboyiii:
Close https://github.com/neo-project/neo-node/issues/1069
Issue brought by https://github.com/neo-project/neo-node/pull/1058

Originally, we add `EnsureNetworkFee` to ensure Minimum `networkfee` for Rpc send method. It's working. 
(Although it's unnecessary in N3, `sendrawtransaction` can bypass this restriction. And in mainnet, `networkfee` is very cheap now and always lower than that limit).
After https://github.com/neo-project/neo-node/pull/1058, it's moved into `SignAndRelay`. The same bump ran after `wallet.Sign()` and after witnesses were attached, without re-signing. Because `NetworkFee` is part of the transaction data covered by the signature, mutating it post-sign invalidated the witness and caused relay failures (InvalidSignature) on mainnet for ordinary wallet RPC transfers.
Now I remove that `+100000` and add resign to ensure consistency.


tests/Neo.Plugins.RpcServer.Tests/UT_RpcServer.Wallet.cs
```diff
@@ -664,6 +666,65 @@ public void TestSignAndRelay_RelayFailure_ThrowsRpcException()
...
+    /// <summary>
+    /// Regression for issue #1069 on mainnet policy (feePerByte=20, execFeeFactor=1):
+    /// legacy SignAndRelay used size*feePerByte+100000 which always exceeded CalculateNetworkFee for single-sig txs.
+    /// </summary>
+    [TestMethod]
+    public void TestSendToAddress_MainnetPolicy_KeepsCalculatedNetworkFee()
+    {
+        var snapshot = _neoSystem.GetSnapshotCache();
+        SetMainnetPolicyFees(snapshot);
+        snapshot.Commit();
```
shargon, line 677: not need to commit the snapshot, otherwise could affect other tests

tests/Neo.Plugins.RpcServer.Tests/UT_RpcServer.Wallet.cs
```diff
@@ -664,6 +665,156 @@ public void TestSignAndRelay_RelayFailure_ThrowsRpcException()
...
```
shargon, line 677:
```suggestion
```

ajara87:
SignAndRelay is signing without checking if callers have previosly sign. Is it ok?

plugins/RpcServer/RpcServer.Wallet.cs
```diff
@@ -941,24 +941,33 @@ private JObject SignAndRelay(DataCache snapshot, Transaction tx)
...
-        }
-        else
-        {
-            return context.ToJson();
         }
+
+        if (NativeContract.GAS.BalanceOf(snapshot, tx.Sender) < tx.SystemFee + tx.NetworkFee)
+            throw new RpcException(RpcError.InsufficientFunds);
+
+        if (tx.NetworkFee > settings.MaxFee)
```
shargon, line 964:
```suggestion
        if (tx.NetworkFee + tx.SystemFee > settings.MaxFee)
```

superboyiii: This changes `MaxFee` behavior, it's only for `networkfee`. So why it's 0.1 as default. I also calculated a pure clean tx with all default settings, it costs about 0.11 GAS for networkfee + systemFee (about 0.1 GAS is on systemfee). So if we change to this, Rpc send will not work on privatenet.

tests/Neo.Plugins.RpcServer.Tests/UT_RpcServer.Wallet.cs
superboyiii: > not need to commit the snapshot, otherwise could affect other tests

`snapshot.Commit()` is required because `SendToAddress` reads policy via a fresh `GetSnapshotCache()`, which only sees committed store state. Without it, the test does not exercise mainnet fee parameters. 
`Cross-test leakage` is unlikely: each test gets a new `MemoryStore` in `TestInitialize`, and `TestCleanup` calls `Reset()`. The same approach is already used in `TestSetup` for the GAS balance: modify the snapshot, then `Commit()` so subsequent `GetSnapshotCache()` calls see the updated state.

plugins/RpcServer/RpcServer.Wallet.cs
```diff
@@ -941,24 +941,33 @@ private JObject SignAndRelay(DataCache snapshot, Transaction tx)
...
+
+        tx.Witnesses = context.GetWitnesses();
+
+        // Align with MakeTransaction, calculatenetworkfee RPC, and on-chain verification.
+        var calculatedFee = tx.CalculateNetworkFee(snapshot, system.Settings, wallet);
+        if (calculatedFee > tx.NetworkFee)
         {
+            tx.NetworkFee = calculatedFee;
+            context = new ContractParametersContext(snapshot, tx, system.Settings.Network);
+            wallet.Sign(context);
```
AnnaShaleva, line 955: Is it possible that it's not the only witness for this transaction? Current wallet can sign the new tx, but other existing signatures are already constructed for the previous version of this tx.

superboyiii:
> SignAndRelay is signing without checking if callers have previosly sign. Is it ok?

Yes, this is intentional. `SignAndRelay` is the single entry point for sign → align NetworkFee → re-sign if needed → relay. It always builds a fresh `ContractParametersContext` and signs; it does not check whether callers already signed or reuse attached witnesses.

Callers like `sendtoaddress` may sign first only to return partial multi-sig JSON early when signing is incomplete, without attempting relay. `canceltransaction` never pre-signs and relies entirely on `SignAndRelay`.

Re-signing is required when `NetworkFee` is raised to match `CalculateNetworkFee` (fee is part of the tx hash). For unchanged single-sig txs, signing twice is redundant but harmless.

Bottom line: the authoritative signature is always the one produced inside SignAndRelay, not any prior caller-side signing.

plugins/RpcServer/RpcServer.Wallet.cs
superboyiii: > Is it possible that it's not the only witness for this transaction? Current wallet can sign the new tx, but other existing signatures are already constructed for the previous version of this tx.

The scenario you’re describing — old sigs from version A mixed with new sigs from version B — is really the `sign` → `relay` path, where `ContractParametersContext.Parse` accumulates signatures step by step. `SignAndRelay` is only used by the wallet `send*` RPCs and doesn’t work that way.

shargon (approved):

ajara87 (approved):

vncoelho (approved):

PR #1084, opened 2026-07-04
Fix: The `start` should be the inclusive max key if `Backward` scan

Wi1l-B0t:
The `start` should be the inclusive max key if `Backward` scan, and the inclusive min key if `Forward` scan.
The current implementation treats the `start` as the min key if `Backward` scan.

### DIFF
--- a/plugins/LevelDBStore/Plugins/Storage/Snapshot.cs
+++ b/plugins/LevelDBStore/Plugins/Storage/Snapshot.cs
@@ -79,51 +79,22 @@ internal class Snapshot : IStoreSnapshot, IEnumerable<KeyValuePair<byte[], byte[
         ArgumentNullException.ThrowIfNull(start);
         ArgumentNullException.ThrowIfNull(end);
 
-        if (start.AsSpan().SequenceCompareTo(end) >= 0)
+        var order = start.AsSpan().SequenceCompareTo(end);
+        if ((direction == SeekDirection.Forward && order >= 0) ||
+            (direction == SeekDirection.Backward && order <= 0))
+        {
             yield break;
-
-        using var iterator = _db.CreateIterator(_readOptions);
-
-        if (direction == SeekDirection.Forward)
-        {
-            iterator.Seek(start);
-
-            while (iterator.Valid())
-            {
-                var key = iterator.Key();
-                if (key is null) break;
-
-                if (key.AsSpan().SequenceCompareTo(end) >= 0)
-                    break;
-
-                yield return (key, iterator.Value()!);
-                iterator.Next();
-            }
         }
-        else
+
+        foreach (var item in _db.Seek(_readOptions, start, direction))
         {
-            iterator.Seek(end);
-
-            if (!iterator.Valid())
+            order = item.Key.AsSpan().SequenceCompareTo(end);
+            if ((direction == SeekDirection.Forward && order >= 0) ||
+                (direction == SeekDirection.Backward && order <= 0))
             {
-                iterator.SeekToLast();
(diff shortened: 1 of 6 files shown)


plugins/LevelDBStore/Plugins/Storage/Snapshot.cs
```diff
@@ -79,51 +79,15 @@ public void Dispose()
         ArgumentNullException.ThrowIfNull(start);
         ArgumentNullException.ThrowIfNull(end);
 
-        if (start.AsSpan().SequenceCompareTo(end) >= 0)
-            yield break;
```
ajara87, line 83: The removal of the start >= end early exit guard for the Forward direction silently accepts inverted range arguments (start > end) instead of failing fast. Consider restoring it as a directional guard.

```
    if (direction == SeekDirection.Forward && start.AsSpan().SequenceCompareTo(end) > 0)
        yield break;
    if (direction == SeekDirection.Backward && start.AsSpan().SequenceCompareTo(end) < 0)
        yield break;
```

Wi1l-B0t: No behavioral change

ajara87: > No behavioral change

Agree, no behavioral change. It's purely a readability preference the explicit guard makes the valid input contract visible at a glance without having to trace the iteration logic

Wi1l-B0t: Modified.

AnnaShaleva (commented):
This will change the existing behaviour of interops like `System.Storage.Find` and some  other core functionality dependent from the Store.

Wi1l-B0t:
> This will change the existing behaviour of interops like `System.Storage.Find` and some other core functionality dependent from the Store.

The current implementation is wrong, and this method is not used by `System.Storage.Find`.

Only  the N4 has this `FindRange` method.

Wi1l-B0t:
> This will change the existing behaviour of interops like `System.Storage.Find` and some other core functionality dependent from the Store.

The `System.Storage.Find` doestn't use `FindRange`:
https://github.com/neo-project/neo/blob/b108895ac6601d21b89d47c5f5589b9a8a948ded/src/Neo/SmartContract/ApplicationEngine.Storage.cs#L205

The start key is the inclusive max key if Backword:
https://github.com/neo-project/neo/blob/b108895ac6601d21b89d47c5f5589b9a8a948ded/src/Neo/SmartContract/Native/Governance.cs#L532-L538

https://github.com/neo-project/neo/blob/b108895ac6601d21b89d47c5f5589b9a8a948ded/src/Neo/Persistence/DataCache.cs#L309-L321

plugins/RocksDBStore/Plugins/Storage/Snapshot.cs
```diff
@@ -62,59 +62,40 @@ public void Put(byte[] key, byte[] value)
...
         if (direction == SeekDirection.Forward)
+        {
             for (it.Seek(keyOrPrefix); it.Valid(); it.Next())
                 yield return (it.Key(), it.Value());
+        }
         else
+        {
             for (it.SeekForPrev(keyOrPrefix); it.Valid(); it.Prev())
                 yield return (it.Key(), it.Value());
+        }
```
roman-khimov: These better be omitted or committed separately, pure stylistic.

roman-khimov (dismissed):
Seems to be a  variation on the old FindRange from N3 affecting N4 only, not a problem to me.

Jim8y (approved):

shargon (approved):

PR #1089, opened 2026-07-18
ToJson max length in rpc node

shargon:
This pull request adds a new setting to control the maximum allowed size of individual item responses in the RPC server and updates the relevant code to enforce this limit. The main changes introduce the `MaxItemResponseSize` configuration property and ensure that JSON serialization of stack items respects this limit.

**Configuration Enhancements:**

* Added a new `MaxItemResponseSize` property to the `RpcServersSettings` record, defaulting to `ushort.MaxValue`, allowing configuration of the maximum size for item responses.
* Updated the `Load` method to read the `MaxItemResponseSize` value from the configuration section.

**Enforcing Response Size Limits:**

* Modified calls to `ToJson` in `GetInvokeResult` to pass the configured `MaxItemResponseSize`, ensuring that event and stack item serialization respects the size limit. [[1]](diffhunk://#diff-8f3dae74fc0523e664fd6521c1debcd63e603e12415a4acdfd602b24c14f04f2L83-R83) [[2]](diffhunk://#diff-8f3dae74fc0523e664fd6521c1debcd63e603e12415a4acdfd602b24c14f04f2L101-R101)
* Updated the `ToJson` method to accept the `maxItemResponseSize` parameter and pass it to the stack item's `ToJson` method, enforcing the response size limit during serialization.

### DIFF
--- a/plugins/RpcServer/RcpServerSettings.cs
+++ b/plugins/RpcServer/RcpServerSettings.cs
@@ -37,6 +37,7 @@ public record RpcServersSettings
     public string[] TrustedAuthorities { get; init; } = [];
     public int MaxConcurrentConnections { get; init; } = 40;
     public int MaxRequestBodySize { get; init; } = 1 * 1024 * 1024;
+    public int MaxItemResponseSize { get; init; } = ushort.MaxValue;
     public string RpcUser { get; init; } = string.Empty;
     public string RpcPass { get; init; } = string.Empty;
     public bool EnableCors { get; init; } = true;
@@ -89,6 +90,7 @@ public record RpcServersSettings
             MaxGasInvoke = GetGasAmount(section, "MaxGasInvoke", @default.MaxGasInvoke),
             MaxFee = GetGasAmount(section, "MaxFee", @default.MaxFee),
             MaxIteratorResultItems = section.GetValue("MaxIteratorResultItems", @default.MaxIteratorResultItems),
+            MaxItemResponseSize = section.GetValue("MaxItemResponseSize", @default.MaxItemResponseSize),
             MaxStackSize = section.GetValue("MaxStackSize", @default.MaxStackSize),
             DisabledMethods = GetStrings(section, "DisabledMethods"),
             MaxConcurrentConnections = section.GetValue("MaxConcurrentConnections", @default.MaxConcurrentConnections),
--- a/plugins/RpcServer/RcpServerSettings.cs
+++ b/plugins/RpcServer/RcpServerSettings.cs
@@ -10,6 +10,7 @@
 // modifications are permitted.
 
 using Microsoft.Extensions.Configuration;
+using Neo.Network.P2P.Payloads;
 using Neo.SmartContract.Native;
 using System.Net;
 
@@ -37,7 +38,7 @@ public record RpcServersSettings
     public string[] TrustedAuthorities { get; init; } = [];
     public int MaxConcurrentConnections { get; init; } = 40;
     public int MaxRequestBodySize { get; init; } = 1 * 1024 * 1024;
-    public int MaxItemResponseSize { get; init; } = ushort.MaxValue;
+    public int MaxItemResponseSize { get; init; } = Transaction.MaxTransactionSize;
     public string RpcUser { get; init; } = string.Empty;
     public string RpcPass { get; init; } = string.Empty;
     public bool EnableCors { get; init; } = true;
--- a/plugins/RpcServer/RpcServer.SmartContract.cs
+++ b/plugins/RpcServer/RpcServer.SmartContract.cs
@@ -80,7 +80,7 @@ partial class RpcServer
                 {
                     ["eventname"] = n.EventName,
                     ["contract"] = n.ScriptHash.ToString(),
-                    ["state"] = ToJson(n.State, session),
+                    ["state"] = ToJson(n.State, session, settings.MaxItemResponseSize),
                 };
             }));
             if (useDiagnostic)
@@ -98,7 +98,7 @@ partial class RpcServer
             {
                 try
                 {
-                    stack.Add(ToJson(item, session));
+                    stack.Add(ToJson(item, session, settings.MaxItemResponseSize));
                 }
                 catch (Exception ex)
                 {
@@ -173,9 +173,9 @@ partial class RpcServer
         return array;
     }
 
-    private static JObject ToJson(StackItem item, Session session)
+    private JObject ToJson(StackItem item, Session session, int maxItemResponseSize)
     {
-        var json = item.ToJson();
+        var json = item.ToJson(maxItemResponseSize);
         if (item is InteropInterface interopInterface && interopInterface.GetInterface<object>() is IIterator iterator)
         {
             Guid id = Guid.NewGuid();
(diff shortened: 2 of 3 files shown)


plugins/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -173,9 +173,9 @@ protected static JArray ToJson(IEnumerable<KeyValuePair<StorageKey, DataCache.Tr
         return array;
     }
 
-    private static JObject ToJson(StackItem item, Session session)
+    private JObject ToJson(StackItem item, Session session, int maxItemResponseSize)
```
roman-khimov, line 176: Since the limit is for individual items, https://github.com/neo-project/neo/pull/4581#discussion_r3608429639 is also relevant to me. Yes, big items will be capped, but it's not normal to have that big items here.

shargon: So you don't consider important this PR now?

roman-khimov: With https://github.com/neo-project/neo/pull/4581 merged it becomes less critical, only if we want to have some configuration and/or lower default limit for RPC. Personally I'd still prefer stricter limits.

vncoelho (approved):

plugins/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -80,7 +80,7 @@ private JObject GetInvokeResult(byte[] script, Signer[]? signers = null, Witness
                 {
                     ["eventname"] = n.EventName,
                     ["contract"] = n.ScriptHash.ToString(),
-                    ["state"] = ToJson(n.State, session),
+                    ["state"] = ToJson(n.State, session, settings.MaxItemResponseSize),
```
cschuchardt88, line 83: **[bug]** Notification `state` uses the new limit with no try/catch. Size (or circular-ref) failures throw out of `GetInvokeResult`, the outer catch disposes the session, and the client gets an RPC error with no `state`/`gasconsumed`/`stack`. Stack items at line 101-106 already convert the same `ToJson` exception into `"error: " + ex.Message`. After this PR a successful VM run with one notification over 102400 bytes of JSON fails the entire `invokefunction`/`invokescript`; previously that only happened for circular refs (or sizes above neo’s much larger default).

**Suggestion:** Wrap each notification `ToJson` like the result stack (replace oversized/circular state with an error string) so invoke still returns HALT, gas, and stack.

plugins/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -173,9 +173,9 @@ protected static JArray ToJson(IEnumerable<KeyValuePair<StorageKey, DataCache.Tr
         return array;
     }
 
-    private static JObject ToJson(StackItem item, Session session)
+    private JObject ToJson(StackItem item, Session session, int maxItemResponseSize)
     {
-        var json = item.ToJson();
+        var json = item.ToJson(maxItemResponseSize);
```
cschuchardt88, line 178: **[suggestion]** Behavior is exception, not truncation. In neo `StackItem.ToJson`, `ByteString`/`Buffer` always `Convert.ToBase64String(item.GetSpan())` before subtracting from `maxSize`, then throws if `maxSize < 0`. A client can still force a large allocation on invoke with a near-`MaxItemSize` buffer, then hit `"error: Max size reached."` (stack) or a full RPC error (notifications). The cap is also per root item: N stack entries each just under 102400 bytes still produce a multi-megabyte HTTP body (`ProcessAsync` writes `response.ToString()` with no total response limit). Nearby knobs are `MaxRequestBodySize` (1 MiB), `MaxIteratorResultItems` (100), `MaxStackSize` (65535), and neo `MaxToJsonSize` (262140).

**Suggestion:** Document exception-vs-truncate. Consider catching size errors before/without relying on full base64 (needs a neo `ToJson` change), and/or a total invoke-response budget in addition to the per-item cap.

plugins/RpcServer/RcpServerSettings.cs
```diff
@@ -89,6 +91,7 @@ public static RpcServersSettings Load(IConfigurationSection section)
             MaxGasInvoke = GetGasAmount(section, "MaxGasInvoke", @default.MaxGasInvoke),
             MaxFee = GetGasAmount(section, "MaxFee", @default.MaxFee),
             MaxIteratorResultItems = section.GetValue("MaxIteratorResultItems", @default.MaxIteratorResultItems),
+            MaxItemResponseSize = section.GetValue("MaxItemResponseSize", @default.MaxItemResponseSize),
```
cschuchardt88, line 94: **[suggestion]** `section.GetValue("MaxItemResponseSize", @default.MaxItemResponseSize)` accepts 0 and negatives. `StackItem.ToJson` has no `maxSize <= 0` guard (unlike `EvaluationStack.ToJson`). A zero/negative setting makes every stack item `"error: Max size reached."` and, with the notification try/catch gap, makes every invoke with notifications fail. 0 is not treated as unlimited.

**Suggestion:** Reject `MaxItemResponseSize <= 0` at load (or treat 0 as unlimited if that is desired) and add a comment that the unit is JSON bytes per stack item, not VM `MaxItemSize`.

plugins/RpcServer/RcpServerSettings.cs
```diff
@@ -37,6 +38,7 @@ public record RpcServersSettings
     public string[] TrustedAuthorities { get; init; } = [];
     public int MaxConcurrentConnections { get; init; } = 40;
     public int MaxRequestBodySize { get; init; } = 1 * 1024 * 1024;
+    public int MaxItemResponseSize { get; init; } = Transaction.MaxTransactionSize;
```
cschuchardt88, line 41: **[suggestion]** No tests cover the new setting or overflow behavior. `UT_RpcServer.TestRpcServerSettings_Load` reads `RpcServer.json` (now `"MaxItemResponseSize": 102400`) but does not assert the property. Existing invoke tests only exercise small items, so Codecov “covered” does not mean the limit works. Sample json 102400 matches `Transaction.MaxTransactionSize`; the PR body still says `ushort.MaxValue`. Coupling the RPC JSON default to a P2P tx constant is a bit odd but the 100 KiB value is reasonable.

**Suggestion:** Assert loaded/default `MaxItemResponseSize`. Add invoke tests for (a) stack item over the limit → `"error: Max size reached."` while other fields remain, (b) notification over the limit (once the notification try/catch is decided), (c) `traverseiterator` respecting the cap, (d) config 0/negative.

cschuchardt88 (changes requested):

plugins/RpcServer/RpcServer.SmartContract.cs
shargon: Exception is expected

shargon:
Merge?

ajara87 (approved):

cschuchardt88:
> Merge?

need to fix https://github.com/neo-project/neo-node/pull/1089#discussion_r3855578739 bug first

shargon:
> > Merge?
> 
> need to fix [#1089 (comment)](https://github.com/neo-project/neo-node/pull/1089#discussion_r3855578739) bug first

exception is expected, we don't want to truncate the output

cschuchardt88 (approved):

roman-khimov (approved):

PR #1090, opened 2026-07-18
Limit invalid tx size

shargon:
This pull request refactors how invalid transactions are tracked in the consensus process by introducing a custom FIFO cache and updating all related usages. The main goal is to improve memory management and provide a more structured way to handle invalid transactions. The changes affect both the core consensus logic and related tests.

**Invalid transaction tracking improvements:**

* Replaced the `Dictionary<UInt256, HashSet<ECPoint>>` previously used for `InvalidTransactions` with a custom FIFO cache class (`InvalidCache`) that holds `UnvalidTxCacheItem` records, improving memory management and eviction of old invalid entries. [[1]](diffhunk://#diff-0508a82b0fba83f356d31bae2fdd86d324fc3534e7d51c18442de656c84fe433R30-R38) [[2]](diffhunk://#diff-0508a82b0fba83f356d31bae2fdd86d324fc3534e7d51c18442de656c84fe433L53-R63) [[3]](diffhunk://#diff-0508a82b0fba83f356d31bae2fdd86d324fc3534e7d51c18442de656c84fe433R128)
* Updated all usages of `InvalidTransactions` to work with the new cache structure, including accessing the `.Value` property of `UnvalidTxCacheItem` and adding new items in the correct format. [[1]](diffhunk://#diff-e95e82bca993e9c23781bc26001999bf0869e4d2cb3d66d85dd441116efe6dd1L87-R88) [[2]](diffhunk://#diff-c9d04939883f3698dadec88d5dd7c52c7743bfe61fa7efe3b371ae97549c9c67L224-R227) [[3]](diffhunk://#diff-38d81abf98259fc161cf3931bf76b1c3a58c9b529b1da0a1dd0a6b2e53c4ca31L321-R321)

**Codebase maintenance:**

* Added the necessary `using Neo.IO.Caching;` directive to support the new cache implementation.

### DIFF
--- a/plugins/DBFTPlugin/Consensus/ConsensusContext.cs
+++ b/plugins/DBFTPlugin/Consensus/ConsensusContext.cs
@@ -13,6 +13,7 @@ using Neo.Cryptography;
 using Neo.Cryptography.ECC;
 using Neo.Extensions;
 using Neo.IO;
+using Neo.IO.Caching;
 using Neo.Ledger;
 using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
@@ -26,6 +27,15 @@ namespace Neo.Plugins.DBFTPlugin.Consensus;
 
 public sealed partial class ConsensusContext : IDisposable, ISerializable
 {
+    public record UnvalidTxCacheItem(UInt256 Key, HashSet<ECPoint> Value);
+    public class InvalidCache(int maxCapacity) : FIFOCache<UInt256, UnvalidTxCacheItem>(maxCapacity)
+    {
+        protected override UInt256 GetKeyForItem(UnvalidTxCacheItem item)
+        {
+            return item.Key;
+        }
+    }
+
     /// <summary>
     /// Key for saving consensus state.
     /// </summary>
@@ -50,7 +60,7 @@ public sealed partial class ConsensusContext : IDisposable, ISerializable
     /// Store all verified unsorted transactions' senders' fee currently in the consensus context.
     /// </summary>
     public TransactionVerificationContext VerificationContext = new();
-    public Dictionary<UInt256, HashSet<ECPoint>> InvalidTransactions = new();
+    public InvalidCache InvalidTransactions;
 
     public StoreCache Snapshot { get; private set; }
     private ECPoint _myPublicKey;
@@ -115,6 +125,7 @@ public sealed partial class ConsensusContext : IDisposable, ISerializable
         _signer = signer;
         this.neoSystem = neoSystem;
         dbftSettings = settings;
+        InvalidTransactions = new InvalidCache(neoSystem.Settings.MemoryPoolMaxTransactions);
 
         if (dbftSettings.IgnoreRecoveryLogs == false)
         {
--- a/plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
+++ b/plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
@@ -221,10 +221,10 @@ partial class ConsensusService
                 foreach (UInt256 hash in message.RejectedHashes)
                 {
                     ECPoint pubkey = context.Validators[message.ValidatorIndex];
-                    if (context.InvalidTransactions.TryGetValue(hash, out var hashset))
-                        hashset.Add(pubkey);
+                    if (context.InvalidTransactions.TryGet(hash, out var hashset))
+                        hashset.Value.Add(pubkey);
                     else
-                        context.InvalidTransactions.Add(hash, [pubkey]);
+                        context.InvalidTransactions.Add(new ConsensusContext.UnvalidTxCacheItem(hash, new HashSet<ECPoint> { pubkey }));
                 }
                 break;
         }
--- a/tests/Neo.Plugins.DBFTPlugin.Tests/UT_ConsensusService.cs
+++ b/tests/Neo.Plugins.DBFTPlugin.Tests/UT_ConsensusService.cs
@@ -318,7 +318,7 @@ public class UT_ConsensusService : TestKit
         InvokeConsensusMethod(actor, "OnConsensusPayload", changeViewPayload);
 
         Assert.AreSame(changeViewPayload, context.ChangeViewPayloads[2]);
-        CollectionAssert.Contains(context.InvalidTransactions[rejectedHash].ToArray(), context.Validators[2]);
+        CollectionAssert.Contains(context.InvalidTransactions[rejectedHash].Value.ToArray(), context.Validators[2]);
 
         context.TransactionHashes = null;
         var commit = new Commit
--- a/plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
+++ b/plugins/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
@@ -84,8 +84,8 @@ partial class ConsensusContext
         // Iterate transaction until reach the size or maximum system fee
         foreach (Transaction tx in txs)
         {
-            if (InvalidTransactions.TryGetValue(tx.Hash, out var hashset))
-                if (hashset.Count > F) continue;
+            if (InvalidTransactions.TryGet(tx.Hash, out var hashset))
+                if (hashset.Value.Count > F) continue;
 
             // Check if maximum block size has been already exceeded with the current selected set
             blockSize += tx.Size;


plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
```diff
@@ -221,10 +221,10 @@ private void OnChangeViewReceived(ExtensiblePayload payload, ChangeView message)
                 foreach (UInt256 hash in message.RejectedHashes)
```
roman-khimov, line 221: One of the problems of this mechanism in general is that this set is completely disconnected from the proposal. Given that `MemoryPoolMaxTransactions` is usually less than `ushort.MaxValue` malicious node can subvert the mechanism completely by sending a full set of random hashes that will wipe out any existing ones.

My suggestion is to revert #984. Another option could be filtering hashes through PrepareRequest set of hashes if it's received and ignoring it completely if it's not.

roman-khimov (commented):
Otherwise the limit is OK.

vncoelho:
Lgtm @shargon , 
I will test performance.

shargon:
> My suggestion is to revert https://github.com/neo-project/neo-node/pull/984. Another option could be filtering hashes through PrepareRequest set of hashes if it's received and ignoring it completely if it's not.

Merge this to fix the issue, and discuss the revert later?

roman-khimov:
This is still relevant:
> malicious node can subvert the mechanism completely by sending a full set of random hashes that will wipe out any existing ones

We need to either fix it by filtering through the proposal or just drop the mechanism completely. I'm ok both ways.

plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
```diff
@@ -221,10 +221,10 @@ private void OnChangeViewReceived(ExtensiblePayload payload, ChangeView message)
...
                 {
                     ECPoint pubkey = context.Validators[message.ValidatorIndex];
-                    if (context.InvalidTransactions.TryGetValue(hash, out var hashset))
-                        hashset.Add(pubkey);
+                    if (context.InvalidTransactions.TryGet(hash, out var hashset))
+                        hashset.Value.Add(pubkey);
                     else
-                        context.InvalidTransactions.Add(hash, [pubkey]);
+                        context.InvalidTransactions.Add(new ConsensusContext.UnvalidTxCacheItem(hash, new HashSet<ECPoint> { pubkey }));
```
cschuchardt88, line 227: **[bug]** Every hash in `message.RejectedHashes` is inserted into `InvalidTransactions` with no check that it is in the mempool or the current `PrepareRequest`. `ChangeView.Deserialize` accepts up to `ushort.MaxValue` hashes, which is larger than the new capacity (`MemoryPoolMaxTransactions`, default 50_000). `FIFOCache` evicts the oldest key on `Add` once full (`Cache.AddInternal`), and FIFO `OnAccess` is a no-op, so later spam stays and earlier real votes disappear. After eviction, `EnsureMaxBlockLimitation` (`TryGet` + `hashset.Value.Count > F`) no longer skips that tx, so the primary can include it again. Honest backups still reject it (`AddTransaction` → `RequestChangeView`), producing repeated view changes — the stall #984 was meant to stop. One Byzantine validator can do this with a single `TxInvalid`/`TxRejectedByPolicy` ChangeView (`NewViewNumber` only has to be higher than that validator's last CV). Safety (commit/prepare of an invalid tx) is unchanged; the bounded cache is what makes wipe-by-eviction possible (the old `Dictionary` grew instead of dropping votes). LRU would not save a 50k+ flood of new keys either.

**Suggestion:** Do not cache hashes that are not in the mempool and/or the current proposal (ignore `RejectedHashes` entirely if no PrepareRequest is in hand). Also cap accepted `RejectedHashes` length to what honest nodes send (one hash per CV). Capacity can stay at `MemoryPoolMaxTransactions` once admission is filtered, because `MemPool_TransactionRemoved` already `Remove`s pool hashes.

tests/Neo.Plugins.DBFTPlugin.Tests/UT_ConsensusService.cs
```diff
@@ -318,7 +318,7 @@ public void TestConsensusServiceRoutesConsensusMessagesIntoContextState()
         InvokeConsensusMethod(actor, "OnConsensusPayload", changeViewPayload);
 
         Assert.AreSame(changeViewPayload, context.ChangeViewPayloads[2]);
-        CollectionAssert.Contains(context.InvalidTransactions[rejectedHash].ToArray(), context.Validators[2]);
+        CollectionAssert.Contains(context.InvalidTransactions[rejectedHash].Value.ToArray(), context.Validators[2]);
```
cschuchardt88, line 321: **[suggestion]** The only test update is `InvalidTransactions[rejectedHash].Value` after one `TxInvalid` ChangeView. Nothing fills the cache to `MemoryPoolMaxTransactions`, asserts `TryGet` false after FIFO wrap, or drives `EnsureMaxBlockLimitation` to skip a tx once `Value.Count > F`. The skip path in `ConsensusContext.MakePayload.cs:87-88` is therefore untested, as is the eviction behavior this PR exists to add.

**Suggestion:** Add a unit test that constructs `InvalidCache` (or a `ConsensusContext`) with a tiny capacity, inserts capacity+1 distinct hashes, and checks the oldest is gone and a later `TryGet`/`EnsureMaxBlockLimitation` still skips a hash with `> F` votes. A second case should show a spam flood of unknown hashes evicting a still-in-mempool invalid tx if admission stays unfiltered.

plugins/DBFTPlugin/Consensus/ConsensusContext.cs
```diff
@@ -26,6 +27,15 @@ namespace Neo.Plugins.DBFTPlugin.Consensus;
 
 public sealed partial class ConsensusContext : IDisposable, ISerializable
 {
+    public record UnvalidTxCacheItem(UInt256 Key, HashSet<ECPoint> Value);
```
cschuchardt88, line 30: **[nit]** `UnvalidTxCacheItem` is a typo for Invalid. The record and `InvalidCache` are also public nested types on `ConsensusContext`, which leaks a cache implementation detail from a type that is already a large public surface.

**Suggestion:** Rename to `InvalidTxCacheItem`. Make both nested types `private` (or file-scoped) if callers only need `InvalidTransactions` lookup/add/remove.

cschuchardt88 (changes requested):

plugins/DBFTPlugin/Consensus/ConsensusContext.cs
shargon: > **[nit]** `UnvalidTxCacheItem` is a typo for Invalid. The record and `InvalidCache` are also public nested types on `ConsensusContext`, which leaks a cache implementation detail from a type that is already a large public surface.
> 
> **Suggestion:** Rename to `InvalidTxCacheItem`. Make both nested types `private` (or file-scoped) if callers only need `InvalidTransactions` lookup/add/remove.

@copilot fix it

plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
cschuchardt88, line 227: **[bug]** Every hash in `message.RejectedHashes` is inserted into `InvalidTransactions` with no check that it is in the mempool or the current `PrepareRequest`. `ChangeView.Deserialize` accepts up to `ushort.MaxValue` hashes, which is larger than the new capacity (`MemoryPoolMaxTransactions`, default 50_000). `FIFOCache` evicts the oldest key on `Add` once full (`Cache.AddInternal`), and FIFO `OnAccess` is a no-op, so later spam stays and earlier real votes disappear. After eviction, `EnsureMaxBlockLimitation` (`TryGet` + `hashset.Value.Count > F`) no longer skips that tx, so the primary can include it again. Honest backups still reject it (`AddTransaction` → `RequestChangeView`), producing repeated view changes — the stall #984 was meant to stop. One Byzantine validator can do this with a single `TxInvalid`/`TxRejectedByPolicy` ChangeView (`NewViewNumber` only has to be higher than that validator's last CV). `Reset(0)` does not `Clear` the cache, and `MemPool_TransactionRemoved` only drops hashes that were actually in the pool, so attacker keys persist across heights. Safety (commit/prepare of an invalid tx) is unchanged; the bounded cache is what makes wipe-by-eviction possible (the old `Dictionary` grew instead of dropping votes). LRU would not save a 50k+ flood of new keys either.

**Suggestion:** Do not cache hashes that are not in the mempool and/or the current proposal (ignore `RejectedHashes` entirely if no PrepareRequest is in hand). Also cap accepted `RejectedHashes` length to what honest nodes send (one hash per CV). Capacity can stay at `MemoryPoolMaxTransactions` once admission is filtered, because `MemPool_TransactionRemoved` already `Remove`s pool hashes.

tests/Neo.Plugins.DBFTPlugin.Tests/UT_ConsensusService.cs
cschuchardt88, line 321: **[suggestion]** The only test update is `InvalidTransactions[rejectedHash].Value` after one `TxInvalid` ChangeView. Nothing fills the cache to `MemoryPoolMaxTransactions`, asserts `TryGet` false after FIFO wrap, or drives `EnsureMaxBlockLimitation` to skip a tx once `Value.Count > F`. The skip path in `ConsensusContext.MakePayload.cs:87-88` is therefore untested, as is the eviction behavior this PR exists to add.

**Suggestion:** Add a unit test that constructs `InvalidCache` (or a `ConsensusContext`) with a tiny capacity, inserts capacity+1 distinct hashes, and asserts the oldest key is gone. Add another that records `F+1` validator votes on one hash and checks `EnsureMaxBlockLimitation` omits that tx.

plugins/DBFTPlugin/Consensus/ConsensusContext.cs
cschuchardt88, line 30: **[nit]** `UnvalidTxCacheItem` is a typo for Invalid. The record and `InvalidCache` are also public nested types on `ConsensusContext`, which leaks a cache implementation detail from a type that is already a large public surface.

**Suggestion:** Rename to `InvalidTxCacheItem`. Make both nested types `internal` (OnMessage needs the record) if callers only need `InvalidTransactions` lookup/add/remove.

cschuchardt88 (commented):
## Summary

Replaces unbounded `Dictionary<UInt256, HashSet<ECPoint>> InvalidTransactions` with a `FIFOCache` capped at `MemoryPoolMaxTransactions` (default 50_000). Call sites switch to `TryGet` / `UnvalidTxCacheItem.Value`. That bounds memory, but it does not filter `ChangeView.RejectedHashes`, so a single Byzantine CV can evict real votes and re-enable the view-change stall #984 was meant to stop.

## Issues

1. **bug** `plugins/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs:227` — Unfiltered `RejectedHashes` are inserted into a FIFO cache smaller than `ushort.MaxValue`, so one Byzantine ChangeView can evict real votes and stall view changes again. Status: open
2. **suggestion** `tests/Neo.Plugins.DBFTPlugin.Tests/UT_ConsensusService.cs:321` — No test fills the cache, asserts FIFO eviction, or drives `EnsureMaxBlockLimitation` to skip a tx once `Value.Count > F`. Status: open
3. **nit** `plugins/DBFTPlugin/Consensus/ConsensusContext.cs:30` — `UnvalidTxCacheItem` is a typo; nested cache types are public on `ConsensusContext`. Status: open

PR #1092, opened 2026-08-17
Add findstorage, getcandidates and NEP-11 methods for RpcClient

superboyiii:
All tested on my local privatenet.

### DIFF
--- a/plugins/RpcClient/Models/RpcNep11Transfers.cs
+++ b/plugins/RpcClient/Models/RpcNep11Transfers.cs
@@ -1,0 +1,94 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// RpcNep11Transfers.cs file belongs to the neo project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using Neo.Json;
+using Neo.Wallets;
+using System.Numerics;
+
+namespace Neo.Network.RPC.Models;
+
+public class RpcNep11Transfers
+{
+    public UInt160 UserScriptHash { get; set; }
+
+    public List<RpcNep11Transfer> Sent { get; set; }
+
+    public List<RpcNep11Transfer> Received { get; set; }
+
+    public JObject ToJson(ProtocolSettings protocolSettings)
+    {
+        return new()
+        {
+            ["sent"] = Sent.Select(p => p.ToJson(protocolSettings)).ToArray(),
+            ["received"] = Received.Select(p => p.ToJson(protocolSettings)).ToArray(),
+            ["address"] = UserScriptHash.ToAddress(protocolSettings.AddressVersion)
+        };
+    }
+
+    public static RpcNep11Transfers FromJson(JObject json, ProtocolSettings protocolSettings)
+    {
+        return new()
+        {
+            Sent = ((JArray)json["sent"]).Select(p => RpcNep11Transfer.FromJson((JObject)p, protocolSettings)).ToList(),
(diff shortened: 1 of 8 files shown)


vncoelho (dismissed):

plugins/RpcClient/Models/RpcNep11Transfers.cs
```diff
@@ -0,0 +1,94 @@
...
+        };
+    }
+
+    public static RpcNep11Transfer FromJson(JObject json, ProtocolSettings protocolSettings)
+    {
+        return new RpcNep11Transfer
+        {
+            TimestampMS = (ulong)json["timestamp"].AsNumber(),
+            AssetHash = json["assethash"].ToScriptHash(protocolSettings),
+            UserScriptHash = json["transferaddress"]?.ToScriptHash(protocolSettings),
```
shargon, line 86: Why different name?

superboyiii: > Why different name?

Just follow the same way in TokensTracker: https://github.com/neo-project/neo-node/blob/be365a286ec0fee7bba7771dd1dbef6e1a31b06d/plugins/TokensTracker/Trackers/TrackerBase.cs#L140

And the same in RpcNep17Transfers model: https://github.com/neo-project/neo-node/blob/be365a286ec0fee7bba7771dd1dbef6e1a31b06d/plugins/RpcClient/Models/RpcNep17Transfers.cs#L83

shargon (commented):
I would like to see any limits, what's happens if the address contains thousands of records?

plugins/RpcClient/RpcClient.cs
```diff
@@ -702,5 +724,45 @@ public async Task<RpcNep17Balances> GetNep17BalancesAsync(string address)
         return RpcNep17Balances.FromJson((JObject)result, protocolSettings);
     }
 
+    /// <summary>
+    /// Returns all the NEP-11 transaction information occurred in the specified address.
+    /// This method is provided by the plugin TokensTracker.
+    /// </summary>
+    /// <param name="address">The address to query the transaction information.</param>
+    /// <param name="startTimestamp">The start block Timestamp, default to seven days before UtcNow</param>
```
ajara87:
```suggestion
    /// <param name="startTimestamp">The start block Timestamp, default to 0 (genesis)</param>
```
It says that the default is 7 days but it starts at 0. If the intention is to start 7 days earlier I would initialize it with `DateTime.UtcNow.AddDays(-7).ToTimestampMS();`

superboyiii: @ajara87 Good catch. The comments are not clear. I've improved it. In Rpc Tokenstracker, `getnep17transfers` sends `0` as a `sentinel`, not a literal Unix timestamp. TokensTracker treats `startTime == 0` as “seven days before UtcNow”: https://github.com/neo-project/neo-node/blob/master/plugins/TokensTracker/Trackers/NEP-11/Nep11Tracker.cs#L223-L225 The same to `endTime`. 
So:
```json
{"jsonrpc":"2.0","method":"getnep17transfers","params":["NfM3NJFuDtBwZchLh6DYpk1yPigRNmjcTQ"],"id":1}
```
equals
```json
{"jsonrpc":"2.0","method":"getnep17transfers","params":["NfM3NJFuDtBwZchLh6DYpk1yPigRNmjcTQ", 0, 0],"id":1}
```

ajara87 (approved):

superboyiii:
> I would like to see any limits, what's happens if the address contains thousands of records?

There's limit in TokensTracker, it's not an SDK's business...
https://github.com/neo-project/neo-node/blob/be365a286ec0fee7bba7771dd1dbef6e1a31b06d/plugins/TokensTracker/TokensTracker.json#L5

shargon (approved):

cschuchardt88 (approved):

PR #1093, opened 2026-08-27
Skip in plugins

shargon:


roman-khimov (commented):
It's an improvement over the previous state, yet still an API change would be more efficient.

plugins/LevelDBStore/IO/Data/LevelDB/Helper.cs
```diff
@@ -16,14 +16,17 @@ namespace Neo.IO.Data.LevelDB;
 
 public static class Helper
 {
-    public static IEnumerable<(byte[], byte[])> Seek(this DB db, ReadOptions options, byte[]? keyOrPrefix, SeekDirection direction)
+    public static IEnumerable<(byte[], byte[])> Seek(this DB db, ReadOptions options, byte[]? keyOrPrefix, SeekDirection direction, int skip = 0)
```
cschuchardt88, line 19: **[bug]** `IReadOnlyStore.Find` requires implementations to reject negative `skip` (`ArgumentOutOfRangeException`; xmldoc: “must reject negative values instead of treating them as zero”). MemoryStore/MemorySnapshot/`DataCache` all call `ArgumentOutOfRangeException.ThrowIfNegative(skip)` first. `Helper.Seek` (and therefore LevelDB `Store.Find` / `Snapshot.Find`) never validates: `Skip`/`SkipPrev` are `while (count-- > 0 && Valid())`, so `skip < 0` is a no-op and the scan returns every seek-order key. Direct `IStore.Find(..., skip: -1)` on LevelDB therefore diverges from MemoryStore (which throws on first `MoveNext()` even for empty-prefix backward).

**Suggestion:** At the start of `Seek` (before creating the iterator), add `ArgumentOutOfRangeException.ThrowIfNegative(skip);` to match MemoryStore. Do not put the check only inside `Skip` unless `SkipPrev` gets the same check — backward scans never call `Skip`.

plugins/RocksDBStore/Plugins/Storage/Store.cs
```diff
@@ -40,17 +40,27 @@ public IStoreSnapshot GetSnapshot()
     }
 
     /// <inheritdoc/>
-    public IEnumerable<(byte[] Key, byte[] Value)> Find(byte[]? keyOrPrefix, SeekDirection direction = SeekDirection.Forward)
+    public IEnumerable<(byte[] Key, byte[] Value)> Find(byte[]? keyOrPrefix, SeekDirection direction = SeekDirection.Forward, int skip = 0)
```
cschuchardt88, line 44: **[bug]** Same contract miss as LevelDB: `Find` forwards `skip` straight into `it.Skip` / `it.SkipPrev` with no `ThrowIfNegative`. Negative offsets are silently treated as zero, so RocksDB `IStore.Find` does not match MemoryStore or the `IReadOnlyStore` exception contract. `StoreCache.SeekInternal` does forward `skip` into this method when the overlay is empty, but DataCache already validates; the hole is any direct plugin-store caller (and keeping engine implementations interchangeable).

**Suggestion:** Call `ArgumentOutOfRangeException.ThrowIfNegative(skip);` at the start of `Find`, before `NewIterator()`.

plugins/RocksDBStore/Plugins/Storage/Snapshot.cs
```diff
@@ -59,18 +59,28 @@ public void Put(byte[] key, byte[] value)
     }
 
     /// <inheritdoc/>
-    public IEnumerable<(byte[] Key, byte[] Value)> Find(byte[]? keyOrPrefix, SeekDirection direction)
+    public IEnumerable<(byte[] Key, byte[] Value)> Find(byte[]? keyOrPrefix, SeekDirection direction = SeekDirection.Forward, int skip = 0)
```
cschuchardt88, line 68: **[bug]** `Snapshot.Find` is a copy of `Store.Find` and also omits negative-`skip` validation. Fixing only `Store` would leave snapshot reads (the path `DataCache`/`IStoreSnapshot` actually uses) inconsistent.

**Suggestion:** Same `ArgumentOutOfRangeException.ThrowIfNegative(skip);` at the start of `Snapshot.Find`.

plugins/RocksDBStore/Extensions/IteratorExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+internal static class Iterator
+{
+    /// <summary>
+    /// Advances the iterator <paramref name="count"/> entries by calling the native
+    /// rocksdb_iter_next repeatedly (rocksdb has no batch-skip primitive).
+    /// Does not marshal keys/values while skipping.
+    /// </summary>
+    public static void Skip(this RocksDbSharp.Iterator it, int count)
+    {
+        while (count-- > 0 && it.Valid())
```
cschuchardt88, line 29: **[suggestion]** `tests/Neo.Plugins.Storage.Tests/StoreTest.cs` already drives LevelDB and RocksDB `Find` forward/backward (store and snapshot) but never passes a non-zero `skip`. Core `UT_MemoryStore` / `UT_DataCache` cover MemoryStore and overlay merge, not engine iterators (`Seek` then `Skip`, `SeekForPrev`/`SkipPrev`, LevelDB’s Seek+Prev positioning). Codecov on this PR also shows the `count > 0` branches of `Skip`/`SkipPrev` as uncovered.

**Suggestion:** Extend `TestStorage` with shared cases for both engines: forward/backward `skip: 1` from an existing seek key, `skip` greater than remaining keys (empty result), and negative `skip` throwing `ArgumentOutOfRangeException`. Include a backward seek that lands off an exact key (e.g. `[0x00, 0x00, 0x03]` when that key is missing) so LevelDB’s Prev adjustment is counted before the offset.

cschuchardt88 (changes requested):

shargon:
> It's an improvement over the previous state, yet still an API change would be more efficient.

I can't see any method to skip entries in the native libraries, could you give me a hint?

roman-khimov:
There is none, what I mean is RPC API that triggers the need to skip. It shouldn't have any "skip" parameters, rather it should bring with it an initial offset. Then the first Seek moves us right in place. Much like in https://github.com/neo-project/neo/issues/4039

shargon:
@superboyiii merge?

superboyiii:
Wait for my final test today.

plugins/RpcServer/RpcServer.Blockchain.cs
superboyiii, line 461:
```suggestion
        base64KeyPrefix.NotNull_Or(RpcError.InvalidParams.WithData($"Invalid 'base64KeyPrefix'"));
        (start >= 0).True_Or(RpcError.InvalidParams.WithData($"Invalid 'start': {start}"));
```
Previously `FindStorage` used LINQ `.Skip(start)`, which treats a negative `start` as zero and still returns a successful page. After switching to `FindContractStorage(..., skip: start)`, the same call throws `ArgumentOutOfRangeException`, which RpcServer maps to a generic HResult error instead of `InvalidParams`. Adding an explicit `start >= 0` check keeps the stricter rejection, but returns the same clean `InvalidParams` response used for other bad RPC arguments.

plugins/RocksDBStore/RocksDBStore.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="rocksdb" Version="10.10.1.1747" />
+    <PackageReference Include="rocksdb" Version="11.8.1.4423" />
```
ajara87, line 10: Why is the RocksDB update needed?

superboyiii:
Tested. Others are good.

plugins/RocksDBStore/RocksDBStore.csproj
shargon: Storage related

ajara87 (dismissed):

superboyiii (dismissed):

superboyiii (approved):

plugins/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -471,7 +472,7 @@ protected internal virtual JToken FindStorage(ContractNameOrHashOrId contractNam
         var items = new JArray();
         int pageSize = settings.FindStoragePageSize;
         int i = 0;
-        using (var iter = NativeContract.ContractManagement.FindContractStorage(snapshot, id, prefix).Skip(count: start).GetEnumerator())
+        using (var iter = NativeContract.ContractManagement.FindContractStorage(snapshot, id, prefix, skip: start).GetEnumerator())
```
AnnaShaleva, line 475: Consider the case:
1. Call `findstorage` once, iterate the result (let it be `range1`), see that some items are still left in the iterator.
2. Update the contract storage in a subsequent block, add (or remove) some entries from the `range1`.
3. Call `findstorage` one more time with `skip` parameter based on the old `range1` number of items. Get `range2` range of KVs.

The result is: either missing or duplicating items in `range2`.

So the problem is in the approach itself. We may change the approach (i.e. pass start *bytes* instead of the number of items to pass), but it also doesn't solve the problem completely. What do you think, does this API change worth the effort? Or may be there's another approach that may solve the problem completely?

AnnaShaleva (commented):
The code itself is good, but there's still a problem remaining with the API.

plugins/RocksDBStore/RocksDBStore.csproj
cschuchardt88, line 10: **[suggestion]** This is a native major-version bump (`10.10.1.1747` → `11.8.1.4423`), not something skip needs. `Skip`/`SkipPrev` only call `Next()`/`Prev()`, which RocksDbSharp already had. The bump landed in the "add ut" commit with no notes on why 11 is required.

Please split it out. A RocksDB 11 native ABI change is a different risk (node operators, existing DBs, platform binaries) than iterator skip, and it makes regressions from this PR harder to bisect.

plugins/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -459,6 +459,7 @@ protected internal virtual JToken FindStorage(ContractNameOrHashOrId contractNam
     {
         contractNameOrHashOrId.NotNull_Or(RpcError.InvalidParams.WithData($"Invalid 'contractNameOrHashOrId'"));
         base64KeyPrefix.NotNull_Or(RpcError.InvalidParams.WithData($"Invalid 'base64KeyPrefix'"));
+        (start >= 0).True_Or(RpcError.InvalidParams.WithData($"Invalid 'start': {start}"));
 
         using var snapshot = system.GetSnapshotCache();
         int id = GetContractId(snapshot, contractNameOrHashOrId);
```
cschuchardt88, line 465: **[suggestion]** `GetContractId` for hash/name calls `ContractManagement.GetContract` on this same `StoreCache`, which inserts at least one key into `DataCache._dictionary`. `DataCache.Seek` then takes the `cachedKeySet.Count > 0` branch and applies `skip` in-memory with `SeekInternal(..., skip: 0)`, so the plugin `Skip` this PR adds is not used for the common `findstorage(hash, ...)` path — skipped keys are still marshalled from LevelDB/RocksDB.

Resolving the id on a separate snapshot (or the `IsId` fast path) and calling `FindContractStorage` on a fresh empty cache would actually hit native skip. Worth doing if the RPC change is meant to avoid that marshalling.

cschuchardt88 (commented):
## Summary

Plugin `IStore` / `IStoreSnapshot.Find` now takes `skip` and applies it on the native iterator after Seek, matching `IReadOnlyStore` in Neo `3.10.2-CI02099`. Negative `skip` throws. Tests cover both engines, both directions, overskip, negatives, and snapshots. `findstorage` maps `start < 0` to `InvalidParams`.

Previous review items from this account are **Fixed** (negative skip was a no-op; no engine skip tests). Copilot rename / nullable stub / tests are also addressed.

`IStore.Find(skip)` is required by the Neo package interface. Native DBs have no batch-skip primitive, so Next/Prev without marshalling keys is the right engine change.

## Other reviewers

- **roman-khimov** (start key instead of skip-N): **Agreed as missed optimization**. That is an RPC/`findstorage` shape change ([neo#4039](https://github.com/neo-project/neo/issues/4039)), not something LevelDB/RocksDB expose as skip. Not a blocker for plugin `Find(skip)`.
- **AnnaShaleva** (index `start` duplicates/misses across writes): **Agreed (pre-existing RPC model)**. `findstorage` already paginated by count; this PR only moves that skip into the store. A start-key/`next` cursor is a follow-up API.
- **ajara87** (RocksDB bump): still open — skip does not need RocksDB 11.
- **superboyiii** (`start >= 0`): **Fixed**.

## Issues

Two suggestions below (RocksDB 11 bump mixed into this PR; `findstorage` hash/name path still marshals skipped keys because `GetContractId` fills `DataCache`).

plugins/RpcServer/RpcServer.Blockchain.cs
shargon: moved to a new issue

shargon: moved to a new issue

PR #1094, opened 2026-08-28
Fix Oracle HTTPS Content-Type NRE and relative redirects

cschuchardt88:
## Summary

`OracleHttpsProtocol.ProcessAsync` crashed with `NullReferenceException` when the HTTP response had no `Content-Type` header (the null check sat *after* the try/catch). It also treated relative `Location` values as non-HTTPS, so in-site HTTPS redirects failed.

## Changes

- Treat a missing `Content-Type` / media type as `ContentTypeNotSupported`.
- Resolve relative redirect URIs against the current request URI; still reject non-HTTPS.
- Dispose the redirect response before following `Location`.
- `Https.MaxRedirects` in `OracleService.json` (default **2**, same as the old hardcoded limit). Negative values clamp to 0.

## Tests

- Absolute HTTPS redirect kept as-is; relative path resolved; HTTP/FTP rejected
- Missing `Content-Type` rejected; `application/json` allowed; `text/html` rejected
- Encoding defaults to UTF-8; declared charset is honored
- `MaxRedirects` default, config value, and negative clamp

### DIFF
--- a/plugins/OracleService/OracleSettings.cs
+++ b/plugins/OracleService/OracleSettings.cs
@@ -16,11 +16,23 @@ namespace Neo.Plugins.OracleService;
 
 class HttpsSettings
 {
+    /// <summary>
+    /// Default number of HTTPS redirects to follow after the first request.
+    /// </summary>
+    public const int DefaultMaxRedirects = 2;
+
     public TimeSpan Timeout { get; }
+
+    /// <summary>
+    /// Maximum number of HTTPS redirects to follow. Zero means the first response is used as-is
+    /// (a <c>Location</c> header is not followed). Negative config values are treated as zero.
+    /// </summary>
+    public int MaxRedirects { get; }
 
     public HttpsSettings(IConfigurationSection section)
     {
         Timeout = TimeSpan.FromMilliseconds(section.GetValue("Timeout", 5000));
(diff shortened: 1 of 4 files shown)


plugins/OracleService/OracleSettings.cs
```diff
@@ -16,11 +16,23 @@ namespace Neo.Plugins.OracleService;
...
+    /// </summary>
+    public const int DefaultMaxRedirects = 2;
+
     public TimeSpan Timeout { get; }
 
+    /// <summary>
+    /// Maximum number of HTTPS redirects to follow. Zero means the first response is used as-is
+    /// (a <c>Location</c> header is not followed). Negative config values are treated as zero.
+    /// </summary>
+    public int MaxRedirects { get; }
```
roman-khimov, line 30: Do we really want it to be configurable?

roman-khimov (commented):
Nice fixes otherwise.

plugins/OracleService/OracleSettings.cs
cschuchardt88: wasn't sure....

shargon (approved):

ajara87 (approved):

roman-khimov (approved):

PR #1095, opened 2026-08-28
Return empty array from getcandidates when the VM stack is empty

cschuchardt88:
## Summary

`getcandidates` is documented as returning a JSON array. When `ApplicationEngine.ResultStack` was empty, the method returned a new `JObject` (`{}`). That is not an array and breaks clients.

## Change

Return `new JArray()` on the empty-stack path. Candidate JSON is built by `FormatCandidates` so the empty-stack path is unit-tested.

## Tests

- Empty VM stack → `[]`
- Empty inner array → `[]`
- Votes as string, `active` true/false
- Invalid stack item throws
- Existing populated `getcandidates` still matches native results

### DIFF
--- a/plugins/RpcServer/RpcServer.Blockchain.cs
+++ b/plugins/RpcServer/RpcServer.Blockchain.cs
@@ -590,7 +590,7 @@ partial class RpcServer
                 ?? throw new RpcException(RpcError.InternalServerError.WithData("Can't get next block validators."));
             return FormatCandidates(resultstack, validators);
         }
-        catch (Exception ex) when (ex is not RpcException)
+        catch
         {
             throw new RpcException(RpcError.InternalServerError.WithData("Can't get next block validators"));
         }
--- a/plugins/RpcServer/RpcServer.Blockchain.cs
+++ b/plugins/RpcServer/RpcServer.Blockchain.cs
@@ -581,37 +581,51 @@ partial class RpcServer
             throw new RpcException(RpcError.InternalServerError.WithData("Can't get candidates."));
         }
 
-        JObject json = new();
         try
         {
-            if (resultstack.Length > 0)
-            {
-                JArray jArray = new();
-                var validators = NativeContract.NEO.GetNextBlockValidators(snapshot, system.Settings.ValidatorsCount)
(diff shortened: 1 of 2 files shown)


roman-khimov:
Clients can rely on current behavior for uninited networks.

cschuchardt88:
> Clients can rely on current behavior for uninited networks.

@roman-khimov does `neo-go` return empty `JObject` as well?

roman-khimov:
Good point, seems like it doesn't, successful results are always arrays (even if empty). In that sense I'm OK with this change.

plugins/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -581,37 +581,51 @@ protected internal virtual JToken GetCandidates()
...
-                        json = new();
-                    }
-                    return jArray;
-                }
-            }
+            var validators = NativeContract.NEO.GetNextBlockValidators(snapshot, system.Settings.ValidatorsCount)
+                ?? throw new RpcException(RpcError.InternalServerError.WithData("Can't get next block validators."));
+            return FormatCandidates(resultstack, validators);
+        }
+        catch (RpcException)
```
shargon: This catch can be removed, already have one

plugins/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -581,37 +581,47 @@ protected internal virtual JToken GetCandidates()
...
-                        json["votes"] = ele[1].GetInteger().ToString();
-                        json["active"] = validators.Contains(publickey);
-                        jArray.Add(json);
-                        json = new();
-                    }
-                    return jArray;
-                }
-            }
+            var validators = NativeContract.NEO.GetNextBlockValidators(snapshot, system.Settings.ValidatorsCount)
+                ?? throw new RpcException(RpcError.InternalServerError.WithData("Can't get next block validators."));
```
cschuchardt88:
```suggestion
                ?? throw new Exception();
```

shargon (approved):

ajara87 (approved):

roman-khimov (approved):

PR #1096, opened 2026-08-28
Report neo-node version from getversion (neo-node#983)

### LINKED ISSUE #983: neo-cli v3.9.2 returns incorrect version (3.9.1)
The published version of neo-cli `v3.9.2` reports its version as `3.9.1` when queried through the `getversion` RPC method.

Image

This appears to be caused by a mismatch in package references, as the neo package dependency has not been updated to `3.9.2` within the CLI project.

Since the RPC Plugin retrieves the version string from `LocalNode.UserAgent`, and this value is derived from the assembly version of the core neo library rather than the CLI, the output remains outdated.
https://github.com/neo-project/neo-node/blob/master-n3/plugins/RpcServer/RpcServer.Node.cs#L153

Image

cschuchardt88:
## Summary

`getversion` reported `useragent` from `LocalNode.UserAgent`, which is the **Neo.dll** assembly, not neo-cli / RpcServer. After a node bump (e.g. 3.9.2 / 3.10.2) the RPC still advertised the older library version.

Fixes #983

## Change

No new JSON field. The existing `useragent` string keeps the product name and uses the RpcServer assembly version (`VersionPrefix`), so `/Neo:3.10.1/` becomes `/Neo:3.10.2/`. Malformed useragents fall back to `/Neo:{version}/`.

## Tests

- `getversion` has no `version` property
- `useragent` contains the RpcServer 3-part assembly version
- `UserAgentWithVersion` keeps the product name and replaces the version
- `rpc.maxiteratorresultitems` / `rpc.sessionenabled` match settings

### DIFF
--- a/plugins/RpcServer/RpcServer.Node.cs
+++ b/plugins/RpcServer/RpcServer.Node.cs
@@ -123,8 +123,7 @@ partial class RpcServer
     ///   "result": {
     ///     "tcpport": 10333, // The TCP port,
     ///     "nonce": 1, // The nonce,
-    ///     "useragent": "The Neo library user agent",
-    ///     "version": "The neo-node / RpcServer assembly version",
+    ///     "useragent": "/Neo:3.10.2/", // Node assembly version in the existing useragent field,
     ///     "rpc": {
     ///       "maxiteratorresultitems": 100, // The maximum number of items in the iterator result,
     ///       "sessionenabled": false // Whether session is enabled,
@@ -153,9 +152,8 @@ partial class RpcServer
         JObject json = new();
         json["tcpport"] = localNode.ListenerTcpPort;
         json["nonce"] = LocalNode.Nonce;
-        json["useragent"] = LocalNode.UserAgent;
-        // UserAgent comes from Neo.dll; version is this node/plugin assembly (neo-node#983).
-        json["version"] = typeof(RpcServer).Assembly.GetName().Version?.ToString(3);
+        var nodeVersion = typeof(RpcServer).Assembly.GetName().Version?.ToString(3) ?? "0.0.0";
+        json["useragent"] = UserAgentWithVersion(LocalNode.UserAgent, nodeVersion);
         // rpc settings
         JObject rpc = new();
         rpc["maxiteratorresultitems"] = settings.MaxIteratorResultItems;
@@ -184,6 +182,19 @@ partial class RpcServer
         json["rpc"] = rpc;
         json["protocol"] = protocol;
         return json;
+    }
+
+    /// <summary>
+    /// Keeps the P2P useragent product name and replaces the version with <paramref name="version"/>.
+    /// <c>/Neo:3.10.1/</c> plus <c>3.10.2</c> becomes <c>/Neo:3.10.2/</c>.
+    /// </summary>
+    internal static string UserAgentWithVersion(string userAgent, string version)
+    {
+        var colon = userAgent.IndexOf(':');
+        var slash = userAgent.LastIndexOf('/');
+        if (colon >= 0 && slash > colon)
+            return string.Concat(userAgent.AsSpan(0, colon + 1), version, userAgent.AsSpan(slash));
+        return $"/Neo:{version}/";
     }
 
     /// <summary>
--- a/plugins/RpcServer/RpcServer.Node.cs
+++ b/plugins/RpcServer/RpcServer.Node.cs
@@ -123,7 +123,8 @@ partial class RpcServer
     ///   "result": {
     ///     "tcpport": 10333, // The TCP port,
     ///     "nonce": 1, // The nonce,
-    ///     "useragent": "The user agent",
+    ///     "useragent": "The Neo library user agent",
+    ///     "version": "The neo-node / RpcServer assembly version",
     ///     "rpc": {
     ///       "maxiteratorresultitems": 100, // The maximum number of items in the iterator result,
     ///       "sessionenabled": false // Whether session is enabled,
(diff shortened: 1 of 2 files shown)


roman-khimov (changes requested):
useragent can be just about anything, I'd rather not introduce new fields.

cschuchardt88:
Dropped the extra `version` field per review.

`getversion` now puts the node / RpcServer assembly version into the existing `useragent` string (same `/Name:x.y.z/` shape). `/Neo:3.10.1/` from Neo.dll becomes `/Neo:3.10.2/` when the node is 3.10.2. No new JSON keys.

shargon (commented):
LocalNode.UserAgent should be changed this is only the relayer

plugins/RpcServer/RpcServer.Node.cs
```diff
@@ -152,7 +152,8 @@ protected internal virtual JToken GetVersion()
         JObject json = new();
         json["tcpport"] = localNode.ListenerTcpPort;
         json["nonce"] = LocalNode.Nonce;
-        json["useragent"] = LocalNode.UserAgent;
+        var nodeVersion = typeof(RpcServer).Assembly.GetName().Version?.ToString(3) ?? "0.0.0";
+        json["useragent"] = UserAgentWithVersion(LocalNode.UserAgent, nodeVersion);
```
cschuchardt88, line 156: `getversion` is only a relayer of `LocalNode.UserAgent`. That property is already `public static string UserAgent { get; set; }` on Neo 3.10.1.

Rewriting it here leaves the P2P `Version` handshake on the Neo.dll version, so RPC and P2P disagree. Set the node version once in neo-cli before `StartNode` (CLI already reads `Assembly.GetAssembly(typeof(Program))!.GetName().Version`) and keep `json["useragent"] = LocalNode.UserAgent`.

This also avoids taking the version from RpcServer.dll, which can be loaded independently of neo-cli (issue #983 is the CLI version).

plugins/RpcServer/RpcServer.Node.cs
```diff
@@ -183,6 +184,19 @@ protected internal virtual JToken GetVersion()
...
+    /// <c>/Neo:3.10.1/</c> plus <c>3.10.2</c> becomes <c>/Neo:3.10.2/</c>.
+    /// </summary>
+    internal static string UserAgentWithVersion(string userAgent, string version)
+    {
+        var colon = userAgent.IndexOf(':');
+        var slash = userAgent.LastIndexOf('/');
+        if (colon >= 0 && slash > colon)
+            return string.Concat(userAgent.AsSpan(0, colon + 1), version, userAgent.AsSpan(slash));
+        return $"/Neo:{version}/";
+    }
```
cschuchardt88, line 198: This is not a version parse. First `:` and last `/` rewrite any string that happens to have that shape; anything else is replaced with `/Neo:{version}/`.

`LocalNode.UserAgent` is publicly settable. Examples:
- `/Neo:3.10.1/extra/` → `/Neo:3.10.2/` (suffix dropped because last `/` is the trailer)
- `not-a-ua` → `/Neo:3.10.2/` (custom product name discarded)

If this helper stays, match the `/Name:x.y.z/` form and otherwise leave `userAgent` unchanged (or only replace when it is the default Neo.dll value).

tests/Neo.Plugins.RpcServer.Tests/UT_RpcServer.Node.cs
```diff
@@ -117,6 +117,17 @@ public void TestGetVersion()
         Assert.IsTrue(json.ContainsProperty("tcpport"));
         Assert.IsTrue(json.ContainsProperty("nonce"));
         Assert.IsTrue(json.ContainsProperty("useragent"));
+        Assert.IsFalse(json.ContainsProperty("version"));
+        var nodeVersion = _rpcServer.GetType().Assembly.GetName().Version?.ToString(3) ?? "0.0.0";
+        Assert.AreEqual(RpcServer.UserAgentWithVersion(LocalNode.UserAgent, nodeVersion), json["useragent"]!.AsString());
```
cschuchardt88, line 122: This asserts `GetVersion` against the same helper `GetVersion` calls, so it cannot fail unless the helper is skipped. Assert the concrete string (e.g. `$"/Neo:{nodeVersion}/"`) instead.

cschuchardt88 (commented):
## Summary

`getversion` still treats `useragent` as the P2P identity string, but this change rewrites only the RPC copy. `LocalNode.UserAgent` is already `{ get; set; }` in Neo 3.10.1; neo-cli already knows its own assembly version in `MainService`. Patching RpcServer leaves the handshake on Neo.dll and can mangle a custom useragent.

## Issues

- suggestion `plugins/RpcServer/RpcServer.Node.cs:156` Status: open
  GetVersion should keep relaying `LocalNode.UserAgent`. Set the node/CLI version on `LocalNode.UserAgent` before `StartNode` so P2P and RPC match. Do not take the version from RpcServer.dll; issue #983 is neo-cli.

- bug `plugins/RpcServer/RpcServer.Node.cs:196` Status: open
  `UserAgentWithVersion` is not a version parse. First `:` plus last `/` rewrites any string with that shape; anything else becomes `/Neo:{version}/`. `LocalNode.UserAgent` is publicly settable, so custom values are corrupted or replaced.

- nit `tests/Neo.Plugins.RpcServer.Tests/UT_RpcServer.Node.cs:122` Status: open
  `TestGetVersion` asserts against `UserAgentWithVersion` itself, so it cannot catch a bad rewrite. Assert the concrete `/Neo:{nodeVersion}/` string.

vncoelho:
> useragent can be just about anything, I'd rather not introduce new fields.

I am in this same line of reasoning, new field.

PR #1097, opened 2026-08-28
Use VMInstruction.ToString for parse script listings

cschuchardt88:
## Summary

`parse` / `ScriptsToOpCode` now uses a single `VMInstruction.ToString()` layout, and `DecodeOperand` annotates operands instead of dumping raw hex.

## Changes

- `ToString()` is `{position} {opcode}` plus operand only when `OperandSize > 0`. Hex position width is at least 4 and grows with script length.
- `ScriptsToOpCode` prefixes that with `L{n}:`. Decimal width is at least 4 (`D04`) and grows with the script size / instruction count (so listings longer than 9999 lines still line up).
- Strict UTF-8 `PUSHDATA*` text (`Utf8.IsValid`) is shown as `hex // text`. Control runes are escaped (`\n`, `\r`, `\t`, `\0`, `\xNN`) so they do not create real newlines in the listing.
- Otherwise, typed comments:
  - 33/65-byte `ECPoint` ΓåÆ `hex // {point}`
  - 20-byte `UInt160` / 32-byte `UInt256` ΓåÆ `hex // {hash}`
  - unix seconds or milliseconds in 2000ΓÇô2100 on `PUSHINT32`/`PUSHINT64` and 4/8-byte `PUSHDATA`
- Remaining payloads: `hex // blob N bytes`.

## Example

`parse` listing of this Base64 `System.Contract.Call(updateFeeds, ΓÇª)` script:

```
L0000:0000 PUSH0
L0001:0001 PUSH0
L0002:0002 PUSH0
L0003:0003 PUSH0
L0004:0004 PUSH0
L0005:0005 PUSH0
L0006:0006 PUSH0
L0007:0007 PUSH0
L0008:0008 PUSH0
L0009:0009 PUSH0
L0010:000A PUSH0
L0011:000B PUSH0
L0012:000C PUSH12
L0013:000D PACK
L0014:000E PUSHDATA1 71BDDFD76DBDEF67BCF1C71AE77E787B973C69F79C79FF1F // blob 24 bytes
L0015:0028 PUSHDATA1 DF4EBBF5E7FDD1FD1BE7DE7979CF3C79A7F879F6F5F7879F // blob 24 bytes
L0016:0042 PUSHDATA1 EFDE7ADDBDDDDF5F7D7396F46BAEDE79AEF6F5B7F4DDAE36 // blob 24 bytes
L0017:005C PUSHDATA1 73B6DDDBAD5EE3DE9DDF969EE79E766FCDDE75AD7CF7AE75 // blob 24 bytes
L0018:0076 PUSHDATA1 D9ED1CEB8DFDD37EB469BF36F7ADB7774EDF77479DE3DF7D // blob 24 bytes
L0019:0090 PUSHDATA1 EB87B4EBC6B67FA7BDF7871E77BF5AD9EE1F7DBEDFEBBF7D // blob 24 bytes
L0020:00AA PUSHDATA1 79E73CEF869D778F1D79B79BD38D1CDF6D9E6F875CDB5775 // blob 24 bytes
L0021:00C4 PUSHDATA1 75B69AEB4F3C777D5B6DAF7AD1EF746B8EF5E5FE3AE37EFB // blob 24 bytes
L0022:00DE PUSHDATA1 DFD6F7E77DF4D38F35EDDD3AE39EDAD5F7F77767F4F766FC // blob 24 bytes
L0023:00F8 PUSHDATA1 739F5DD9ED34E9EE1CDB6D7DDF5F1FEF47F67BADF675FEB8 // blob 24 bytes
L0024:0112 PUSHDATA1 EDEE1BD78D7CEF6EDDF5EEBBF1EE1E6DAE9AF1B77C7F5EBB // blob 24 bytes
L0025:012C PUSHDATA1 D5CE3BE39E76F77F1B6FD738D9BD3B7F67BB71ADDE71C774 // blob 24 bytes
L0026:0146 PUSH12
L0027:0147 PACK
L0028:0148 PUSHINT32 1787908101
L0029:014D PUSHINT32 1787908101
L0030:0152 PUSHINT32 1787908101
L0031:0157 PUSHINT32 1787908101
L0032:015C PUSHINT32 1787908101
L0033:0161 PUSHINT32 1787908101
L0034:0166 PUSHINT32 1787908101
L0035:016B PUSHINT32 1787908101
L0036:0170 PUSHINT32 1787908101
L0037:0175 PUSHINT32 1787908101
L0038:017A PUSHINT32 1787908101
L0039:017F PUSHINT32 1787908101
L0040:0184 PUSH12
L0041:0185 PACK
L0042:0186 PUSHINT32 148800
L0043:018B PUSHINT16 6265
L0044:018E PUSHINT32 771039978
L0045:0193 PUSHINT32 504880000
L0046:0198 PUSHINT32 87170
L0047:019D PUSHINT32 1417300
L0048:01A2 PUSHINT32 999930
L0049:01A7 PUSHINT32 1000020
L0050:01AC PUSHINT32 339700
L0051:01B1 PUSHINT64 2491470000
L0052:01BA PUSHINT64 79426600000
L0053:01C3 PUSHINT32 2162000
L0054:01C8 PUSH12
L0055:01C9 PACK
L0056:01CA PUSHINT32 1787908101
L0057:01CF PUSHINT32 1787908101
L0058:01D4 PUSHINT32 1787908101
L0059:01D9 PUSHINT32 1787908101
L0060:01DE PUSHINT32 1787908101
L0061:01E3 PUSHINT32 1787908101
L0062:01E8 PUSHINT32 1787908101
L0063:01ED PUSHINT32 1787908101
L0064:01F2 PUSHINT32 1787908101
L0065:01F7 PUSHINT32 1787908101
L0066:01FC PUSHINT32 1787908101
L0067:0201 PUSHINT32 1787908101
L0068:0206 PUSH12
L0069:0207 PACK
L0070:0208 PUSHDATA1 5457454C5645444154413A434E592D555344 // TWELVEDATA:CNY-USD
L0071:021C PUSHDATA1 5457454C5645444154413A4A50592D555344 // TWELVEDATA:JPY-USD
L0072:0230 PUSHDATA1 5457454C5645444154413A5350592D555344 // TWELVEDATA:SPY-USD
L0073:0244 PUSHDATA1 5457454C5645444154413A4D5346542D555344 // TWELVEDATA:MSFT-USD
L0074:0259 PUSHDATA1 5457454C5645444154413A444F47452D555344 // TWELVEDATA:DOGE-USD
L0075:026E PUSHDATA1 5457454C5645444154413A5852502D555344 // TWELVEDATA:XRP-USD
L0076:0282 PUSHDATA1 5457454C5645444154413A555344432D555344 // TWELVEDATA:USDC-USD
L0077:0297 PUSHDATA1 5457454C5645444154413A555344542D555344 // TWELVEDATA:USDT-USD
L0078:02AC PUSHDATA1 5457454C5645444154413A5452582D555344 // TWELVEDATA:TRX-USD
L0079:02C0 PUSHDATA1 5457454C5645444154413A4554482D555344 // TWELVEDATA:ETH-USD
L0080:02D4 PUSHDATA1 5457454C5645444154413A4254432D555344 // TWELVEDATA:BTC-USD
L0081:02E8 PUSHDATA1 5457454C5645444154413A4E454F2D555344 // TWELVEDATA:NEO-USD
L0082:02FC PUSH12
L0083:02FD PACK
L0084:02FE PUSH6
L0085:02FF PACK
L0086:0300 PUSH15
L0087:0301 PUSHDATA1 7570646174654665656473 // updateFeeds
L0088:030E PUSHDATA1 ABCC7F51C334D4F958BE8B6C54142AC4493F0103 // 0x03013f49c42a14546c8bbe58f9d434c3517fccab
L0089:0324 SYSCALL   [System.Contract.Call]
```

### DIFF
--- a/src/Neo.CLI/CLI/MainService.Tools.cs
+++ b/src/Neo.CLI/CLI/MainService.Tools.cs
@@ -409,7 +409,7 @@ partial class MainService
     /// <summary>
     /// Base64 Smart Contract Script Analysis
     /// input: DARkYXRhAgBlzR0MFPdcrAXPVptVduMEs2lf1jQjxKIKDBT3XKwFz1abVXbjBLNpX9Y0I8SiChTAHwwIdHJhbnNmZXIMFKNSbimM12LkFYX/8KGvm2ttFxulQWJ9W1I=
-    /// output (VMInstruction.ToString, with a line number prefix):
+    /// output (VMInstruction.ToString, with an L-prefix padded to at least 4 digits):
     /// L0000:0000 PUSHDATA1 64617461 // data
     /// L0001:0006 PUSHINT32 500000000
     /// L0002:000B PUSHDATA1 0AA2C42334D65F69B304E376559B56CF05AC5CF7
@@ -420,17 +420,7 @@ partial class MainService
         try
         {
             var bytes = Convert.FromBase64String(base64);
-            var sb = new StringBuilder();
-            var line = 0;
-
-            foreach (var instruct in new VMInstruction(bytes))
-            {
-                sb.AppendFormat("L{0:D04}:{1}", line, instruct);
-                sb.AppendLine();
-                line++;
-            }
-
-            return sb.ToString();
+            return VMInstruction.FormatListing(bytes);
         }
         catch
         {
--- a/src/Neo.CLI/CLI/MainService.Tools.cs
+++ b/src/Neo.CLI/CLI/MainService.Tools.cs
@@ -409,17 +409,10 @@ partial class MainService
     /// <summary>
     /// Base64 Smart Contract Script Analysis
     /// input: DARkYXRhAgBlzR0MFPdcrAXPVptVduMEs2lf1jQjxKIKDBT3XKwFz1abVXbjBLNpX9Y0I8SiChTAHwwIdHJhbnNmZXIMFKNSbimM12LkFYX/8KGvm2ttFxulQWJ9W1I=
-    /// output:
-    /// PUSHDATA1 data
-    /// PUSHINT32 500000000
-    /// PUSHDATA1 0x0aa2c42334d65f69b304e376559b56cf05ac5cf7
-    /// PUSHDATA1 0x0aa2c42334d65f69b304e376559b56cf05ac5cf7
-    /// PUSH4
-    /// PACK
-    /// PUSH15
-    /// PUSHDATA1 transfer
-    /// PUSHDATA1 0xa51b176d6b9bafa1f0ff8515e462d78c296e52a3
-    /// SYSCALL System.Contract.Call
+    /// output (VMInstruction.ToString, with a line number prefix):
+    /// L0000:0000 PUSHDATA1 64617461 // data
+    /// L0001:0006 PUSHINT32 500000000
+    /// L0002:000B PUSHDATA1 0AA2C42334D65F69B304E376559B56CF05AC5CF7
     /// </summary>
     [ParseFunction("Base64 Smart Contract Script Analysis")]
     private string? ScriptsToOpCode(string base64)
@@ -432,10 +425,8 @@ partial class MainService
 
             foreach (var instruct in new VMInstruction(bytes))
             {
-                if (instruct.OperandSize == 0)
-                    sb.AppendFormat("L{0:D04}:{1:X04} {2}{3}", line, instruct.Position, instruct.OpCode, Environment.NewLine);
-                else
-                    sb.AppendFormat("L{0:D04}:{1:X04} {2,-10}{3}{4}", line, instruct.Position, instruct.OpCode, instruct.DecodeOperand(), Environment.NewLine);
+                sb.AppendFormat("L{0:D04}:{1}", line, instruct);
+                sb.AppendLine();
                 line++;
             }
 
--- a/src/Neo.CLI/Tools/VMInstruction.cs
+++ b/src/Neo.CLI/Tools/VMInstruction.cs
@@ -100,10 +100,42 @@ internal sealed class VMInstruction : IEnumerable<VMInstruction>
 
     public override string ToString()
     {
+        var posWidth = HexWidth(_script.Length);
         if (OperandSize == 0)
-            return string.Format("{0:X04} {1}", Position, OpCode);
-        return string.Format("{0:X04} {1,-10}{2}", Position, OpCode, DecodeOperand());
+            return string.Format($"{{0:X{posWidth}}} {{1}}", Position, OpCode);
+        return string.Format($"{{0:X{posWidth}}} {{1,-10}}{{2}}", Position, OpCode, DecodeOperand());
     }
+
+    /// <summary>
+    /// Lists instructions as <c>L0000:0000 NOP</c>. The L-prefix width is at least 4 digits
+    /// and grows with the script size / instruction count.
+    /// </summary>
+    internal static string FormatListing(ReadOnlyMemory<byte> script)
+    {
+        var rows = new List<VMInstruction>();
+        foreach (var instruct in new VMInstruction(script))
+            rows.Add(instruct);
+
+        var lastIndex = Math.Max(0, rows.Count - 1);
+        var width = DecimalWidth(Math.Max(lastIndex, script.Length));
+        var sb = new StringBuilder();
+        for (var i = 0; i < rows.Count; i++)
+        {
+            sb.Append('L');
+            sb.Append(i.ToString($"D{width}", CultureInfo.InvariantCulture));
+            sb.Append(':');
+            sb.Append(rows[i]);
+            sb.AppendLine();
+        }
+
+        return sb.ToString();
+    }
+
+    internal static int DecimalWidth(int value)
+        => Math.Max(4, Math.Max(0, value).ToString(CultureInfo.InvariantCulture).Length);
+
+    internal static int HexWidth(int length)
+        => Math.Max(4, Math.Max(0, length == 0 ? 0 : length - 1).ToString("X").Length);
 
     public T AsToken<T>(uint index = 0)
         where T : unmanaged
--- a/src/Neo.CLI/Tools/VMInstruction.cs
+++ b/src/Neo.CLI/Tools/VMInstruction.cs
@@ -9,11 +9,13 @@
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
+using Neo.Cryptography.ECC;
 using Neo.SmartContract;
 using Neo.VM;
 using System.Buffers.Binary;
 using System.Collections;
 using System.Diagnostics;
+using System.Globalization;
 using System.Numerics;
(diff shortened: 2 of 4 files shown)


src/Neo.CLI/Tools/VMInstruction.cs
```diff
@@ -157,17 +190,140 @@ OpCode.STSFLD or
...
+            return $"{hex} // {new UInt160(operand)}";
+
+        if (operand.Length == UInt256.Length)
+            return $"{hex} // {new UInt256(operand)}";
+
+        if (operand.Length == 4 && TryFormatUnixTimestamp(BinaryPrimitives.ReadUInt32LittleEndian(operand), out var ts32))
+            return $"{hex} // {ts32}";
+
+        if (operand.Length == 8 && TryFormatUnixTimestamp(BinaryPrimitives.ReadInt64LittleEndian(operand), out var ts64))
+            return $"{hex} // {ts64}";
```
superboyiii, line 228: `L0051:01B1 PUSHINT64 2491470000 // 2048-12-13T11:00:00Z`
`2491470000 ` is regarded as `Timestamp` in your case, seems it's not  a true `Timestamp`. This way is not good.

cschuchardt88: Agreed. A 9-10 digit PUSHINT in 2000-2100 is not enough to call it a timestamp (2491470000 is a contract integer, not 2048-12-13). Timestamp guessing is removed; integers and 4/8-byte blobs print as numbers/hex only. Text, UInt160, UInt256, and ECPoint comments stay.

src/Neo.CLI/Tools/VMInstruction.cs
```diff
@@ -157,17 +190,109 @@ OpCode.STSFLD or
...
+        if (operand.Length is not (33 or 65))
+            return false;
+        if (operand[0] is not (0x02 or 0x03 or 0x04))
+            return false;
+        try
+        {
+            point = ECPoint.DecodePoint(operand, ECCurve.Secp256r1);
+            return true;
+        }
+        catch (FormatException)
```
cschuchardt88, line 284: `ECPoint.DecodePoint` does not report “not a point” as `FormatException`:

- Compressed (`0x02`/`0x03`): `DecompressPoint` throws **`ArithmeticException`** when `alpha.Sqrt()` is null (X is not on the curve).
- Uncompressed (`0x04`): `ECFieldElement` throws **`ArgumentException`** when x or y ≥ Q.

Those exceptions leave `DecodeOperand` → `ToString` → `FormatListing`, and `ScriptsToOpCode`’s empty `catch` returns **null**. A valid script that `PUSHDATA1`s 33 bytes starting with `0x02`/`0x03` that are not a secp256r1/k1 point is then treated as unparseable. The same gap skips the k1 fallback: an r1 `ArithmeticException` never reaches the second `DecodePoint`.

Catch `Exception` (or at least `FormatException`, `ArithmeticException`, and `ArgumentException`) and return `false` so the payload is labeled `// blob 33 bytes`. A test with `02` + 32 zero bytes would lock this in.

Uncompressed `DecodePoint` also does not check the point is on the curve, so most 65-byte `0x04` blobs will still be annotated as keys. Worth verifying `Y² == X³ + aX + b` if that comment should only appear for real public keys.

src/Neo.CLI/Tools/VMInstruction.cs
```diff
@@ -97,11 +100,43 @@ IEnumerator IEnumerable.GetEnumerator() =>
...
+    /// and grows with the script size / instruction count.
+    /// </summary>
+    internal static string FormatListing(ReadOnlyMemory<byte> script)
+    {
+        var rows = new List<VMInstruction>();
+        foreach (var instruct in new VMInstruction(script))
+            rows.Add(instruct);
+
+        var lastIndex = Math.Max(0, rows.Count - 1);
+        var width = DecimalWidth(Math.Max(lastIndex, script.Length));
```
cschuchardt88, line 120: `L{n}` is the instruction index (`0 .. rows.Count-1`), not a byte offset — the hex field after `:` already grows via `HexWidth(_script.Length)`.

`DecimalWidth(Math.Max(lastIndex, script.Length))` therefore over-pads sparse scripts: a 10 KiB script with a dozen `PUSHDATA` instructions becomes `L00000:` even though the last line is `L0011:`. The remark about listings longer than 9999 **lines** only needs `lastIndex`.

Prefer `DecimalWidth(lastIndex)`.

src/Neo.CLI/CLI/MainService.Tools.cs
```diff
@@ -409,37 +409,18 @@ private static string Base64Fixed(string str)
...
-    /// PUSH4
-    /// PACK
-    /// PUSH15
-    /// PUSHDATA1 transfer
-    /// PUSHDATA1 0xa51b176d6b9bafa1f0ff8515e462d78c296e52a3
-    /// SYSCALL System.Contract.Call
+    /// output (VMInstruction.ToString, with an L-prefix padded to at least 4 digits):
+    /// L0000:0000 PUSHDATA1 64617461 // data
+    /// L0001:0006 PUSHINT32 500000000
+    /// L0002:000B PUSHDATA1 0AA2C42334D65F69B304E376559B56CF05AC5CF7
```
cschuchardt88, line 415: This 20-byte operand is not valid UTF-8, so `FormatPushData` will append a `UInt160` comment (`// 0x…`). The sample still shows raw hex only; worth updating so the doc matches `parse` output.

cschuchardt88 (commented):
## Summary

`parse` / `ScriptsToOpCode` now shares one layout via `VMInstruction.ToString` / `FormatListing`, and `DecodeOperand` annotates PUSHDATA as strict UTF-8 text, ECPoint, UInt160/UInt256, or `blob N bytes`. Dropping unix-timestamp guesses (commit `893abd3`) matches the earlier review thread.

One real bug: invalid compressed keys throw out of `TryDecodeEcPoint` and make `parse` return nothing for an otherwise valid script. L-prefix width is also tied to byte length rather than instruction index, and the XML sample / PR body are slightly stale.

## Issues

1. **bug** `src/Neo.CLI/Tools/VMInstruction.cs:284` — Status: open  
   `TryDecodeEcPoint` only catches `FormatException`. Invalid compressed points throw `ArithmeticException`; uncompressed x/y ≥ Q throw `ArgumentException`. Those escape into `FormatListing` and `ScriptsToOpCode` returns null.

2. **suggestion** `src/Neo.CLI/Tools/VMInstruction.cs:120` — Status: open  
   `DecimalWidth` pads `L{n}` from script **byte** length. `L{n}` is an instruction index; hex positions already grow with `HexWidth`.

3. **nit** `src/Neo.CLI/CLI/MainService.Tools.cs:415` — Status: open  
   The 20-byte sample is a `UInt160` and should include the `// 0x…` comment.

The PR description still lists unix seconds/milliseconds on `PUSHINT32`/`PUSHINT64` and 4/8-byte PUSHDATA. That path was removed in `893abd3`; please update the description.

PR #1099, opened 2026-08-28
Add neo-tui console UI with plugin load and live show state

cschuchardt88:
## Summary

`neo-tui` (`src/Neo.ConsoleUI`) is a cross-platform console UI for neo-cli (Windows, Linux, macOS). It is meant to **replace `neo-gui`**, the Windows-only desktop app: same node (`MainService`), same plugins, same commands, without WinForms.

It starts the same `MainService` node, loads plugins the same way, and then shows a live **show state** dashboard instead of a line prompt.

Docs and screenshots: [`docs/neo-tui/README.md`](https://github.com/cschuchardt88/neo-node/blob/feature/console-ui/docs/neo-tui/README.md)

```bash
dotnet run --project src/Neo.ConsoleUI
```

Startup flags match neo-cli: `--config`/`-c`, `--wallet`/`-w`, `--password`/`-p`, `--db-engine`, `--db-path`, `--plugins`, `--noverify`.

## Live status

The home screen is neo-cli `show state`, refreshed in place: blockchain, memory/CPU, mempool, network, sync bar, wallet, and loaded plugins.

**CONNECTED PEERS** comes from `LocalNode.GetRemoteNodes()` (address, height, user agent). The title sits on the time/uptime row; the table lines up with **BLOCKCHAIN STATUS**. Five rows are visible; extra peers scroll with **Up/Down** (also J/K, PgUp/PgDn, Home/End).

| Key | Action |
| --- | --- |
| `M` / Enter | Command menu |
| `C` | Command line (same `neo>` prompt as neo-cli) |
| `H` | Help |
| Up / Down | Scroll peers |
| `Q` / Esc | Quit |

Wide status

Peers scrolled

Narrow status

## Menus and command output

Command menu

Plugin commands

Please wait

Output stays on this console:

1. Spinner: **Please wait — running …** until the command finishes
2. Bordered popup with captured stdout
3. **Rerun command** or **Close**

plugins output

log block 1

Command parameters

Command line (`C`) — same `neo>` prompt as neo-cli:

Command line

## Plugin loading (same path as neo-cli)

- Working directory is `AppContext.BaseDirectory` (`Plugins/`, `config.json`, auto-install zips).
- Waits for `MainService.WhenStarted` before building the command catalog, then scans `Plugin.Plugins` for `[ConsoleCommand]`.
- `--plugins` is forwarded to neo-cli.
- Build copies every node plugin to `bin/$(Configuration)/$(TargetFramework)/Plugins/<Name>/` (created if missing), plus extra dependency dlls and RID natives (`libleveldb.dll`, …) next to the exe. Same copy for neo-cli and neo-tui (`src/NodePlugins.props`, `src/CopyNodePlugins.targets`).

## Implementation notes

- `MainService.WhenStarted` completes after `StartCore` so the TUI does not race plugin registration.
- `CreateBroadcastTask` / `GetMaxPeerBlockHeight` are `internal` (`InternalsVisibleTo` `neo-tui`).
- README and screenshots live in `docs/neo-tui/`.


shargon:
I like it, @superboyiii could you test it please?

superboyiii:
> I like it, @superboyiii could you test it please?

Sure

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -329,6 +333,19 @@ void DisplayError(string primaryMessage, string? secondaryMessage = null)
     }
 
     public async void Start(CommandLineOptions options)
+    {
+        try
+        {
+            await StartCore(options).ConfigureAwait(false);
+            _startCompleted.TrySetResult();
```
superboyiii:
```suggestion
            if (_neoSystem is null)
            {
                _startCompleted.TrySetException(new InvalidOperationException(
                    "NEO node failed to start. See console output for details."));
                return;
            }
```

superboyiii:
Another changes are good. Tested.

superboyiii (dismissed):

ajara87 (dismissed):

vncoelho (dismissed):

ajara87:
@shargon merge?

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -329,6 +333,24 @@ void DisplayError(string primaryMessage, string? secondaryMessage = null)
...
+    {
+        try
+        {
+            await StartCore(options).ConfigureAwait(false);
+            if (_neoSystem is null)
+            {
+                _startCompleted.TrySetException(new InvalidOperationException(
+                    "NEO node failed to start. See console output for details."));
+                return;
+            }
```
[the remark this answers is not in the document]
cschuchardt88: Fixed: _startCompleted.TrySetResult() after a successful StartCore so WhenStarted completes and neo-tui can enter the UI.

src/Neo.ConsoleUI/CommandOutputPopup.cs
```diff
@@ -0,0 +1,128 @@
...
+        var stdout = Console.Out;
+        var stderr = Console.Error;
+        var found = true;
+        var ok = true;
+
+        AnsiConsole.Clear();
+        AnsiConsole.Status()
+            .Spinner(Spinner.Known.Dots)
+            .SpinnerStyle(Style.Parse("green"))
+            .Start($"Please wait — running {commandLine}", _ =>
```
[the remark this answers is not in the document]
cschuchardt88: Fixed: menu password values are invoked quoted but displayed as `****`, and Spectre headers/status text use `Markup.Escape`.

src/Neo.ConsoleUI/CommandOutputPopup.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                try
+                {
+                    Console.SetOut(buffer);
+                    Console.SetError(buffer);
+                    found = invoke(commandLine);
```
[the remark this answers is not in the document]
cschuchardt88: Fixed: command output is teed to the live console instead of capturing behind a spinner, so `ReadUserInput` prompts stay visible.

src/Neo.ConsoleUI/InteractiveShell.cs
```diff
@@ -0,0 +1,272 @@
...
+        table.AddRow("Ctrl+U", "Clear line");
+        table.AddRow("Ctrl+C", "Cancel line / later Exit");
+        AnsiConsole.Write(table);
+        AnsiConsole.WriteLine();
+    }
+
+    private static string FormatChoice(CommandInfo command)
+        => string.IsNullOrWhiteSpace(command.Description)
+            ? command.Key
+            : $"{command.Key}  [grey]{Markup.Escape(command.Description)}[/]";
```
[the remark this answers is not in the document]
cschuchardt88: Fixed: menu labels include the parameter type list (`broadcast block  (UInt256)` vs `(UInt32)`) so overloads are distinct.

src/Neo.ConsoleUI/InteractiveShell.cs
```diff
@@ -0,0 +1,272 @@
...
+
+    private static bool IsPassword(ParameterInfo parameter)
+        => parameter.Name is not null && parameter.Name.Contains("password", StringComparison.OrdinalIgnoreCase);
+
+    private static string Quote(string value)
+        => value.Contains(' ', StringComparison.Ordinal) ? $"\"{value.Replace("\"", "\\\"", StringComparison.Ordinal)}\"" : value;
```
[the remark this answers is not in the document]
cschuchardt88: Fixed: every menu value is quoted and backslashes/quotes are escaped so Windows paths tokenize as a single argument.

src/Neo.CLI/CLI/MainService.cs
cschuchardt88: The null `_neoSystem` check is already on this branch. Also complete `WhenStarted` on the success path so neo-tui does not wait forever after a good start.

src/Neo.ConsoleUI/CommandLineApp.cs
```diff
@@ -0,0 +1,122 @@
...
+            var rest = new Argument<string[]>("args")
+            {
+                Description = "Command arguments (same syntax as neo-cli).",
+                Arity = ArgumentArity.ZeroOrMore
+            };
+            current.Arguments.Add(rest);
+            current.SetAction(parseResult =>
+            {
+                var extra = parseResult.GetValue(rest) ?? [];
+                var line = extra.Length == 0 ? info.Key : $"{info.Key} {string.Join(' ', extra)}";
```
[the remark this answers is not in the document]
cschuchardt88: Fixed: extra System.CommandLine tokens are re-quoted with CommandTokenizer.Quote before OnCommand so a message with spaces stays one argument.

src/Neo.ConsoleUI/Program.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        var startArgs = bootstrap.ToMainServiceArgs(parseResult);
+        if (!startArgs.Contains("--config", StringComparer.OrdinalIgnoreCase))
+        {
+            var besideExe = Path.Combine(AppContext.BaseDirectory, "config.json");
+            if (File.Exists(besideExe))
+                startArgs = [.. startArgs, "--config", besideExe];
+        }
+
+        if (!service.OnStart(startArgs))
```
cschuchardt88, line 44: `OnStart` always runs before the real command catalog exists, so `neo-tui --help` (and `-h`) still constructs `MainService`, opens storage, loads plugins, and waits for `WhenStarted`.

Handle help/version from the first `ParseResult` (e.g. `parseResult.Action` is a help/version action in System.CommandLine 2.x) and return before starting the node. Same for a parse that is only `--help` with no interactive intent.

src/Neo.ConsoleUI/Program.cs
```diff
@@ -0,0 +1,101 @@
...
+            return 1;
+
+        try
+        {
+            service.WhenStarted.GetAwaiter().GetResult();
```
cschuchardt88, line 49: This wait is effectively uninterruptible. `ConsoleServiceBase.OnStart` registers `CancelHandler`, which sets `e.Cancel = true` and calls `TriggerGracefulShutdown()`. That method returns immediately while `_running` is false — and the TUI never calls `RunConsole()` / `WaitForShutdown()`, so `_running` stays false through plugin install and `ImportBlocksFromFile`.

Ctrl+C during startup therefore neither kills the process nor completes/cancels `WhenStarted`. Honor cancellation here (or set `_running` / don’t swallow SIGINT until the shell is actually up).

src/Neo.ConsoleUI/Program.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            foreach (var error in fullParse.Errors)
+                AnsiConsole.MarkupLine($"[red]{Markup.Escape(error.Message)}[/]");
+            service.OnStop();
+            return 1;
+        }
+
+        try
+        {
+            if (fullParse.CommandResult.Command != app.Root && !IsShowState(fullParse))
```
cschuchardt88, line 74: `--help` / `--version` keep `CommandResult.Command == Root` (the help option lives on the root). This condition then skips `Invoke()` and falls through to `InteractiveShell`, so help text is never printed even after the node is already running.

Treat a help/version action (or `fullParse.Action is not null` while still on Root) as `Invoke()` + exit, and only enter the TUI when the user asked for the root interactive command (or `show state`).

src/NodePlugins.props
```diff
@@ -0,0 +1,28 @@
+<Project>
+  <!-- All node plugins (not RpcClient / MPTTrie libraries). Copied next to the exe: $(TargetDir)Plugins/<Name>/ -->
+  <ItemGroup>
+    <NodePlugin Include="ApplicationLogs" />
+    <NodePlugin Include="DBFTPlugin" />
```
cschuchardt88, line 5: `Plugin.LoadPlugins()` instantiates every directory under `Plugins/` next to the exe. Shipping **all** of these on every neo-cli / neo-tui build is not “available to `install`” — they actually start.

`plugins/DBFTPlugin/DBFTPlugin.json` has `"AutoStart": true`. `RpcServer.json` binds `127.0.0.1:10332` with no auth. Because `Neo.CLI.csproj` imports this file too, `dotnet run --project src/Neo.CLI` also becomes a consensus+RPC node, which is a behavior change for the existing CLI.

Copy only storage/wallet providers by default, or keep optional plugins out of `Plugins/` until `install` / `--plugins`. At minimum, don’t auto-copy `DBFTPlugin` onto neo-cli.

src/Neo.ConsoleUI/CommandLineApp.cs
```diff
@@ -0,0 +1,125 @@
...
+        Config = NewOption<string?>("--config", "Specifies the config file.", "-c", "/config");
+        Wallet = NewOption<string?>("--wallet", "The path of the neo3 wallet [*.json].", "-w", "/wallet");
+        Password = NewOption<string?>("--password", "Password to decrypt the wallet.", "-p", "/password");
+        DbEngine = NewOption<string?>("--db-engine", "Specify the db engine.", "/db-engine");
+        DbPath = NewOption<string?>("--db-path", "Specify the db path.", "/db-path");
+        Plugins = new Option<string[]>("--plugins")
+        {
+            Description = "Plugins to install if not already present. Repeat the option for each plugin.",
+            Arity = ArgumentArity.ZeroOrMore,
+            AllowMultipleArgumentsPerToken = false
```
cschuchardt88, line 40: The PR says startup flags match neo-cli, but they don’t:

- neo-cli’s `--plugins` is `string[]` and accepts `--plugins plugin1 plugin2`. `AllowMultipleArgumentsPerToken = false` requires repeating `--plugins` for each name; extra tokens become unmatched on the second parse (after the node has already started).
- `--verbose` and `--background` from `CommandLineOptions` are never declared, so `neo-tui --verbose` fails `fullParse.Errors` and shuts the node down.

Mirror neo-cli’s option shapes (or parse with the same `CommandLineOptions` map) and reject unknown flags **before** `OnStart`.

src/Neo.ConsoleUI/InteractiveShell.cs
```diff
@@ -0,0 +1,278 @@
...
+            if (line.Equals("exit", StringComparison.OrdinalIgnoreCase) ||
+                line.Equals("quit", StringComparison.OrdinalIgnoreCase))
+                return true;
+            if (line.Trim().Equals("show state", StringComparison.OrdinalIgnoreCase))
+                return false;
+            RunLine(line.Trim());
+        }
+    }
+
+    private void RunCommand(CommandInfo command)
```
cschuchardt88, line 182: `RunCommandLine` treats `exit`/`quit` as leave-the-TUI, and `show state` is special-cased back to the dashboard. Palette/menu `exit` still comes through `RunCommand` → `OnCommand("exit")` → `OnExit()`, which only sets `_running = false`. The status loop never consults `_running`, so the user gets an empty output popup and lands back on the dashboard.

Handle `exit`/`quit` here the same way as the command line (return a “quit shell” flag) instead of invoking them as captured commands.

plugins/Directory.Build.props
```diff
@@ -28,4 +28,31 @@
     </None>
   </ItemGroup>
 
+  <!-- Copy this plugin next to neo-cli / neo-tui binaries: {TargetDir}Plugins/{Name}/ -->
+  <PropertyGroup>
+    <CopyToNodePlugins Condition="'$(CopyToNodePlugins)' == ''">true</CopyToNodePlugins>
+  </PropertyGroup>
+
+  <Target Name="CopyPluginToNodeOutput" AfterTargets="Build" Condition="'$(CopyToNodePlugins)' == 'true'">
```
cschuchardt88, line 36: This copies each plugin into both neo-cli and neo-tui `Plugins/` on the plugin’s own Build. `src/CopyNodePlugins.targets` then copies the same set again after the host Build (plus natives / MPTTrie).

Two AfterBuild copy pipelines will drift (config json names, extra dependency dlls, RID natives). Prefer a single copy target owned by the host projects.

neo-node.sln
```diff
@@ -73,140 +73,414 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Neo.Plugins.Storage.Tests",
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Neo.Plugins.DeferredRelay.Tests", "tests\Neo.Plugins.DeferredRelay.Tests\Neo.Plugins.DeferredRelay.Tests.csproj", "{8F4BAC23-5C6E-7F90-B012-34567890BCDE}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Neo.ConsoleUI", "src\Neo.ConsoleUI\Neo.ConsoleUI.csproj", "{9B429DB4-922F-460E-9A6A-1541FCAA866C}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
```
cschuchardt88, line 81: This PR adds `Debug|x64` / `x86` and `Release|x64` / `x86` for every project, all mapped back to Any CPU (~270 lines), plus a UTF-8 BOM on line 1. None of that is needed for neo-tui (the project is already added under the src folder).

Please drop the extra solution platforms and keep Any CPU only, unless there is a real RID-specific build to land in this change.

src/Neo.ConsoleService/Neo.ConsoleService.csproj
```diff
@@ -4,4 +4,9 @@
     <PackageReference Include="System.ServiceProcess.ServiceController" Version="10.0.9" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Neo.ConsoleUI" />
```
cschuchardt88, line 8: `Neo.ConsoleUI.csproj` sets `<AssemblyName>neo-tui</AssemblyName>`, so `InternalsVisibleTo Include="Neo.ConsoleUI"` never matches. The `neo-tui` entry below is the one that works — this line can go.

cschuchardt88 (commented):
## Summary

Adds `neo-tui` as a Spectre.Console front-end over the same `MainService` node as neo-cli (live `show state`, peer scroll, menus/palette, line editor, one-shot subcommands), plus `WhenStarted`, `CommandTokenizer.Quote`, and copy-on-build for node plugins.

Previous notes around `WhenStarted` completion, path quoting, password masking, overload labels, and teed command output look addressed. Remaining issues are launch/packaging: `--help` starts a full node and then the TUI; Ctrl+C cannot abort `WhenStarted`; `--plugins`/`--verbose` do not match neo-cli; and copying every plugin into `Plugins/` auto-loads DBFT (`AutoStart: true`) and RpcServer for **both** neo-tui and neo-cli.

Inline comments are on the right-side lines below.

ajara87:
UT fails

PR #1100, opened 2026-08-29
fix(WalletAPI): invoke unclaimedGas at GetBlockCount height

cschuchardt88:
`getunclaimedgas` uses `NativeContract.NEO.UnclaimedGas(..., CurrentIndex + 1)`. `WalletAPI.GetUnclaimedGasAsync` invoked `unclaimedGas` with `GetBlockCount() - 1` (`CurrentIndex`), so the client helper under-counted by one block versus the RPC method.

Independent of other PRs. RpcClient helper only; not consensus-critical.


shargon (approved):

ajara87 (approved):

superboyiii (approved):
Correct.

PR #1102, opened 2026-08-29
fix(WalletAPI): sleep between WaitTransaction retries on RPC errors

cschuchardt88:
`WaitTransactionAsync` swallowed `GetRawTransaction` failures with an empty catch and no delay. A missing transaction or persistent RPC error busy-looped the CPU until timeout.

Sleep on both unknown-tx and exception paths, with a 100ms floor, and retry until the tx has confirmations.

Independent of other PRs. RpcClient helper only; not consensus-critical.

### DIFF
--- a/plugins/RpcClient/WalletAPI.cs
+++ b/plugins/RpcClient/WalletAPI.cs
@@ -199,6 +199,7 @@ public class WalletAPI
     {
         DateTime deadline = DateTime.UtcNow.AddSeconds(timeout);
         RpcTransaction rpcTx = null;
+        var pollDelay = Math.Max(100, (int)rpcClient.protocolSettings.MillisecondsPerBlock / 2);
         while (rpcTx == null || rpcTx.Confirmations == null)
         {
             if (deadline < DateTime.UtcNow)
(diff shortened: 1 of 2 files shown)


vncoelho (approved):

plugins/RpcClient/WalletAPI.cs
```diff
@@ -211,10 +212,15 @@ public async Task<RpcTransaction> WaitTransactionAsync(Transaction transaction,
...
-                    await Task.Delay((int)rpcClient.protocolSettings.MillisecondsPerBlock / 2);
+                    await Task.Delay(pollDelay).ConfigureAwait(false);
                 }
             }
-            catch (Exception) { }
+            catch (Exception)
+            {
+                // Unknown tx and transient RPC errors retry until timeout;
+                // sleep so a persistent failure cannot busy-loop the CPU.
+                await Task.Delay(pollDelay).ConfigureAwait(false);
```
shargon, line 222: use finally and do it only once?

shargon (approved):

PR #1108, opened 2026-08-29
fix(ApplicationLogs): reject invalid getapplicationlog triggerType

cschuchardt88:
Docs say an invalid trigger throws. `Enum.TryParse` failing skipped the filter, so callers got unfiltered executions. Numeric strings such as `16` also parsed as undefined enum values.

Require a named `TriggerType` or return `InvalidParams`.

Independent of other PRs.

### DIFF
--- a/plugins/ApplicationLogs/LogReader.cs
+++ b/plugins/ApplicationLogs/LogReader.cs
@@ -118,13 +118,18 @@ public class LogReader : Plugin
             if (raw == null) throw new RpcException(RpcError.InvalidParams.WithData("Unknown transaction/blockhash"));
         }
 
-        if (!string.IsNullOrEmpty(triggerType) && Enum.TryParse(triggerType, true, out TriggerType _))
+        if (!string.IsNullOrEmpty(triggerType))
         {
+            if (!Enum.TryParse(triggerType, true, out TriggerType trigger)
+                || !Enum.GetNames<TriggerType>().Any(n => n.Equals(triggerType, StringComparison.OrdinalIgnoreCase)))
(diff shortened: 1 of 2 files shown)


plugins/ApplicationLogs/LogReader.cs
```diff
@@ -118,13 +118,18 @@ public JToken GetApplicationLog(UInt256 hash, string triggerType = "")
             if (raw == null) throw new RpcException(RpcError.InvalidParams.WithData("Unknown transaction/blockhash"));
         }
 
-        if (!string.IsNullOrEmpty(triggerType) && Enum.TryParse(triggerType, true, out TriggerType _))
+        if (!string.IsNullOrEmpty(triggerType))
         {
+            if (!Enum.TryParse(triggerType, true, out TriggerType trigger)
+                || !Enum.GetNames<TriggerType>().Any(n => n.Equals(triggerType, StringComparison.OrdinalIgnoreCase)))
```
shargon, line 124: Isn't TryParse enough?

ajara87 (approved):

plugins/ApplicationLogs/LogReader.cs
superboyiii: Seems not enough. .NET’s `Enum.TryParse` accepts numeric strings as well as names. So values like `"16"` (undefined) or `"64"` (Application =`0x40`) both succeed, even though they aren’t named `TriggerType` values.

superboyiii (approved):
Tested. It works.

shargon (approved):

